PR #425, opened 2014-12-01
Update updatecheck.py

eigerman:
Check for the addition of an icon_hash key in a package plist. If the key is present and matches a hash of an icon that has already been downloaded no attempt will be made to download the icon again. If the icon_hash is not present or the hash does not match the behavior will revert back to normal.

### DIFF
--- a/code/client/munkilib/updatecheck.py
+++ b/code/client/munkilib/updatecheck.py
@@ -28,6 +28,7 @@ import subprocess
 import socket
 import urllib2
 import urlparse
+import xattr
 from urllib import quote_plus
 from OpenSSL.crypto import load_certificate, FILETYPE_PEM
 
@@ -2746,7 +2747,12 @@ def download_icons(item_list):
         icon_url = icon_base_url + urllib2.quote(icon_name)
         icon_path = os.path.join(icon_dir, icon_name)
         if os.path.isfile(icon_path):
-            local_icon_hash = munkicommon.getsha256hash(icon_path)
+            local_xattrs = xattr.xattr(icon_path)
+            if 'icon_hash' in local_xattrs:
+                local_icon_hash = local_xattrs['icon_hash']
+            else:
+                local_icon_hash = munkicommon.getsha256hash(icon_path)
+                local_xattrs['icon_hash'] = local_icon_hash
         else:
             local_icon_hash = 'nonexistent'
         icon_subdir = os.path.dirname(icon_path)
--- a/code/client/munkilib/updatecheck.py
+++ b/code/client/munkilib/updatecheck.py
@@ -2746,10 +2746,10 @@ def download_icons(item_list):
         icon_url = icon_base_url + urllib2.quote(icon_name)
         icon_path = os.path.join(icon_dir, icon_name)
         if os.path.isfile(icon_path):
-            xattr_hash = fetch.getxattr(icon_path, XATTR_SHA)
+            xattr_hash = fetch.getxattr(icon_path, fetch.XATTR_SHA)
             if not xattr_hash:
               xattr_hash = munkicommon.getsha256hash(icon_path)
-              fetch.writeCachedChecksum(icon_path, XATTR_SHA)
+              fetch.writeCachedChecksum(icon_path, xattr_hash)
         else:
             xattr_hash = 'nonexistent'
         icon_subdir = os.path.dirname(icon_path)
@@ -2772,7 +2772,7 @@ def download_icons(item_list):
                     icon_name, err)
             else:
                 if os.path.isfile(icon_path):
-                    fetch.writeCachedChecksum(icon_path, XATTR_SHA)
+                    fetch.writeCachedChecksum(icon_path)
     # remove no-longer needed icons from the local directory
     for (dirpath, dummy_dirnames, filenames) in os.walk(
             icon_dir, topdown=False):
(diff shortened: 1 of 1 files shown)


gregneagle:
Ping

gregneagle:
Ed: waiting for some sort of response from you. I commented on Dec 1st and again on the 15th; Arjen also has an alternative implementation. If I don't hear from you within a few days I will close this and work with Arjen on his implementation.

eigerman:
Hey, sorry. I'll work on that this week.

gregneagle:
Please talk with Arjen as well; would be nice to have one implementation you can agree on.

eigerman:
Will do.

code/client/munkilib/updatecheck.py
```diff
@@ -3056,8 +3062,11 @@ def check(client_id='', localmanifestpath=None):
         # and active installs/removals
         item_list = list(installinfo.get('optional_installs', []))
         item_list.extend(installinfo['managed_installs'])
-        item_list.extend(installinfo['removals'])
         download_icons(item_list)
+        # We don't need to download iconds for items being removed,
+        # they should alredy be present and if they aren't we're
```
eigerman: But if we're removing Photoshop CS4 and don't already have the icon, why do we want it?

gregneagle: So that when Managed Software Center displays info to the user it has a pretty icon as well. Same as when we are installing.

eigerman: OK. I'm going to update this whole thing shortly. I'll revert this line.

code/client/munkilib/updatecheck.py
```diff
@@ -2751,16 +2760,19 @@ def download_icons(item_list):
...
-        item_name = item.get('display_name') or item['name']
-        message = 'Getting icon for %s...' % item_name
-        try:
-            dummy_value = getResourceIfChangedAtomically(
-                icon_url, icon_path, message=message)
-        except fetch.MunkiDownloadError, err:
-            munkicommon.display_debug1(
-                'Could not retrieve icon %s from the server: %s',
-                icon_name, err)
+        if pkginfo_icon_hash != xattr_hash:
```
eigerman: Our internal implementation uses:

if pkginfo_icon_hash and (pkginfo_icon_hash != xattr_hash):

So it will only check for icons if the hash is set, but that's probably not what most users want.

eigerman:
OK, Please take another look.

code/client/munkilib/updatecheck.py
gregneagle: More importantly:

Not what most _admin_ would want, and most importantly:

Would break existing sites (IOW, icons would no longer be retrieved, since no existing sites other than you are using pkginfo_icon_hash)

eigerman: Yeah, I just wanted to note this if others want to tinker with it.

gregneagle:
This looks good to me -- bochoven, can you give it a once-over?

gregneagle:
And interesting that GitHub is now saying it can't auto-merge this...

code/client/munkilib/updatecheck.py
```diff
@@ -2738,11 +2739,19 @@ def download_icons(item_list):
...
         if not os.path.splitext(icon_name)[1] in icon_known_exts:
             icon_name += '.png'
         icon_list.append(icon_name)
         icon_url = icon_base_url + urllib2.quote(icon_name)
         icon_path = os.path.join(icon_dir, icon_name)
+        if os.path.isfile(icon_path):
+            xattr_hash = fetch.getxattr(icon_path, XATTR_SHA)
+            if not xattr_hash:
+              xattr_hash = munkicommon.getsha256hash(icon_path)
+              fetch.writeCachedChecksum(icon_path, XATTR_SHA)
```
gregneagle: The second parameter is the hash value to store, not the name of the attribute. I think your previous commit was correct here.

code/client/munkilib/updatecheck.py
```diff
@@ -2751,16 +2760,19 @@ def download_icons(item_list):
...
+            try:
+                dummy_value = getResourceIfChangedAtomically(
+                    icon_url, icon_path, message=message)
+            except fetch.MunkiDownloadError, err:
+                munkicommon.display_debug1(
+                    'Could not retrieve icon %s from the server: %s',
+                    icon_name, err)
+            else:
+                if os.path.isfile(icon_path):
+                    fetch.writeCachedChecksum(icon_path, XATTR_SHA)
```
gregneagle: Same as above comment -- here I think you want to leave off the second parameter so that fetch.writeCachedChecksum will calculate the (current) checksum for you.

gregneagle:
fetch.writeCachedChecksum:

```python
def writeCachedChecksum(file_path, fhash=None):
    """Write the sha256 checksum of a file to an xattr so we do not need to
       calculate it again. Optionally pass the recently calculated hash value.
    """
```

code/client/munkilib/updatecheck.py
eigerman: Yeah, I was trying to do some last minute work on this as I rushed out the door, which is always a mistake.

code/client/munkilib/updatecheck.py
eigerman: Done.

gregneagle:
This overall looks good to me, but since GitHub can't auto merge it, it's going to have to wait for me to clone, test, then manually merge. I wonder what is the conflict preventing auto-merge?

timsutton:
AFAIK GitHub's "auto-merge" is a fast-forward merge, and that can only happen if you haven't made any changes to `updatecheck.py` (or whatever was changed in Ed's master) since he initially forked it. Since you've made changes to that file in (at least) the config profile support, that's what's preventing GitHub from merging.

One of you would need to do the manual merge or Ed could rebase his master on top of the current master.

eigerman:
I can't do the manual merge, I think, but I can rebase in a new pull
request.

eigerman:
I fixed it, I think.

gregneagle:
Merged. Please pull and test.

PR #435, opened 2014-12-18
Add support for icon_hash

bochoven:
A rewrite of #425 which uses the xattr attribute com.googlecode.munki.sha256 to store the current hash. 

Check for the addition of an icon_hash key in a package plist. If the key is present and matches a hash of the corresponding icon that has already been downloaded, no attempt will be made to download the icon/verify the icon with the server. If the icon_hash is not present or the hash does not match the standard procedure for checking icons will be followed.

### DIFF
--- a/code/client/munkilib/fetch.py
+++ b/code/client/munkilib/fetch.py
@@ -258,6 +258,8 @@ def getResourceIfChangedAtomically(url,
             xattr_hash = writeCachedChecksum(destinationpath)
         if xattr_hash == expected_hash:
             #File is already current, no change.
+            munkicommon.display_debug1('%s: stored hash matches '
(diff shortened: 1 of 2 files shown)


bochoven:
bump

gregneagle:
I really wish Ed E would weigh in: seems he has disappeared.

bochoven:
Ok, thought you were holding off until I added hash generation in munkiimport..

gregneagle:
Adding hash generation would be appreciated; seems like that would take a few minutes at most to implement.

bochoven:
Writing the hashing code is simple, but knowing where to put it is hard...

bochoven:
Issue has been addressed by #425

PR #436, opened 2014-12-19
new MunkiStatus view

SteveKueng:
New more AppleUpdate like MunkiStatus view:
screen shot 2014-12-19 at 07 27 24


keeleysam:
A link to the Google Groups discussion: https://groups.google.com/forum/#!msg/munki-dev/OlqlvFjM7So/piUnbpyoWpQJ

I'll have another look at this and what it needs to be merged.

code/client/munkilib/appleupdates.py
```diff
@@ -640,7 +640,7 @@ def _GetPreferredLocalization(self, list_of_localizations):
         localizations. Returns a single language/localization name.'''
         # pylint: disable=no-self-use
         localization_preferences = (
-            munkicommon.pref('AppleSoftwareUpdateLanguages') or ['English'])
+            munkicommon.getAppleLanguages() or ['English'])
```
gregneagle: Not really liking this change. Munki, running as root, needs a definitive set of languages to get. If instead we tie it to the currently logged-in user, that could change: first root's languages, when no one is logged in, then some logged in user's languages; then a different logged-in user's languages...

code/client/munkilib/munkicommon.py
```diff
@@ -842,6 +843,15 @@ def getFirstPlist(textString):
     return (textString[plist_start_index:plist_end_index],
             textString[plist_end_index:])
 
+def getAppleLanguages():
+    user = getconsoleuser()
```
gregneagle: getconsoleuser can return None or 'loginwindow' as well as an actual username. Does the next line handle these cases correctly?

code/client/munkilib/munkicommon.py
```diff
@@ -842,6 +843,15 @@ def getFirstPlist(textString):
...
+    userHome = pwd.getpwnam(user).pw_dir
+    try:
+        globalPreferences = FoundationPlist.readPlist(userHome + "/Library/Preferences/.GlobalPreferences.plist")
```
gregneagle: This is not-great Python. Use the os.path routines:

os.path.join(userHome, 'Library/Preferences/.GlobalPreferences.plist')

Additionally, it's generally _not_ a great idea to bypass cfprefs and directly read user preference plist files...

SteveKueng:
yes we need to split this up!
my mistake... this should be a different PR

code/client/munkilib/appleupdates.py
SteveKueng: hmm... the problem is, we have multiple users with different languages on one mac...

gregneagle: I understand that. This change won't really help that at all.

Even if it operated as (I think) intended, if Johann is logged in when managedsoftwareupdate is running in the background, Munki will get German dist files. Then Johann logs out and Jean-Louis logs in. The .dist files are still in German, at least until the next background run (and maybe even then).

Much much much more work will need to be done to correctly support the use-case of multiple users with different languages on one Mac. It's not a need I have, so I'm not really willing to go into all the effort.

SteveKueng: the behaivior is as you expected:
Johann is logged in when managedsoftwareuodate is running... munki get German dist files.
then Jean-Louis logs in and the dist are still in German... until munki runs in the background then munki get French dists

gregneagle: and if no-one is logged in? Crash or English I'd expect. So that's not great.

SteveKueng: if no one is logged in munki uses English
it's also possible to use the default/loginwindow languange

gregneagle: And that's terribly non-deterministic. For the more common case of a single user of a machine, munkicommon.pref('AppleSoftwareUpdateLanguages') ensures a consistent language for Apple's update descriptions and therefore a better user experience.

I don't want to make the common case worse when attempting to improve the behavior for your less-common use case.

PR #439, opened 2015-01-14
Warn when any Munki repo URLs are not https

natewalck:
This is to follow up on the discussion we had had at MacTech regarding munki repos that do not use https.

Any suggestions on a better place to put this sort of check is appreciated, although I think warning at preference load is probably the best as it is less likely to get lost in the output of a normal run.

### DIFF
--- a/code/client/munkilib/munkicommon.py
+++ b/code/client/munkilib/munkicommon.py
@@ -1220,6 +1220,11 @@ def pref(pref_name):
     if isinstance(pref_value, NSDate):
         # convert NSDate/CFDates to strings
         pref_value = str(pref_value)
+    if 'URL' in pref_name and not pref_value == None:
+        if not 'https' in pref_value:
+            display_warning(
+                "%s is not secure. See the munki wiki for details on"
+                "securing your repository." % pref_value)
     return pref_value
 
 #####################################################


keeleysam:
This doesn't check for the beginning of the URL - just that https is in it.  http://www.httpsolutions.com/munki wouldn't pop the warning.

natewalck:
I'm not putting any more effort into this until I know #1 this is the right place for it and #2 this sort of change will be accepted.

It would be trivial to split on :// and check the first part, etc.

gregneagle:
?? Munki currently warns if the SoftwareRepoURL is set to the default:

WARNING: Client is configured to use the default repo, which is insecure. Client could be trivially compromised when off your organization's network. Consider using a non-default URL, and preferably an https:// URL.

If an admin is not paying attention to warnings, I don't think it matters when or where the warning is printed.

natewalck:
I forgot that message was added. Would it be ok to add the https message as well? (Check for default and non secure). Although we cannot force people to use https, I think warning on non https would be good form. 


Nate

gregneagle:
The warning _does_ mention https. Here, I'll post it again:

WARNING: Client is configured to use the default repo, which is insecure. Client could be trivially compromised when off your organization's network. Consider using a non-default URL, **and preferably an https:// URL**.

(Emphasis not in the original.)

natewalck:
It says the default repo, I would like to check even non default URLs for insecure URLs. Also there are more URLs than just SoftwareRepoURL. Since you already have that message in there, I was going to slide in another check and warning. 






Nate

natewalck:
I just looked at the code. I'll submit a different pull request tonight as this method of warning is not nearly as good as the existing one.


Nate

RichardLafontaine:
Please add a way to disable these new warnings in a setting somewhere. We know we use http, I don't want my reporting software filled with 5000 of these warnings. This will prevent us from seeing other issues we do not know (like low disk space, ...). The default repo warning is important, the https, not really if we are aware of it and just didn't have the time to address it.

gregneagle:
Nate:

I think it's worthwhile to warn people who are using the default out-of-the-box settings that there are security issues.

I think once an admin has specifically, purposely set server URLs the time for warnings is over. I don't think an admin that has decided to use http URLs needs to be punished with lots of warnings.

Additionally, there's not a big security risk if IconURL is set to a http URL. I think the warning as it now stands is fine.

natewalck:
Yea. Sam decided he doesn't care anymore and I certainly don't care enough
to argue it, so kill this pull request.

Nate

Issue #334, opened 2014-09-19
Munki2 - Adding items in "other updates available" processes straight away 


natewalck:
_From [bart.rea...@gmail.com](https://code.google.com/u/111977608897407650952/) on June 19, 2014 18:32:06_

What steps will reproduce the problem? on a  system with multiple "other updates available" adding each additional item processes the download immediately after adding to pending updates, after which the view is refreshed. What is the expected output? What do you see instead? would rather be able to add all pending updates and not have them process until after "update all" is clicked. What version of the Munki tools/InstallOSpkg tools are you using? On what version of OS X? MSC 4.0.2107
OS X 10.9 Please provide any additional information below. Not sure if this is intended behaviour or not. Either way, it's an interruption to workflow that doesn't really need to happen, especially if it's a largish download on a slowish link.

_Original issue: http://code.google.com/p/munki/issues/detail?id=334_

mrala:
Would changing the `Other available updates` to checkboxes with an `Add` button help address the issue? Instead of the user having to wait 30 to 60 seconds for the `managedsoftwareupdate  —update check` each time an app is added, the update check would be performed when the user is done selecting software. Any additional updates could be presented after that and I believe should avoid any dependencies. Please feel free to correct me.

grahampugh:
How about an "Update All" button? That's consistent with Apple App Stores.

gregneagle:
"Would changing the Other available updates to checkboxes with an Add button help address the issue? Instead of the user having to wait 30 to 60 seconds for the managedsoftwareupdate --updatecheck each time an app is added, the update check would be performed when the user is done selecting software. "

How do we know when the user is done selecting software?

mrala:
`Update All` wouldn’t be optimal if the user only wants to select certain apps, I think it should be `Add`. Also, there’s is already an `Update All` button for pending updates.

@gregneagle When the click the `Add` button.

gregneagle:
What "Add" button? You might need to add detail to your proposal. Visual mockups might help, too. You have a picture in your head of how it might look/how it might work. Now help me see that picture, too.

grahampugh:
"Update all" doesn't do the "other available updates", does it? 




"Add all" would do, if you like. 


Cheers, Graham












Dr Graham R PughIT Support Specialist

University of Bristol / Zone D

Subject Specialist - Maths

Mac System Administration Specialist

PR #443, opened 2015-01-20
Wait until the user has finished making changes before refreshing.

chymb:
Hi,
One thing I've had complaints about is that it is very hard to make changes when using the Updates tab as it refreshes after each click, so I've attempted to put a delay in.
I initially had a 5 second delay but figured that was too long, so I've settled on 1.5, though I'd be interested to get others' thoughts.
Cheers,
Chris

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center/MSCMainWindowController.py
+++ b/code/apps/Managed Software Center/Managed Software Center/MSCMainWindowController.py
@@ -42,6 +42,7 @@ class MSCMainWindowController(NSWindowController):
     _alertedUserToOutstandingUpdates = False
     
     _update_in_progress = False
+    _pending_update = 0
     managedsoftwareupdate_task = None
     _update_queue = set()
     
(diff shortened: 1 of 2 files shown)


gregneagle:
I assume you are referring to what happened when you add or remove "Other available updates" in the Updates view.

See https://github.com/munki/munki/issues/334 for a discussion of the issue.

I'll try testing this proposed change to see how it "feels".

gregneagle:
I tried it. I don't think it works as I'd expect.

I think a better approach would be each time one clicks a + or x in the updates view, a timer is started. New clicks reset the timer. Once the timer expires, the update check is started. As long as people are making changes within the timer window, the timer doesn't expire.

erikng:
I would like this change.

chymb:
Yes, that's the issue #334. Sorry, it didn't occur to me to check for one.
The approach you described it is what it's supposed to be doing. Perhaps the timeout of 1.5s is not long enough to feel right.

gregneagle:
The observed behavior is that after the first click of a + or x I can continue to make other "choices" for an additional 1.5 seconds, and then the update check starts.

chymb:
Hi Greg,
I've amended that commit (is that bad practice?) so that the incrementation of the _pending_update counter happens before the CSS transition. I also increased the time-out to 3.5 seconds, which feels a bit better to me.
Let me know what you think.

gregneagle:
I don't see a significant difference in behavior. I make a change, then another, and then the update check starts.

chymb:
Hi Greg,
Sorry for the delay in getting back to this. It looks like the delay is not honoured if the MunkiItems.updateCheckNeeded() is not true, so I've taken that check out.
I initially created a "dontCheckForUpdatesAfterDelay" function but that then involved tracking the changes that had been queued to that point so that updateCheckNeeded was accurate and it seemed like a more complicated change for little benefit.
Let me know how it goes.
Cheers,
Chris

gregneagle:
Closing this old PR, which would need to be completely re-done for the new MSC.app written in Swift. This does not mean the issue has been resolved; just that this PR is unlikely to be useful

PR #444, opened 2015-01-21
modify makecatalogs to automatically generate icon hashes and insert them into catalogs

keeleysam:


bochoven:
When you run makecatalogs on a copy of the repo without the icons, you'll lose all icon references. Maybe honouring -f flag for makecatalogs makes sense?

gregneagle:
Rather than overloading the meaning of the -f flag, perhaps this functionality should be specifically invoked via a new flag. Maybe -I/--icons or similar.

keeleysam:
Updated so if the icons directory doesn't exist it keeps existing behavior

code/client/makecatalogs
```diff
@@ -150,6 +199,29 @@ def makecatalogs(repopath, options):
...
+
+            if iconhashing:
+
+                name = pkginfo.get('name')
+
+                if pkginfo.get('icon_name'):
+                    try:
+                        iconhash = icons.get(icon_name)
+                    except:
+                        del(pkginfo['icon_name'])
```
bochoven: Not so sure about deleting icon_name here, now admins don't get feedback when they use an icon_name that is pointing to a non-existent icon.

gregneagle: Agreed.

timsutton: What if it logs a warning about the missing icon file similar to a missing installer item (in pkgs) is handled, and mentions that the key was simply ignored?

code/client/makecatalogs
```diff
@@ -150,6 +199,29 @@ def makecatalogs(repopath, options):
...
                     del(pkginfo[key])
 
+
+            if iconhashing:
+
+                name = pkginfo.get('name')
+
+                if pkginfo.get('icon_name'):
+                    try:
+                        iconhash = icons.get(icon_name)
```
timsutton: `icon_name` is undefined - so this would raise an exception, but maybe not for the reason you want..

If it _were_ defined, the try/except and the one below (line 214) are doing a try on something that's unlikely to ever raise an exception, since `icons.get()` will just return `None` here on a missing key. I think it's more idiomatic to just say `icons.get(thething)` and then do what you want based on whether you got a not-None value - or in the case below, you're later checking if `iconhash` is `not None` anyway.

keeleysam:
I'm fixing this up - will update in a few mins

timsutton:
After another test run:

`updatecheck.py` currently tries to add the known extensions to an `icon_name` if one is specified without a filename extension:
https://github.com/munki/munki/blob/847ce2d3331db3f15cb8b86aed591bd7bc87af4f/code/client/munkilib/updatecheck.py#L2771-L2774

For example, I can have a "Silverlight_Settings" pkg with an `icon_name` of `Silverlight`, and MSC will display an icon called `Silverlight.png`.

You're checking membership of the `icon_name` pkginfo key in your `icons` dict, and so in the case of not having the extension in the pkginfo key (which I happened to have a lot of), there's a false warning:

`WARNING: icon_name specified in info file plugins/Silverlight/Silverlight_Settings-2014.12.12.plist but it does not exist`

timsutton:
Although png is definitely the most common extension, I'd think the name check should check all the known extensions that are done by `updatecheck.py`:

https://github.com/munki/munki/blob/847ce2d3331db3f15cb8b86aed591bd7bc87af4f/code/client/munkilib/updatecheck.py#L2763-L2764

keeleysam:
@timsutton updatecheck.py only automatically adds .png so this only does that too.  If you have an extension in icon_name this will still hash it.

timsutton:
Ah right. Sorry, I was confusing things.

bochoven:
@gregneagle add this to 2.2 'release'?

gregneagle:
That's possible, but would then require a new Release Candidate and delay the final release until at least Jan 29.

bochoven:
It would tie together all icon_hash changes though..

gregneagle:
Yes, and that's an argument for doing it. In retrospect I should have held back the icon_hash stuff.

code/client/makecatalogs
```diff
@@ -150,6 +199,31 @@ def makecatalogs(repopath, options):
...
+                    else:
+                        errors.append("WARNING: icon_name specified"
+                        " in info file %s but it does not exist" % filepath[len(pkgsinfopath)+1:])
+                else:
+                    if icons.get(name+'.png'):
+                        iconhash = icons.get(name+'.png')
+
+                # we are generating the icon hashes above so strip this key if it exists - we'll be replacing it.
+                if pkginfo.get('icon_hash'):
+                    del(pkginfo['icon_hash'])
```
gregneagle: Why bother? Just replace it later.

code/client/makecatalogs
```diff
@@ -150,6 +199,31 @@ def makecatalogs(repopath, options):
...
+
+                if iconhash is not None:
+                    pkginfo.update({'icon_hash':iconhash})
```
gregneagle: That's a complex way to do:
           pkginfo['icon_hash'] = iconhash

code/client/makecatalogs
```diff
@@ -150,6 +199,31 @@ def makecatalogs(repopath, options):
...
+                    iconhash = None
+
             #simple sanity checking
             do_pkg_check = True
             installer_type = pkginfo.get('installer_type')
```
gregneagle: Make sure all your lines are < 80 characters and have no trailing whitespace.

code/client/makecatalogs
```diff
@@ -95,6 +96,54 @@ def makecatalogs(repopath, options):
...
+
+                filepath = os.path.join(dirpath, filename)
+
+                iconpath = filepath.rsplit(iconspath+'/',1)[1]
+
+                # Try to read the icon file
+                try:
+                    print_utf8("Hashing %s..." % (iconpath))
+                    iconhash = hashlib.sha256(open(filepath, 'rb').read()).hexdigest()
+                    icons.update({iconpath:iconhash})
```
gregneagle: Again, why not:
icons[iconpath] = iconhash

keeleysam:
@gregneagle made all suggested changes

code/client/makecatalogs
```diff
@@ -150,6 +202,27 @@ def makecatalogs(repopath, options):
                 if key.startswith('_'):
                     del(pkginfo[key])
 
+            if iconhashing:
+
+                name = pkginfo.get('name')
+
+                if pkginfo.get('icon_name'):
```
gregneagle: A lot of this is repetitive. You could condense lines 210-217 to: 

```
iconhash = icons.get(pkginfo['icon_name']) or icons.get(pkginfo['icon_name'] + '.png')
if not iconhash:
    errors.append("WARNING: icon_name specified"
                  " in info file %s but it does not exist"
                  % filepath[len(pkgsinfopath) + 1:])

File: code/client/makecatalogs [hunk 8]
```diff
@@ -150,6 +202,27 @@ def makecatalogs(repopath, options):
...
+                        iconhash = icons.get(pkginfo.get('icon_name'))
+                    elif icons.get(pkginfo.get('icon_name')+'.png'):
+                        iconhash = icons.get(pkginfo.get('icon_name')+'.png')
+                    else:
+                        errors.append("WARNING: icon_name specified"
+                        " in info file %s but it does not exist"
+                        % filepath[len(pkgsinfopath)+1:])
+                else:
+                    if icons.get(name+'.png'):
+                        iconhash = icons.get(name+'.png')
```
gregneagle: Same here; this should be just:

```
iconhash = icons.get(name + '.png')
```

without the if. icon hash will get set to a value or None. No need to call `icons.get(name+'.png')` twice.

code/client/makecatalogs
keeleysam: updated, but that's more than 80 characters wide ;)

code/client/makecatalogs
keeleysam: done

code/client/makecatalogs
gregneagle: So wrap it!

```
iconhash = (icons.get(pkginfo['icon_name']) or 
            icons.get(pkginfo['icon_name'] + '.png'))
```

keeleysam: i did in the latest commits

gregneagle:
I'm going to merge, then do some additional formatting cleanup.

PR #452, opened 2015-02-09
Support http(s) redirects with client certificates

kujeger:
This commit enables redirects for http(s).
In order for https-redirects to work with client certificates, we also
need to add a wildcard identity-preference in the format of
*.example.com .
This is done with the CertificateWildcardURL preference.

### DIFF
--- a/code/client/munkilib/fetch.py
+++ b/code/client/munkilib/fetch.py
@@ -105,7 +105,7 @@ def header_dict_from_list(array):
 
 def get_url(url, destinationpath,
             custom_headers=None, message=None, onlyifnewer=False,
-            resume=False, follow_redirects=False):
+            resume=False, follow_redirects=True):
     """Gets an HTTP or HTTPS URL and stores it in
     destination path. Returns a dictionary of headers, which includes
     http_result_code and http_result_description.
@@ -233,7 +233,7 @@ def getResourceIfChangedAtomically(url,
                                    message=None,
                                    resume=False,
                                    verify=False,
-                                   follow_redirects=False):
+                                   follow_redirects=True):
     """Gets file from a URL.
        Checks first if there is already a file with the necessary checksum.
        Then checks if the file has changed on the server, resuming or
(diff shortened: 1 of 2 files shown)


gregneagle:
I'd like to treat support for HTTP redirects with https and wildcard identity preferences as separate issues. Tying them together here is just going to stall this forever.

kujeger:
As per Nate's suggestion on munki-dev I'm planning on writing it so that http-redirect is only enabled if CertificateWildcardURL is configured, just need the time for it (which I will have once I'm done creating the environment that uses the above patch)

gregneagle:
But that's not what I'm asking for.

I want these things to be two separate features, not tied together in any way.

There is no intrinsic reason to limit support for http redirects to scenarios where client certificates are in use.

natewalck:
I agree with Greg. They are not tied to each other in any way, so
separating them out makes sense.

To be clear, you saying that you want two new preferences?

Perhaps:

CertificateWildcardURL
HttpRedirect


Nate

kujeger:
Ah, apologies.
Would you prefer http(s) to be a toggleable preference, or is always-on good enough?

natewalck:
I am against always-on for following HTTP Redirects as it would be a change
to the default behavior of munki as it exists today. I think it should be a
setting, which defaults to False (as is the case with munki today).

Nate

gregneagle:
I think this needs discussion on munki-dev.

It's possible that allowing redirects for https URLs, even though it's a change, won't break anyone, and so  allowing it won't require Yet Another Preference.

I'm still not convinced of the need for CertificateWildcardURL, or that it is the best/correct solution to the problem. There are other ways to solve this issue; here a two off of the top of my head:

1) Require the admin create a properly configured munki.keychain and distribute it. This method can support any arbitrarily complex set of identity preferences.

2) A preference that is an array of strings, something like:
```xml
<key>UseClientCertificateForURLs</key>
<array>
    <string>https://server1.external.com/</string>
    <string>https://munki-*.internal.com/<string>  (or whatever the correct syntax is here -- another reason I am hesitant to do this)
    <string>https://testmunki.internal.com/<string>
</array>
```

My point is that is we are going to add a preference and more code, it better handle all foreseeable needs and not just your one scenario you see yourself using.

PR #458, opened 2015-03-13
Added uninstaller_item_location check to makecatalogs

nmcspadden:


code/client/makecatalogs
```diff
@@ -268,6 +269,46 @@ def makecatalogs(repopath, options):
                         exit_code = -1
                         continue
 
+            #uninstaller sanity checking
+            do_pkg_check = False
+            uninstaller_type = pkginfo.get('uninstall_method')
+            if uninstaller_type in ['AdobeCCPUninstaller']:
+                do_pkg_check = True
```
gregneagle: What's the purpose of defining this variable? Why not just delete lines 276-280?

code/client/makecatalogs
```diff
@@ -268,6 +269,46 @@ def makecatalogs(repopath, options):
                         exit_code = -1
                         continue
 
+            #uninstaller sanity checking
+            do_pkg_check = False
+            uninstaller_type = pkginfo.get('uninstall_method')
+            if uninstaller_type in ['AdobeCCPUninstaller']:
```
gregneagle: I think there is at least one other uninstaller_type that uses an uninstaller_item...

nmcspadden:
This should be accommodated with https://github.com/nmcspadden/munki/commit/a77c42b2a9b2c802d576f6bf0ea69b4d7f8f1746.

nmcspadden:
Should now work for the correct uninstaller types with https://github.com/nmcspadden/munki/commit/01c0b5393d3232307c41be4e8f59c9815c1d0b43

keeleysam:
Why even check the Adobe stuff?  It should just verify that if an uninstaller_item_location is defined that the file actually exists.  This is the behavior for installers.

nmcspadden:
That's also an option, but it won't address the possible case where the uninstall method is defined as "AdobeCCPUninstaller" but uninstaller_item_location is missing.

PR #460, opened 2015-03-14
Fix issue 459

dayglojesus:
This is a fairly straightforward patch. I think it addresses the desired outcome as per the list discussion. The validation routine could get cranked up a notch but I'm not sure it's required and it might get clunky.

### DIFF
--- a/code/client/makepkginfo
+++ b/code/client/makepkginfo
@@ -391,9 +391,10 @@ def has_valid_install_critieria(catinfo):
     Returns boolean.
     """
     results = []
-    for key in ['installs', 'receipts', 'installcheck_script']:
-        results.append(catinfo.has_key(key) and len(catinfo[key]) != 0)
-    any(results)
+    keys = ['installs', 'receipts', 'installcheck_script', 'PayloadIdentifier']
+    for key in keys:
+        results.append(catinfo.get(key))
+    return any(results)
 
 def main():
     '''Main routine'''
--- a/code/client/makepkginfo
+++ b/code/client/makepkginfo
@@ -1124,13 +1124,13 @@ def main():
 
     if not has_valid_install_critieria(catinfo):
         item = catinfo['name']
-        print >> sys.stderr, (
-            "WARNING: %s contains no discernible install criteria!\n"
-            "If you do not supply a receipt, install item, "
-            "installcheck_script \n"
-            "or PayloadIdentifier, munki will have no way to determine the \n"
-            "package's installation state. This can result in undesirable or \n"
-            "unexpected behaviour (ie. looping).\n") % item
+        msg  = """WARNING: \'%s\' contains no discernible install criteria!
+        If you do not supply a receipt, install item, or
+        installcheck_script, Munki will have no way to
+        determine the package's installation state. This can
+        result in repeated installation attempts or failure
+        to attempt installation at all."""
+        print >> sys.stderr, (msg) % item
 
     # and now, what we've all been waiting for...
     print FoundationPlist.writePlistToString(catinfo)
--- a/code/client/makepkginfo
+++ b/code/client/makepkginfo
@@ -749,7 +749,7 @@ def main():
         exit(0)
 
     if options.verify_options_only:
-        if (len(arguments) == 0 and not options.apple_update and 
+        if (len(arguments) == 0 and not options.apple_update and
                 not options.nopkg):
             print >> sys.stderr, 'makepkginfo options FAILED verification'
             print >> sys.stderr, 'No installer item was provided.'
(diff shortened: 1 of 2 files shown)


code/client/makepkginfo
```diff
@@ -385,6 +385,15 @@ def check_mode(option, opt, value, parser):
                                                             (opt, mode))
     setattr(parser.values, option.dest, value)
 
+def has_valid_install_critieria(catinfo):
```
gregneagle: What happens if you import a configuration profile? I imagine we'll get a bogus warning.

dayglojesus:
I'll check

code/client/makepkginfo
dayglojesus: You are correct. I could weed this out according to the `installer_type` key?

gregneagle: That would probably work; return True if installer_type == 'profile'. (for bonus points also could make sure PayloadIdentifier exists, but I can't imagine makepkginfo generating a plist without that...)

code/client/makepkginfo
```diff
@@ -385,6 +385,15 @@ def check_mode(option, opt, value, parser):
...
+    """
+    Inspects compiled catalog info dictionary for valid install criteria.
+    Returns boolean.
+    """
+    results = []
+    for key in ['installs', 'receipts', 'installcheck_script']:
+        results.append(catinfo.has_key(key) and len(catinfo[key]) != 0)
+    any(results)
```
gregneagle: Bravo! Here's a Python function I've never used and never seen used. Neat.

code/client/makepkginfo
dayglojesus: Checking the `PayloadIdentifier` key is simply enough. Should it fail if that key is absent or None? Or just produce a warning?

code/client/makepkginfo
dayglojesus: Noted.

code/client/makepkginfo
gregneagle: if installer_type == 'profile' and PayloadIdentifier is absent or None the pkginfo is not a valid pkginfo. I don't think we can get to this state, and anyway, it's outside of the intention of this function.

code/client/makepkginfo
gregneagle: Just for comparison, here's how I might have written this in Python:

```python
return (catinfo.get('installs') 
        or catinfo.get('receipts')
        or catinfo.get('installcheck_script')
        or catinfo['installer_type'] == 'profile'))
```

.get returns the attribute if it exists, or None.
None and empty arrays are Falsish, so that covers it.

dayglojesus: I found `any()` by looking for the Ruby corollary `Array#any?`
```Ruby
# Ruby equivalent
a = [false, false, true]
a.any?
```
There is also `all()` Python and Ruby does `Array#all?`

gregneagle: Yes, I looked up their definitions and they appear to behave the same as their Ruby equivilents.

dayglojesus: Combining your suggestion DRYs it out a tad. What about this...?
```Python
def has_valid_install_critieria(catinfo):
    """
    Inspects compiled catalog info dictionary for valid install criteria.
    Returns boolean.
    """
    results = []
    keys = ['installs', 'receipts', 'installcheck_script', 'PayloadIdentifier']
    for key in keys:
        results.append(catinfo.get(key))
    return any(results)
```

And...

```Python
    if not has_valid_install_critieria(catinfo):
        item = catinfo['name']
        print >> sys.stderr, (
            "WARNING: %s contains no discernible install criteria!\n"
            "If you do not supply a receipt, install item, "
            "installcheck_script \n"
            "or PayloadIdentifier munki will have no way to determine \n"
            "the package's installation state. This can result in \n"
            "undesirable or unexpected behaviour (ie. looping).\n") % item
```

gregneagle: While functionally it works, it implies something that's not true: that a PayloadIdentifier is used to detect installation status in the general case. At the very least I'd remove the mention of the PayloadIdentifier from the warning.

I'm not too worried about the case of a pkginfo for a configuration profile install that is missing a PayloadIdentifier, see https://github.com/munki/munki/blob/master/code/client/makepkginfo#L122-L126

You're just using the existence of the PayloadIdentifier key as a way to tell that the pkginfo is for a config profile install, and therefore we don't need to check for the existence of any of ['installs', 'receipts', 'installcheck_script']. 

One additional note about the warning -- you spelled "behavior" wrong. :-)

```python
print >> sys.stderr, (
            "WARNING: %s contains no discernible install criteria!\n"
            "If you do not supply a receipt, install item, or"
            "installcheck_script, \n"
            "Munki will have no way to determine \n"
            "the package's installation state. This can result in \n"
            "Munki repeatedly attempting to install this item, \n"
            "or failure to attempt install at all.") % item
```

I reserve the right to further tweak the warning message, but let's proceed with the patch.

dayglojesus:
My bad. So nix the PayloadIdentifier check, update the warning message and it's good?

gregneagle:
We still need to return True from has_valid_install_critieria if the pkginfo is describing a configuration profile, and the current code in that function does that.

PR #465, opened 2015-03-23
Add AllowHTTPRedirects client option: true, false, "https"

antifuchs:
This adds an option to the munki client called AllowHTTPRedirects,
which can have three values:

* `false` (the default) - never follow redirects,
* `true` - always follow redirects,
* `"https"` - follow redirects only to HTTPS URLs.

I believe this may be what was requested in #452 - hope this is more acceptable (:

### DIFF
--- a/code/client/munkilib/fetch.py
+++ b/code/client/munkilib/fetch.py
@@ -101,10 +101,9 @@ def header_dict_from_list(array):
             header_dict[key.strip()] = value.strip()
     return header_dict
 
-
 def get_url(url, destinationpath,
             custom_headers=None, message=None, onlyifnewer=False,
-            resume=False, follow_redirects=None):
+            resume=False, follow_redirects=False):
     """Gets an HTTP or HTTPS URL and stores it in
     destination path. Returns a dictionary of headers, which includes
     http_result_code and http_result_description.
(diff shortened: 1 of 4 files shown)


antifuchs:
As per discussion on ##osx-server, the preference should represent its trinary states as a string, like `PackageVerificationMode` does. 

I've reworked the code in fetch.py and gurl.py to that effect, and added a unit test to make sure my logic checks out this time (-:

antifuchs:
Feedback in https://groups.google.com/forum/#!topic/munki-dev/ve42byxzZlU seemed fairly positive; do you think anything else is necessary to get this into a merge-able state?

tests/code/client/munkilib/gurl_redirects_test.py
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/python
+# encoding: utf-8
+"""
+gurl_redirects_test.py
+
+Unit tests for munkicommon's display_* functions.
```
gregneagle: I don't think that's what's being tested here...

code/client/munkilib/fetch.py
```diff
@@ -133,7 +131,8 @@ def get_url(url, destinationpath,
 
     options = {'url': url,
                'file': tempdownloadpath,
-               'follow_redirects': follow_redirects,
+               'must_follow_redirects': follow_redirects,
+               'allow_redirects': munkicommon.pref('AllowHTTPRedirects'),
```
gregneagle: I just don't like this complexity of two similarly-named redirect flags. Can't we just pass the right value (none, 'https', 'all') to getResourceIfChangedAtomically, and let it pass it along as well?

clburlison:
Can this be closed as #530 was merged?

antifuchs:
Yes, and thanks for the note.

gregneagle:
Thanks for the initial work on this.

PR #466, opened 2015-03-23
Add options for --ondemand_script and --OnDemand

erikng:
Example:

makepkginfo --name RepairPermissions --displayname "Repair Permissions" --description "This tool will repair your disk permissions. This is helpful if you are experiencing issues with your machine." --nopkg --pkgvers=1.0 --unattended_install --ondemand_script="RepairPermissions.sh" --OnDemand

### DIFF
--- a/code/client/makepkginfo
+++ b/code/client/makepkginfo
@@ -130,6 +130,7 @@ def getCatalogInfoForProfile(profile_path):
         cataloginfo['description'] = profile.get('PayloadDescription', '')
         cataloginfo['version'] = '1.0'
         cataloginfo['installer_type'] = 'profile'
+        cataloginfo['OnDemand'] = True
         cataloginfo['uninstallable'] = True
         cataloginfo['uninstall_method'] = 'remove_profile'
         cataloginfo['unattended_install'] = True
@@ -460,6 +461,13 @@ def main():
         p, 'Script Options',
         'All scripts are read and embedded into the pkginfo.')
     script_options.add_option(
+        '--ondemand_script', '--ondemand-script',
+        metavar='SCRIPT_PATH',
+        help=('Path to an OnDemand script to be '
+              'run as root.  '
+              'Always removed one finished. There are no receipts.')
+        )
+    script_options.add_option(
         '--installcheck_script', '--installcheck-script',
         metavar='SCRIPT_PATH',
         help=('Path to an optional installcheck script to be '
@@ -675,6 +683,12 @@ def main():
               'not listed in any applicable \'managed_installs\'.')
         )
     additional_options.add_option(
+        '--OnDemand',
+        action='store_true',
+        help=('Indicates this package should be an OnDemand package '
+              'not listed in any applicable \'managed_installs\'.')
+        )
+    additional_options.add_option(
         '--minimum_munki_version', '--minimum-munki-version',
         metavar='VERSION',
         help=('Minimum version of munki required to perform installation. '
@@ -781,6 +795,7 @@ def main():
             not options.file and
             not options.nopkg and
             not options.installer_environment and
+            not options.ondemand_script and
             not options.installcheck_script and
             not options.uninstallcheck_script and
             not options.preinstall_script and
@@ -1027,6 +1042,10 @@ def main():
         # remove minimum_os_version as we don't include it for --file only
         catinfo.pop('minimum_os_version')
 
+    if options.ondemand_script:
+        scriptstring = readfile(options.ondemand_script)
+        if scriptstring:
+            catinfo['ondemand_script'] = scriptstring
     if options.installcheck_script:
         scriptstring = readfile(options.installcheck_script)
         if scriptstring:
@@ -1060,6 +1079,8 @@ def main():
         catinfo['autoremove'] = True
     if options.minimum_munki_version:
         catinfo['minimum_munki_version'] = options.minimum_munki_version
+    if options.OnDemand:
+        catinfo['OnDemand'] = True
     if options.unattended_install:
         catinfo['unattended_install'] = True
     if options.unattended_uninstall:
--- a/code/client/makepkginfo
+++ b/code/client/makepkginfo
@@ -461,13 +461,6 @@ def main():
         p, 'Script Options',
         'All scripts are read and embedded into the pkginfo.')
     script_options.add_option(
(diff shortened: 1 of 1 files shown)


gregneagle:
There's no such thing as an "ondemand_script"... you can put one in a pkginfo file, but Munki isn't going to actually _do_ anything with it...

erikng:
Thought about this after I did the PR. 

I need to modify MSU as well or is there anything else I may be missing?

gregneagle:
1) Why do you think adding a ondemand_script is needed? Why aren't any of the existing scripts sufficient?
2) You'd have to make changes in several different places in the Munki code base: updatecheck.py and installer.py at least
3) Why do this at all, what does it add except additional complexity?
4) What happens if people add an ondemand_script but forget the OnDemand key?

and unrelated: I'd rather not change makepkginfo until more of the OnDemand functionality is finalized. Heck I even said I wasn't sure I'd stick with the name "OnDemand"...

erikng:
1. Perhaps --ondemand_script is the wrong name. We have preinstall/postinstall, installcheck and uninstall, but this is different. I know personally when I create a "nopkg" option I use the postinstall, but I'm not sure if this is really the right approach. What if an admin needs to run a preinstall/postinstall and actually run the action they need? You could create a pkg that runs a script as it's own action, but that seems more complex.
2. Okay
3. It adds complexity to munki but further enhances "nopkg". See my answer above as I feel it still stands.
4. As stated above, perhaps it should be treated as an install action. We can rename.

That's fine. I figured you would say that but that could be as simple as a find all / rename.

gregneagle:
"Perhaps --ondemand_script is the wrong name. We have preinstall/postinstall, installcheck and uninstall, but this is different."

Why is it different?

"I know personally when I create a "nopkg" option I use the postinstall, but I'm not sure if this is really the right approach."

Why isn't it? It's a script that is run during the install process. Payload-free pkgs behave similarly.

"What if an admin needs to run a preinstall/postinstall and actually run the action they need?"

How is this any different than what we have now? All the OnDemand key does is allow an end-user to install a package/run a script/copy items from a dmg over and over, whenever they want, on demand. Nothing else needs to change about how Munki handles the actual install process. OnDemand need not be limited to "nopkg" type items.

erikng:
"Why is it different?"

Maybe different is the wrong description, but if we created a new script option and recommended it's use for this, it would (probably) reduce the support questions that will inevitably come from this new feature. 

"Why isn't it? It's a script that is run during the install process. Payload-free pkgs behave similarly."

Fair point, and payload-free packages suffer from the same limitation. Typically those actions aren't complex so it doesn't matter. Some "OnDemand" actions may be more complex and require a pre/run/post.

"How is this any different than what we have now? All the OnDemand key does is allow an end-user to install a package/run a script/copy items from a dmg over and over, whenever they want, on demand. Nothing else needs to change about how Munki handles the actual install process. OnDemand need not be limited to "nopkg" type items."

I'm not saying it should be limited. --ondemand_script and --OnDemand are two separate entities.

If an admin creates a payload-free package, they can have the flexibility through munki to have a pre/post action on top of the package logic. If the admin doesn't want to go through the steps of creating a package, they are limited to pre/post only. Adding a new script will allow an admin to have up to three actions during the run without additional overhead.

gregneagle:
""Why isn't it? It's a script that is run during the install process. Payload-free pkgs behave similarly."

Fair point, and payload-free packages suffer from the same limitation."

Which is what, exactly? What new thing could you do if payload-free packages supported preinstall, install, and postinstall scripts?

"if an admin creates a payload-free package, they can have the flexibility through munki to have a pre/post action on top of the package logic. If the admin doesn't want to go through the steps of creating a package, they are limited to pre/post only. Adding a new script will allow an admin to have up to three actions during the run..."

I guess I just don't understand your thinking here. If the admin is the one writing the scripts, what can one possibly do with three script that they could not do with one single script?

The preinstall_ and postinstall_scripts were created to allow admins to use existing packages AS-IS without needing to repackage, but be able to add a task before or after install that addressed some issue. Example here: https://github.com/munki/munki/wiki/Adobe-Reader#details-for-those-who-wish-to-inflict-more-pain-upon-themselves

But if you are writing the script you wish to run, just write it the way you need it to run. No need to add extra scripts to work around issues in the "main" script.

"..without additional overhead."

Hah. Managing three scripts is more overhead than managing one.
And more overhead for me as well.

erikng:
> Which is what, exactly? What new thing could you do if payload-free packages supported preinstall, install, and postinstall scripts?

Without being pedantic, the install part section. 

>I guess I just don't understand your thinking here. If the admin is the one writing the scripts, what can one possibly do with three script that they could not do with one single script?

Fair point again, but there may be a time where splitting tasks out makes more logical sense. We are used to payload free packages, but they are odd architecturally. "Use the preinstall or postinstall script feature to run your script". Wouldn't it make more sense to use a predefined location for the script itself? Let's forget pre/post actions for now.

> The preinstall_ and postinstall_scripts were created to allow admins to use existing packages AS-IS without needing to repackage, but be able to add a task before or after install that addressed some issue. Example here: https://github.com/munki/munki/wiki/Adobe-Reader#details-for-those-who-wish-to-inflict-more-pain-upon-themselves

> But if you are writing the script you wish to run, just write it the way you need it to run. No need to add extra scripts to work around issues in the "main" script.

I see your point and I would still argue that this is even more reason to add a new variable. Recommend admins to use --newfoo_script and when they inevitably ask for help in munki-dev, we can look in the pkginfo for this value. From a functional standpoint you are correct that it technically doesn't do anything differently than a pre/post action.

>Hah. Managing three scripts is more overhead than managing one. And more overhead for me as well.

I can't claim to be an expert in the munki code, but I don't think it is that radical of a change. If you don't see enough value in it, that's fine. At the very least let's add the --whateveryouendupcallingit variable for makepkginfo.

gregneagle:
I choose not to implement additional redundant scripts. They are not needed.

gregneagle:
So amend your PR to include only the --OnDemand flag and functionality...

erikng:
I will tonight.

erikng:
Done.

gregneagle:
erikng: Please think about the issues here: https://groups.google.com/d/msg/munki-dev/KW9OeH4jtYI/hJNz_vgi7JIJ and weigh-in.

PR #473, opened 2015-05-01
follow symlinks when walking pkgs and pkgsinfo

dayglojesus:
To protect against infinite recursion, `os.walk` will not follow symlinks by
default. However, this safety feature can hamper the utility of makecatalogs if
admins want to symlink-in resources from multiple munki repos, so turn it off.

### DIFF
--- a/code/client/makecatalogs
+++ b/code/client/makecatalogs
@@ -167,7 +167,7 @@ def makecatalogs(repopath, options):
     catalogs['all'] = []
 
     # Walk through the pkginfo files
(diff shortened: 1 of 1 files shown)


dayglojesus:
Thanks @gregneagle !

szuta:
Regression in FileRepo.py.

gregneagle:
Would you mind being more verbose?

szuta:
REGRESSION IN FILEREPO.PY.

kidding - I opened #791, will update with a suggested fix and a better description.

Issue #476, opened 2015-05-05
makepkginfo: KeyError: 'installer_type' when attempting to parse payload-free pkg


dayglojesus:
The `has_valid_install_critieria` func is attempting to access keys that may not exist in the `catinfo` object and crashing as a result:

``` Bash
[15:21:59]hcc-shaggy$ makepkginfo ./iampayloadfree.pkg 
WARNING: No valid Distribution or PackageInfo found.
Traceback (most recent call last):
  File "/usr/local/munki/makepkginfo", line 1157, in <module>
    main()
  File "/usr/local/munki/makepkginfo", line 1134, in main
    if not has_valid_install_critieria(catinfo):
  File "/usr/local/munki/makepkginfo", line 397, in has_valid_install_critieria
    or catinfo['installer_type'] == 'profile'
KeyError: 'installer_type'
```

This is a quick fix, PR imminent.

PR #477, opened 2015-05-05
fix issue #476

dayglojesus:
makepkginfo was crashing when trying to access non-existent keys in the catinfo
object. Replace `[]` accessor method with `get`.


gregneagle:
Looks good.

PR #480, opened 2015-05-08
add gradient to css

clburlison:
Per Greg's comment [here](https://groups.google.com/d/msg/munki-discuss/xczxKwu9rXg/NUFRk-3VZXgJ). I have modifed the base css for Managed Software Center. 

**Before**
before

**After**
after

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center/WebResources/base.css
+++ b/code/apps/Managed Software Center/Managed Software Center/WebResources/base.css
@@ -221,9 +221,11 @@ a {
 }
 
 html, body {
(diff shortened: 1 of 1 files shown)


gregneagle:
How does this compare to the App Store? If the App Store has a similar gradient, I'm not seeing it.

screen shot 2015-05-08 at 1 04 39 pm

clburlison:
You're correct, most pages do not contain a gradient. It looks as if an App has changed the background color in the App Store a gradient is applied however this pull request doesn't make sense to apply to all pages. I should have invested farther on what the App Store is doing.

PR #484, opened 2015-06-03
builtin.super doesn't work with Cocoa classes, use objc.super instead

ccaviness:
This will allow support for more recent PyObjC installs (3.x) than what's on the system (2.x).

Using objc.super works with system python/pyobjc on both 10.9 and 10.10.

Ref: https://bitbucket.org/ronaldoussoren/pyobjc/issue/110/

### DIFF
--- a/code/client/munkilib/gurl.py
+++ b/code/client/munkilib/gurl.py
@@ -25,6 +25,9 @@ curl replacement using NSURLConnection and friends
 import os
 import xattr
 
+# builtin super doesn't work with Cocoa classes in recent PyObjC releases.
+from objc import super
+
 # PyLint cannot properly find names inside Cocoa libraries, so issues bogus
 # No name 'Foo' in module 'Bar' warnings. Disable them.
 # pylint: disable=E0611


gregneagle:
I need it to work (or at least not fail) back to OS X 10.6. Can you test?

ccaviness:
Let me see if I can wrangle up a 10.6 VM somewhere.

ccaviness:
I rigged up a 10.6(.0) Server in a VM, and this code snippet worked:
```python
from Foundation import NSObject
from objc import super

class c(NSObject):
  def init(self):
    self = super(c, self).init()

c.alloc().init()
```

gregneagle:
That's reassuring, certainly. I'd still like someone to do a full test on 10.6. I guess I can merge and ask for the community to test.

timsutton:
Aside from the 10.6 issue I just raised in #485 which is unrelated, I've run this code on 10.6 without any issues, having tested a few installations and removals using MSC and MunkiStatus.

PR #485, opened 2015-06-04
Check for a nil NSNotification.userInfo dict for 10.6

timsutton:
I found this issue introduced with 42e2cf17f1e63511bb717f0c38637b3784f0ea6e in the process of testing for #484. I'm guessing that it's ok for this whole block to not execute on 10.6, since it's only for handling data from Notification Center. This whole section of Munki is code I'm very unfamiliar with but I hope this simple check is sufficient for supporting 10.6.

The comment spells out the reasoning, though if it's preferable it could instead be done with a 10.6 OS version check is done a little further down in this method, assuming that in other OSes userInfo() doesn't return a NoneType. I tested that on Lion this issue isn't present, and I haven't ever noticed this issue on 10.8 and 10.9 VMs that have been running on master since the initial NC support was committed.

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center/MSCAppDelegate.py
+++ b/code/apps/Managed Software Center/Managed Software Center/MSCAppDelegate.py
@@ -46,13 +46,15 @@ class MSCAppDelegate(NSObject):
     def applicationDidFinishLaunching_(self, sender):
         '''NSApplication delegate method called at launch'''
         
-        userNotification = sender.userInfo().get('NSApplicationLaunchUserNotificationKey')
-        # we get this notification at launch because it's too early to have declared ourself
-        # a NSUserNotificationCenterDelegate
-        if userNotification:
-            NSLog("Launched via Notification interaction")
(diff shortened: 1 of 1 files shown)


timsutton:
Erm, a sample traceback of what was happening on 10.6 would have helped:
```
test-vm-sl:~ test$ /Applications/Managed\ Software\ Center.app/Contents/MacOS/Managed\ Software\ Center

2015-06-04 10:33:44.173 Managed Software Center[1871:903] PyObjC: Converting exception to Objective-C:
  File "/Applications/Managed Software Center.app/Contents/Resources/MSCAppDelegate.py", line 49, in applicationDidFinishLaunching_
    userNotification = sender.userInfo().get('NSApplicationLaunchUserNotificationKey')
AttributeError: 'NoneType' object has no attribute 'get'
2015-06-04 10:33:44.186 Managed Software Center[1871:903] <type 'exceptions.AttributeError'>: 'NoneType' object has no attribute 'get'
```

PR #487, opened 2015-06-10
add arguments check to criteria test

klou:
Checks for a single argument (installer item) as well as has_valid_install_critieria.  

Multiple arguments fail (with warnings) via other checks, zero arguments (script_options) pass for pkginfo generation.


gregneagle:
LGTM

PR #490, opened 2015-06-22
Update getmanifest to create unique filename for local copy of manifest

Multimac:
One thing I noticed in the way Munki handles manifests is that it uses the filename part when storing locally. This means that having multiple manifests with the same name but in different folders will cause clients to play up.

This changes the local naming scheme for manifests to allow multiple manifests with the same name, but in different folders on the server. It does this by including the relative folder portion of the path in the manifest filename.

### DIFF
--- a/code/client/munkilib/updatecheck.py
+++ b/code/client/munkilib/updatecheck.py
@@ -2635,7 +2635,7 @@ def cleanUpManifests(subdir=""):
             else:
                 os.unlink(path)
 
-    # Remove folder is empty
+    # Remove folder if empty
     try:
         os.rmdir(manifest_dir)
     except OSError:
--- a/code/client/munkilib/updatecheck.py
+++ b/code/client/munkilib/updatecheck.py
@@ -2635,11 +2635,12 @@ def cleanUpManifests():
 
     for (dirpath, dirnames, filenames) in os.walk(manifest_dir, topdown=False):
         for name in filenames:
+
+            if name in exceptions:
+                continue
+                
             abs_path = os.path.join(dirpath, name)
             rel_path = abs_path[len(manifest_dir):].lstrip("/")
-
-            if rel_path in exceptions:
-                continue
 
             if rel_path not in MANIFESTS.keys():
                 os.unlink(abs_path)
--- a/code/client/munkilib/updatecheck.py
+++ b/code/client/munkilib/updatecheck.py
@@ -2553,6 +2553,42 @@ def getmanifest(partialurl, suppress_errors=False):
       string local path to the downloaded manifest.
     """
     #global MANIFESTS
+
+    def walkManifestCache(path):
+        """Walks MANIFESTS to the specified path, creating any intermediate
+        dictionaries if they don't exist.
+
+        Returns:
+          The dictionary at the specified path.
+        """
+        branch = MANIFESTS
+
+        if path:
+            for part in path.split('/'):
+                branch = branch.setdefault(part, { })
+
+        return branch
+
+    def addToManifestCache(path, value):
+        """Sets the item at the specified path to the given value.
+        """
+        parts = os.path.split(path);
+
+        folder = walkManifestCache(parts[0])
+        folder[parts[1]] = value
+
+
+    def getFromManifestCache(path):
+        """Retrieves the item at the specified path.
+
+        Returns:
+            The item at path in MANIFESTS
+        """
+        parts = os.path.split(path);
+
+        folder = walkManifestCache(parts[0])
+        return folder.get(parts[1], None)
+
     manifestbaseurl = (munkicommon.pref('ManifestURL') or
                        munkicommon.pref('SoftwareRepoURL') + '/manifests/')
     if (not manifestbaseurl.endswith('?') and
@@ -2572,10 +2608,7 @@ def getmanifest(partialurl, suppress_errors=False):
     else:
         # request for nested manifest
         manifestdisplayname = partialurl
-        
-        valid_chars = "-_.() %s%s" % (string.ascii_letters, string.digits)
-        manifestname = ''.join(c if c in valid_chars else "_" for c in partialurl)
-
+        manifestname = partialurl
         manifesturl = manifestbaseurl + urllib2.quote(partialurl)
 
     if manifestname in MANIFESTS:
@@ -2584,6 +2617,15 @@ def getmanifest(partialurl, suppress_errors=False):
     munkicommon.display_debug2('Manifest base URL is: %s', manifestbaseurl)
     munkicommon.display_detail('Getting manifest %s...', manifestdisplayname)
     manifestpath = os.path.join(manifest_dir, manifestname)
+    manifestdir = os.path.dirname(manifestpath)
+
+    # Create the folder the manifest shall be stored in
+    try:
+        os.makedirs(manifestdir)
+    except OSError:
+        if not os.path.isdir(manifestdir):
+            raise
+
     message = 'Retrieving list of software for this machine...'
     try:
         dummy_value = getResourceIfChangedAtomically(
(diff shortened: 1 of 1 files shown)


erikng:
I'm curious as to why you are putting your manifests in folders.

Multimac:
Find it helps keep them a bit more organised.

I've set it up so there's a couple global manifests, followed by a folder per department with department-wide manifests, finally followed by a folder containing each computer's settings.

Have uploaded a screenshot of my folder structure. The printers manifest in '/manifests/' conflicts with the manifest in '/manifests/dept/it/'.

screen shot 2015-06-22 at 2 22 00 pm

erikng:
I do something similar but I've never hit this issue (actually surprised I haven't).

code/client/munkilib/updatecheck.py
```diff
@@ -2572,7 +2572,10 @@ def getmanifest(partialurl, suppress_errors=False):
     else:
         # request for nested manifest
         manifestdisplayname = partialurl
-        manifestname = os.path.split(partialurl)[1]
+        
+        valid_chars = "-_.() %s%s" % (string.ascii_letters, string.digits)
+        manifestname = ''.join(c if c in valid_chars else "_" for c in partialurl)
+
```
gregneagle: Not sure why you are doing this substitution. This could cause the the same problem you are attempting to solve:
manifests/dept1/faculty/printers and manifests/dept1_faculty/printers would be stored as the same name.

Multimac:
Yeah, very good point. Thought that may be a problem.

I've tweaked the caching method to mirror the folder structure on the server, so if a manifest is ".../manifests/dept/it/base", it will be in the same location on the client (only in "/Library/Managed Installs/manifests" instead).

code/client/munkilib/updatecheck.py
```diff
@@ -2572,15 +2608,24 @@ def getmanifest(partialurl, suppress_errors=False):
...
         # request for nested manifest
         manifestdisplayname = partialurl
-        manifestname = os.path.split(partialurl)[1]
+        manifestname = partialurl
         manifesturl = manifestbaseurl + urllib2.quote(partialurl)
 
-    if manifestname in MANIFESTS:
-        return MANIFESTS[manifestname]
+    if getFromManifestCache(manifestname)
+        return getFromManifestCache(manifestname)
```
gregneagle: Seems less than optimal to call getFromManifestCache() twice.

code/client/munkilib/updatecheck.py
```diff
@@ -2553,6 +2553,42 @@ def getmanifest(partialurl, suppress_errors=False):
       string local path to the downloaded manifest.
     """
     #global MANIFESTS
+
+    def walkManifestCache(path):
+        """Walks MANIFESTS to the specified path, creating any intermediate
+        dictionaries if they don't exist.
+
```
gregneagle: Doc string doesn't match actual behavior here

code/client/munkilib/updatecheck.py
```diff
@@ -2553,6 +2553,42 @@ def getmanifest(partialurl, suppress_errors=False):
...
+        if path:
+            for part in path.split('/'):
+                branch = branch.setdefault(part, { })
+
+        return branch
+
+    def addToManifestCache(path, value):
+        """Sets the item at the specified path to the given value.
+        """
+        parts = os.path.split(path);
```
gregneagle: Python lines don't end with a semicolon

code/client/munkilib/updatecheck.py
```diff
@@ -2553,6 +2553,42 @@ def getmanifest(partialurl, suppress_errors=False):
...
+        folder[parts[1]] = value
+
+
+    def getFromManifestCache(path):
+        """Retrieves the item at the specified path.
+
+        Returns:
+            The item at path in MANIFESTS
+        """
+        parts = os.path.split(path);
```
gregneagle: same here

code/client/munkilib/updatecheck.py
```diff
@@ -2572,15 +2608,24 @@ def getmanifest(partialurl, suppress_errors=False):
...
 
-    if manifestname in MANIFESTS:
-        return MANIFESTS[manifestname]
+    if getFromManifestCache(manifestname)
+        return getFromManifestCache(manifestname)
 
     munkicommon.display_debug2('Manifest base URL is: %s', manifestbaseurl)
     munkicommon.display_detail('Getting manifest %s...', manifestdisplayname)
     manifestpath = os.path.join(manifest_dir, manifestname)
+    manifestdir = os.path.dirname(manifestpath)
```
gregneagle: It's confusing to have manifest_dir and manifestdir

code/client/munkilib/updatecheck.py
```diff
@@ -2553,6 +2553,42 @@ def getmanifest(partialurl, suppress_errors=False):
       string local path to the downloaded manifest.
     """
     #global MANIFESTS
+
+    def walkManifestCache(path):
```
gregneagle: Not really sure why we've added these functions. Can't we just continue using the MANIFESTS dictionary, but use the full (or relative) path as the key instead of just the filename?

code/client/munkilib/updatecheck.py
```diff
@@ -2605,7 +2650,7 @@ def getmanifest(partialurl, suppress_errors=False):
         raise ManifestException(errormsg)
     else:
         # plist is valid
-        MANIFESTS[manifestname] = manifestpath
+        addToManifestCache(manifestname, manifestpath)
         return manifestpath
```
gregneagle: Probably could now also use a new cleanUpManifests() function, similar to cleanUpCatalogs, to remove any manifests from the local cache that are no longer used, including any empty subdirectories. (Be sure to leave the SelfServeManifest alone if it exists!)

Multimac:
Good point on sub directories in MANIFESTS, was trying to solve a non-existent problem. Have reverted to storing relative paths (to root manifest folder) in MANIFESTS.

Also created the cleanUpManifests function as mentioned (using cleanUpCatalogs as a base). Runs recursively to check sub directories. SelfServeManifest can't be called anything different or located anywhere other than the root manifest directory, can it?

gregneagle:
"SelfServeManifest can't be called anything different or located anywhere other than the root manifest directory, can it?" Not without breaking every Munki user who already has a SelfServeManifest...

code/client/munkilib/updatecheck.py
```diff
@@ -2608,6 +2617,39 @@ def getmanifest(partialurl, suppress_errors=False):
...
+        if cachename not in MANIFESTS.keys():
+            path = os.path.join(manifest_dir, item)
+            if os.path.isdir(path):
+                cleanUpManifests(cachename)
+            else:
+                os.unlink(path)
+
+    # Remove folder if empty
+    try:
+        os.rmdir(manifest_dir)
```
gregneagle: This would have the effect of removing the "root" manifests directory if it's empty, I think we should avoid that.

gregneagle: Might I suggest https://github.com/munki/munki/blob/master/code/client/munkilib/updatecheck.py#L2825-L2842
as a model for a slightly different implementation?

code/client/munkilib/updatecheck.py
```diff
@@ -2581,6 +2581,15 @@ def getmanifest(partialurl, suppress_errors=False):
...
     munkicommon.display_detail('Getting manifest %s...', manifestdisplayname)
     manifestpath = os.path.join(manifest_dir, manifestname)
+
+    # Create the folder the manifest shall be stored in
+    destinationdir = os.path.dirname(manifestpath)
+    try:
+        os.makedirs(destinationdir)
+    except OSError:
+        if not os.path.isdir(destinationdir):
+            raise
```
gregneagle: Why bother with the except if we are only going to raise? Either use munkicommon.display_error() to give a more helpful error message, use ManifestException to do something similar, or don't bother at all.  The exception will be raised anyway...

Multimac:
Oh, that question about the SelfServeManifest was just making sure it couldn't be placed anywhere else, in case people could do tricky things.

Good point on cleanUpManifests, I've rewritten it to use os.walk instead of recursion.

Also re-raising the os.makedirs exception is to filter out the case where the directory already exists. I've changed it to use display_error() and return None instead of re-raising the exception.

PR #492, opened 2015-06-29
Move paths.d to core package

erikng:


gregneagle:
I think it might be better/safer/less likely to trigger future issues if _both_ packages contained the paths.d file.

Without that, due to how package upgrade logic works, if someone were to install the new admin pkg without installing the new core pkg (don't ask why someone might do that; I cease to be surprised at unexpected things people do) the paths.d file would be removed.

erikng:
I originally was going to do that.

Added. Also note that I changed the chown path from /private to /private/etc/paths.d

Neither would be flagged on El Cap's installer, but I figured it would be better anyway.

code/tools/make_munki_mpkg.sh
```diff
@@ -341,10 +346,11 @@ done
 # Set permissions.
 chmod -R go-w "$ADMINROOT/usr/local/munki"
 chmod +x "$ADMINROOT/usr/local/munki"
+
 # make paths.d file
 mkdir -p "$ADMINROOT/private/etc/paths.d"
 echo "/usr/local/munki" > "$ADMINROOT/private/etc/paths.d/munki"
-chmod -R 755 "$ADMINROOT/private"
+chmod -R 755 "$ADMINROOT/private/etc/paths.d"
```
gregneagle: We're setting the mode on the files/directories we've created in our $ADMINROOT. Now the mode of private and etc are unclear. What are they?

erikng: I guess that's a valid point. While they are the same permissions, it might be assumed that someone could have changed these values.

PR #494, opened 2015-07-06
Change LastNotifiedstring default to ''

oliof:
https://github.com/oliof/munki/blob/master/code/client/managedsoftwareupdate#L407 checks for the boolean value of lastNotifiedString. In the original version, the result would be "1970-01-01 +00:00:00" which always resolved to True in this check, making everything in lines 408 and below effectively dead code. Returning an empty string by default retains type behavior of LastNotifiedString and fixes this boolean check.

### DIFF
--- a/code/client/munkilib/munkicommon.py
+++ b/code/client/munkilib/munkicommon.py
@@ -1203,7 +1203,7 @@ def pref(pref_name):
         'AppleSoftwareUpdatesOnly': False,
         'SoftwareUpdateServerURL': '',
         'DaysBetweenNotifications': 1,
-        'LastNotifiedDate': NSDate.dateWithTimeIntervalSince1970_(0),
+        'LastNotifiedDate': '',
         'UseClientCertificate': False,
         'SuppressUserNotification': False,
         'SuppressAutoInstall': False,


gregneagle:
LastNotifiedDate is a date and not a string. Setting it to a string would be wrong. When I have access to a computer with Internet access I'll read through this and see if I can better understand the overall context.

oliof:
According to my reading, t's a string representation of a date, and setting
it to epoch(0) is misleading and breaks the logic that checks it as a
boolean. This was the smallest change to make existing code work, but I am
happy to make amends. I could for example make the test in
managedsoftwareupdate test whether the date returned is midnight January
1st 1970, or remove the default value and catch the resulting TypeError,
but either of these options seem more convoluted (whereas my approach
exploits a peculiarity of the type system).

2015-07-06 19:44 GMT+02:00 Greg Neagle <notifications@github.com>:

bochoven:
It's definitely a date:

```xml
        <key>LastNotifiedDate</key>
        <date>2015-07-01T08:37:24Z</date>
```

Not sure what you're trying to fix, this code has been in munki since 2009.

gregneagle:
I think perhaps you are thinking `if lastNotifiedString:` is a boolean comparison, when it's probably closer to being intended as "if there is a valid value in lastNotifiedString"...

I'm not sure why you say "making everything in lines 408 and below effectively dead code."

As long as lastNotifiedString doesn't have a value that evaluates as False, 0 or empty, the code will execute.

oliof:
Sorry, I was explaining it the wrong way around, it's *undead code*: lastNotifiedString will _always_ have a valid value: If it's not in the preferences file, it's default will set it to a date string ('Jan 1 1970 +00:00:00'), if it is in the preferences file, it will be a date (written by managedsoftwareupdate). The only way this could be a False, 0 or empty string is if some third party would write the preference.

gregneagle:
Ok: it's still not clear to me what undesired behavior it is you are trying to fix.

Issue #445, opened 2015-01-21
Certificate Bundles - Munki Certificate Issue after changing fro cURL to NSURL


ltinker:
certificate import process using the ca.pem file needs to import certificate bundles and not just the first certificate in the file. this is was appeared to occur when using cURL

**What steps will reproduce the problem?**
1. requiring client certificate authentication on Munki Server
2. Having a CA Chain in Managed Installs/cert/ca.pem (structure root CA then intermediate CA)
3. running Munkitools 2.1.1.2353 on client machine
4. running managedsoftwareupdate --checkonly

**What is the expected output?**

Managed Software Update Tool
Copyright 2010-2014 The Munki Project
https://github.com/munki/munki

Starting...
    Performing preflight tasks...
    preflight stdout: 
        -- Executing: /usr/local/munki/preflight.d/disableUpdates --
        Disable update check requires run type. Assuming enabled and exiting...

        -- Executing: /usr/local/munki/preflight.d/munkiwebadmin_preflight --
        Preflight report submmitted for mac01.edu.au.
    Completed creation of client keychain at /Library/Managed Installs/Keychains/munki.keychain
Checking for available updates.

**What do you see instead?**

Managed Software Update Tool
Copyright 2010-2014 The Munki Project
https://github.com/munki/munki

Starting...
    Performing preflight tasks...
    preflight stdout: 
        -- Executing: /usr/local/munki/preflight.d/disableUpdates --
        Disable update check requires run type. Assuming enabled and exiting...

        -- Executing: /usr/local/munki/preflight.d/munkiwebadmin_preflight --
        Preflight report submmitted for mac01.edu.au.
    Completed creation of client keychain at /Library/Managed Installs/Keychains/munki.keychain
Checking for available updates...
2015-01-21 12:31:18.943 Python[49605:2b23] CFNetwork SSLHandshake failed (-9831)
2015-01-21 12:31:18.944 Python[49605:2b23] NSURLConnection/CFURLConnection HTTP load failed (kCFStreamErrorDomainSSL, -9831)
ERROR: Could not retrieve manifest mac01 from the server: Error -1205: The server “munkiprod.edu.au” did not accept the certificate.
ERROR: Could not retrieve managed install primary manifest.

**What version of the Munki tools/InstallOSpkg tools are you using? On what
version of OS X?** running OS X 10.9.5 with Munkitools 2.1.1.2353 

**Please provide any additional information below.**

from what i can see,
[...]

PR #500, opened 2015-07-29
Adjust the import options passed to security based on certificate types

jlrgraham:
The default options passed to `security` will fail if the certificate isn't self signed.  This patch alters the import behavior to check each certificate file for the issuer and subject.  If they match the certificate is self signed and imported as usual.  If the certificate isn't self signed then pass "-r unspecified" to `security` to allow the import of an intermediate.

The import will still fail if the user specifies a single file that contains both the root and the intermediates(s).  So users would need to specify the SoftwareRepoCAPath option and then place each certificate in a separate file there.

This is would likely be a solution for issue #445.

### DIFF
--- a/code/client/munkilib/keychain.py
+++ b/code/client/munkilib/keychain.py
@@ -27,6 +27,7 @@ and here:
 import base64
 import hashlib
 import os
+import re
 import subprocess
 
 import munkicommon
@@ -88,6 +89,29 @@ def pem_cert_sha1_digest(cert_path):
     except BaseException, err:
         munkicommon.display_error('Error reading %s: %s' % (cert_path, err))
         return None
+
+
+def is_x509_cert_self_signed(cert_path):
+    '''Test for matching certificate issuer and subject to identify
+    self signed certificates.'''
+    openssl_output = openssl(
+        'x509', '-in', cert_path, '-noout', '-issuer', '-subject'
+    )
+
+    properties = {}
+    property_match = re.compile(r'^([^=]+)= (/.*)')
+    for line in openssl_output.split('\n'):
+        match = property_match.search(line)
+        if match:
+            properties[match.group(1)] = match.group(2)
+
+    if 'subject' in properties and 'issuer' in properties:
+        if properties['subject'] == properties['issuer']:
+            return True
+        else:
+            return False
+
+    raise OpenSSLError('Unable to identify issuer and subject of certificate.')
 
 
 def get_munki_server_cert_info():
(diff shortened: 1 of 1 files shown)


gregneagle:
Any reason to not just always pass `r unspecified` ? There's lots here I simply do not understand.

jlrgraham:
If `-r unspecified` is used with a self-signed/root certificate then it will be imported into the keychain but not marked as trusted.  This would invalidate any intermediate certificate authorities as well.

gregneagle:
So can we pass additional options to mark it as trusted?

gregneagle:
If your Munki server has a cert connected to an internal CA, presumably you are using the internal CA for other things. In that case, wouldn't you have already configured System.keychain with the correct [root, intermediate] CA certs? And if so, then you don't need to provide anything to Munki at all.

jlrgraham:
The default resultType for `add-trusted-cert` is `trustRoot`, so your existing implementation of the import will work without issue for self-signed/root certificates.  It's just the intermediate CAs where the option needs to be different.

jlrgraham:
You are correct regarding my CAs; I have a custom root with several intermediates and those are installed/configured as part of the deployment process.

Just wanted to supply some code that would make the internal CA import a bit more flexible.  If you prefer not not alter the security related code I understand completely and will close the pull request.

gregneagle:
I'd rather leave it alone since I barely understand all the ramifications of any change to this highly-complex stuff. But I am having others look at it.

pudquick:
Coming back to this since @gregneagle asked me to look it over.

Just writing some additional "theory of munki HTTPS" here.

When munki moved from curl to OS X's own APIs for HTTP and HTTPS communications, we picked up some benefits - namely automatic trust for HTTPS connections that Safari would trust, whereas with curl you still had to manually configure much of the CA trust / provide certificates locally on the client. Certificates purchased from an outside vendor generally "just work".

With this simplification, it became possible to focus directly on the chain of trust with munki: ideally, if you're importing a CA cert, you're importing the outermost CA that OS X does not natively trust on its own.

But yes, intermediate CAs pose a configuration question - how to deal with them?

For this, I'm going to take a page from general web hosting configuration guidelines: intermediate CAs should be included in webserver-side configuration, not client-side.

If you purchase a certificate from a CA vendor and their CA is not part of the root CAs shipped with the major operating systems, they provide you with a CA certificate bundle and configuration information on how to make sure that your website, in addition to offering up your personal certificate, offers up their intermediate CA. A client machine can't determine if it trusts a certificate if it can't see all of the certificates in the chain - or at least enough of the chain to reach a trusted certificate.

The standard "best practice" is that your webserver (in this case your munki repo) offer every intermediate certificate necessary to reach a certificate that's implicitly trusted by a visiting client.

This is why most of the major CA cert vendors have extensive documentation about this procedure:

https://www.godaddy.com/help/installing-an-ssl-certificate-in-microsoft-iis-7-4801
https://www.digicert.com/ssl-certificate-installation-apache.htm
https://support.globalsign.com/customer/portal/articles/1225234-install-certificate---apache-http-server

etc.

Following this idea, this would mean munki covers most of the common HTTPS configurations: self-signed server cert, internal CA issued cert, external CA issued cert, external intermediate CA issued cert

The one outlier is: internal intermediate CA issued cert (unfortunately the scenario here).

But ... if you're doing an internal CA and internal intermediate CA(s), it's not terribly common to have a workflow where you receive a single bundled .pem file (like you would if you purchase a cert externally).

I would **hope** it's just as easy to obtain the primary internal CA cert (which would be used with munki) and then the intermediates separately (which would be used webserver-side).

For being the configuration that's "most rare" in HTTPS scenarios, this pull as written includes a rather heavy change: the dependency on /usr/bin/openssl to parse the certificate structure.

So far munki has been able to avoid relying on the openssl binary directly. For the time being, I'd like to keep it that way.

PR #501, opened 2015-07-29
AllowInsecureConnections preference

keeleysam:
this allows an admin to disable the injection if the ATS behavior is desired

### DIFF
--- a/code/client/munkilib/munkicommon.py
+++ b/code/client/munkilib/munkicommon.py
@@ -1208,7 +1208,8 @@ def pref(pref_name):
         'SuppressUserNotification': False,
         'SuppressAutoInstall': False,
         'SuppressStopButtonOnInstall': False,
-        'PackageVerificationMode': 'hash'
+        'PackageVerificationMode': 'hash',
+        'AllowInsecureConnections': True
     }
     pref_value = CFPreferencesCopyAppValue(pref_name, BUNDLE_ID)
     if pref_value == None:
--- a/code/client/munkilib/appleupdates.py
+++ b/code/client/munkilib/appleupdates.py
@@ -319,13 +319,9 @@ class AppleUpdates(object):
             "User-Agent: managedsoftwareupdate/%s Darwin/%s (%s) (%s)"
             % (machine['munki_version'], darwin_version,
                machine['arch'], machine['machine_model']))
-
-        allow_insecure_connections= munkicommon.pref('AllowInsecureConnections')
-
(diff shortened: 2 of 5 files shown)


gregneagle:
Some questions:
1) Is this the right name for the preference?
2) Should munkicommon.pref return a default value of True instead of None?
3) Will you write documentation?

keeleysam:
1. I changed it to `AllowInsecureConnections` which should be better in the long run if munki ever defaults to only run on secure connections in the future or has to switch away from NSURLConnection.
2. It shouldn't need to, setting the key to true or not setting it does the same thing.  The only current action is if the key is set to false.
3. Yes, no problem.

keeleysam:
Alright this should be much simpler now and gurl no longer has ties to munkicommon.

code/client/munkilib/munkicommon.py
```diff
@@ -1208,7 +1208,8 @@ def pref(pref_name):
         'SuppressUserNotification': False,
         'SuppressAutoInstall': False,
         'SuppressStopButtonOnInstall': False,
-        'PackageVerificationMode': 'hash'
+        'PackageVerificationMode': 'hash',
+        'AllowInsecureConnections': True
```
keeleysam: I do think that this should be changed to false in a future release with ample warning.

natewalck: +1

Once there is thorough documentation on setting up munki using SSL and appropriate configurations server side, this would be good to have.

natewalck:
I was talking with Mike Dodge and he had a good idea that would prevent people's environments from getting completely broken if we ever decide to make it HTTPS by default.

We could have a preinstall script that looks at the munki config and if any of the URLs are HTTP and the override is not present, it would bail the whole installer. Perhaps there could also be an installer option that bypasses this check.

This would prevent people who use HTTP from breaking their environments because they didn't read the release notes and didn't test sufficiently. 

Then, if they *really* want HTTP and understand the ramifications, they can still do so, but it makes it harder to just use HTTP because you do not know any better.

gregneagle:
I am considering a code refactor that would address the ATS issue in a completely different (read non-hacky) way. If I go that route, enabling/disabling ATS protections would require a change to the Info.plist of an application bundle as God and Apple intended, making this preference non-functional. I think I'm just going to park this for a bit. (Apple needs to provide a real API for ATS...)

keeleysam:
moving to #563

PR #502, opened 2015-07-31
only create Managed Software Update.app symlink if it previously existed

keeleysam:
Currently the Munki installer creates a symlink from `/Applications/Utilities/Managed Software Update.app` to `/Applications/Managed Software Center.app`.  This is good for those upgrading from Munki 1, but doesn't need to happen for fresh installs.

### DIFF
--- a/code/pkgtemplate/Scripts_app/preinstall
+++ b/code/pkgtemplate/Scripts_app/preinstall
@@ -1,13 +1,11 @@
 #!/bin/sh
 
-# remove any Managed Software Update.app bundle since the installer
-# won't replace a bundle with a symlink, leading to yucky stuff like
-# "/Applications/Utilities/Managed Software Update.app 1"
+# Replace any Managed Software Update.app bundle with a symlink to Managed
+# Software Center.app
 
 if [ -d "$3/Applications/Utilities/Managed Software Update.app" ] ; then
     /bin/rm -r "$3/Applications/Utilities/Managed Software Update.app"
-fi
-
-if [ -d "$3/Applications/Utilities/Managed Software Update.app 1" ] ; then
-    /bin/rm -r "$3/Applications/Utilities/Managed Software Update.app 1"
+    # make a symlink for the old MSU.app
+	/bin/ln -s "$3/Applications/Managed Software Center.app" \
+	 "$3/Applications/Utilities/Managed Software Update.app"
 fi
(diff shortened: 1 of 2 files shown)


erikng:
This is great as launchpad is very finicky with this symlink.

gregneagle:
I originally tried an approach like this. Here are the problems:

The pkg identifier is "com.googlecode.munki.app" for both the version 1 (and earlier) pkg that installs MSU.app and the version 2 (and later) pkg that installs MSC.app.

This triggers the "upgrade" behavior in installer: /Applications/Utilities/Managed Software Update.app exists in version < 2 of the package. But in version 2+ of the pkg, this app bundle has been replaced with a symlink. Installer refuses to replace a bundle (or perhaps a directory) with a symlink, so it _renames_ "/Applications/Utilities/Managed Software Update.app" to "/Applications/Utilities/Managed Software Update.app 1", then installs the symlink.

That's why you see the odd cleanup code in the original preflight removing "/Applications/Utilities/Managed Software Update.app 1" if it exists -- to clean up after some of my earlier packages that caused this issue.

So now I remove the "/Applications/Utilities/Managed Software Update.app" if it exists so that the Installer can just install the symlink.

What I think will happen with your change is this, when upgrading from pre-2 to a version containing your change:

You remove the "/Applications/Utilities/Managed Software Update.app" and create a symlink in its place.
The upgrade behavior sees that "/Applications/Utilities/Managed Software Update.app" existed in the prior version of the pkg, but doesn't exist in the current version, so installer will remove the symlink you just created.

Please test the various scenarios around this. I'll think you'll find they don't always end up with the expected result.

keeleysam:
I will figure something out here and test thoroughly.

keeleysam:
I've tested this extensively, and under no circumstance does the symlink go missing or the machine ends up with `/Applications/Utilities/Managed Software Update.app 1`.

1.0.0.1896 to 2.2.4.2518: symlink created by script
1.0.0.1896 to 2.2.4.2431 to 2.2.4.2518: symlink created by second package, replaced by script
2.2.4.2431 to 2.2.4.2518: symlink created by package, replaced by script
2.2.4.2518: symlink never created

Bouncing around between versions, even going back to 1.0.0.1896 and upgrading again to 2.2.4.2518 always ends up with the right thing.  Reinstalling over and over also doesn't create that second symlink or destroy one that shouldn't exist.

Here's the installer I'm using: https://www.dropbox.com/s/mnnos86ypeil2ty/munkitools-2.2.4.2518.pkg?dl=0

erikng:
Is there a possibility of creating another case where if certain pkgs were never installed, go ahead and remove the symlink?

IE remove symlink if 1.0 never existed on machine.

gregneagle:
How would one determine that 1.0 never existed on the machine?

gregneagle:
(If you just want it _gone_, a postinstall_script in the pkginfo could handle this)

erikng:
Package receipts.

erikng:
I was just asking. :)

gregneagle:
Nope -- that only tells you what is currently installed. It does not give a complete history.

PR #506, opened 2015-08-16
Allows gURL to send post data

grahamgilbert:
This is the modification I made to gurl.py to allow Imagr to send POST data. This would probably be useful for pre and postflight scripts to have available to them.

### DIFF
--- a/code/client/munkilib/gurl.py
+++ b/code/client/munkilib/gurl.py
@@ -33,7 +33,7 @@ from objc import super
 # pylint: disable=E0611
 from Foundation import NSBundle
 from Foundation import NSRunLoop, NSDate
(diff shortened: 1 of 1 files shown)


gregneagle:
I'll let you two boys has this one out...

grahamgilbert:
@bochoven if you want to do that, I'd be happy to have that in over my PR. This would do what I need for Imagr though.

bochoven:
I'm ok with merging this, I might add the possibility to get the output directly in a later stage.. For munkireport I cannot count on people being up-to-date with the latest munki anyway.

grahamgilbert:
Closing this - on second thoughts I don't think this should be part of Munki until Munki itself needs it. Third party tools should be self contained when possible.

PR #507, opened 2015-08-17
preupgrade_alert support

erikng:
This allows an admin to display an alert prior to upgrading a software package.

Useful for COSXIP upgrades.


gregneagle:
Looks OK to me. I probably won't merge until after the 2.3 official release.

jessepeterson:
Howdy. Been looking into this feature and having some difficulty. :sob:

1. The biggest issue is probably that when on the "Updates" tab of MSC where the "Other available updates" (suggestions, for lack of a better term) are listed: when attempting to hit the little + button the `preupgrade_alert` doesn't seem to trigger. It does seem to trigger on the normal Software tab browsing through the listing of software. Does this seem like intended behavior? This seems like the most important place to to have these alerts/warning to stave off users accidentally updating based off the suggestion that Munki is asking of users.
2. The wiki in [Supported-Pkginfo-Keys](https://github.com/munki/munki/wiki/Supported-Pkginfo-Keys) nor [here](https://github.com/munki/munki/wiki/Pkginfo-Files#preinstall_alertpreuninstall_alert) aren't yet updated to reflect that these keys/options exist. I'm happy to do so if it's confirmed that they're now generally available.
3. Very minor issue (and honestly probably too late now) the keys are named `upgrade` but everywhere throughout Munki it seems the word `update` is used. Is there an appreciable difference?

Thanks!

erikng:
Jesse,

I have noticed some odd behaviors with all alerts. Eric Holtan and I found the issues exist from their first introduction. I nor Bart have had time to figure out where exactly the issue is, but I was forced for now to not use any alerts.

jessepeterson:
Thanks for jumping in @erikng. Is it articulated somewhere what the exact "odd behaviors" you and Eric Holtan are seeing?

erikng:
https://github.com/munki/munki/issues/517

gregneagle:
@jessepeterson: Please see https://github.com/munki/munki/commit/8fc18792c846bd45f72547192a8de53030234c04

PR #510, opened 2015-08-19
only create Managed Software Update.app symlink if it previously existed

keeleysam:
Resolves the issue from the last commit of #502


keeleysam:
@gregneagle would you consider re-merging this?  it behaves properly since the command `mkdir "$APPROOT/Applications"` actually exists now.

keeleysam:
closing, to be part of #563

Issue #512, opened 2015-08-31
manifestutil 'find' only returns first instance of matching criteria per section per manifest


poundbangbash:
When using the 'find' command in manifestutil I noticed that it only reports back the first instance in each section for each manifest (catalogs, managed_installs, etc).

The below example uses just pkgs to illustrate:

```
> new-manifest testFind

> add-pkg my2016 --manifest testFind
WARNING: Package my2016 is not available in catalogs [] of manifest testFind.
Added my2016 to section managed_installs of manifest testFind.

> add-pkg your2016 --manifest testFind
WARNING: Package your2016 is not available in catalogs [] of manifest testFind.
Added your2016 to section managed_installs of manifest testFind.

> display-manifest testFind
 catalogs:
 included_manifests:
 managed_installs:
     my2016
     your2016
 managed_uninstalls:

> find 2016
testFind (managed_installs): my2016
1 items found.
```

I've confirmed this behavior in version 2.3.0.2519.

-Eric

poundbangbash:
This has been merged in #513

PR #513, opened 2015-09-01
Remove find() breaks

poundbangbash:
Removing the breaks in the find function to allow multiple matches per section per manifest.
Addresses issue https://github.com/munki/munki/issues/512

Before the edits:

```
> find 2016 
testFind (managed_installs): my2016
1 items found.
> find 2016 --section managed_installs
testFind: my2016
1 items found.
```

After the edits:

```
> find 2016
testFind (managed_installs): my2016
testFind (managed_installs): your2016
2 items found.
> find 2016 --section managed_installs
testFind: my2016
testFind: your2016
2 items found.
```

### DIFF
--- a/code/client/manifestutil
+++ b/code/client/manifestutil
@@ -424,7 +424,7 @@ def find(args):
                     print '%s (%s): %s' % (name, key, value)
                     count += 1
 
-    print '%s items found.' % count
+    print '%s matches found.' % count
     return 0
 
 
--- a/code/client/manifestutil
+++ b/code/client/manifestutil
@@ -404,7 +404,6 @@ def find(args):
                             if findtext.upper() in item.upper():
                                 print '%s: %s' % (name, item)
                                 count += 1
-                                break
                         except AttributeError, err:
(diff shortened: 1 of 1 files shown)


gregneagle:
Will that confuse people? It says "2 items found" when it found only one item (that had the search string twice) I work with people who would be confused by that.

poundbangbash:
I guess it depends on what you're looking to find; a manifest or items in the manifest.
There are two found things listed.  I'm not sure how that would get confused.

With the breaks in the code the find doesn't find the full set of data, only a subset.  If I'm looking for manifests with all things 2016 where one manifest has Excel2016 only and one has Excel2016 and Word2016, I wouldn't know the second had Word2016 as Excel2016 would be the only hit in that section for that manifest.  I'd think complete information is more beneficial than a subset.

gregneagle:
So change the results text to "2 matches found" to reflect that.

poundbangbash:
Updated the find() output verbiage to better relay what it found.

PR #533, opened 2015-10-28
Log warning when "Products" key not found in filtered Apple catalog.

juofmgoblue:
Log warning when "Products" key is not found in filtered Apple Update catalog

### DIFF
--- a/code/client/munkilib/appleupdates.py
+++ b/code/client/munkilib/appleupdates.py
@@ -330,6 +330,8 @@ class AppleUpdates(object):
         of these files."""
         catalog = FoundationPlist.readPlist(self.filtered_catalog_path)
         if not 'Products' in catalog:
+            munkicommon.display_warning(
+                '"Products" not found in %s', self.filtered_catalog_path)
             return
 
         for product_key in catalog['Products'].keys():


gregneagle:
Any context here? Why that key specifically?

juofmgoblue:
Context here: https://groups.google.com/forum/#!topic/simian-discuss/2DiGacqWcPU
Which I've migrated here: https://groups.google.com/forum/#!topic/munki-dev/hlwEfLol-Fw

TL;DR:  admin is reporting the following error, and I'm adding logging to help debug: "ERROR: Missing local Software Update catalog at /tmp/munki_swupd_cache/content/catalogs/local_install.sucatalog"

As for the preexisting check for "Products" key in the first place, I forget why that's present, but I'm guessing it's just confirming that the catalog is an appropriately formatted Apple Update catalog?

gregneagle:
If there are no Products in the catalog there's not much point in proceeding, no...

There were lots of issues late last week with Apple's sucatalogs -- are you sure this person's problem wasn't caused by that and has cleared up now?

BTW, Munki's softwareupdate support in El Cap completely ignores the local file:// catalogs since El Cap's softwareupdate errors if you set CatalogURL to a file:// URL.

juofmgoblue:
They're still experiencing the problem, but I suppose that could be due to Simian's catalog caching.  That said, Simian is parsing the catalogs to pull out product data, and completely regenerating the catalog from scratch (ala Reposado), so I'm not even sure if such issues would be realized all the way through.

It's doubtful the Products check will show anything, but filling gaps in logging coverage is prudent regardless.  The only other place I see this method exist without logging is if the user requested such: https://github.com/munki/munki/blob/master/code/client/munkilib/appleupdates.py#L380   Both the "real" cache dir (in /L/MI) and the symlink are created in init, with logging upon failure, so I can't think why else either may not exist: https://github.com/munki/munki/blob/master/code/client/munkilib/appleupdates.py#L155

I'm guessing (but haven't confirmed) they're not running on El Cap, otherwise this issue would not likely arise, as you note.  That said, endless fun keeping up with OS X releases!

Issue #536, opened 2015-11-10
ManifestName in ManagedInstallReport.plist should be reported as after HTTP redirect


mkuron:
With the HTTP redirect support introduced in 9da705a/#530, it would make sense if `getPrimaryManifest` in updatecheck.py would set the `ManifestName` in _/Library/Managed Installs/ManagedInstallReport.plist_ based on the name of the file redirected to. Right now, it sets it to the ClientIdentifier (e.g. long host name, short host name, serial number, `site_default`) used to request the manifest before the redirect.

I have a PHP script that dynamically redirects clients to different manifests, based on their short host name. In MunkiWebAdmin, which reads the `ManifestName` field in the ManagedInstallReport, I always see the short host name reported as the client's manifest name -- obviously, clicking it leads nowhere because the real manifest name is different.

gregneagle:
I'm not using redirects or a PHP script to dynamically redirect clients to different manifests, so writing and testing code to provide the feature you ask for is 1) inconvenient for me, and 2) not currently useful to me. I'd be happy for someone else with the proper pieces in place to do it: pull requests accepted.

PR #537, opened 2015-11-13
Set User-Agent header in HTTP requests

mkuron:
Right now Munki only sets a User-Agent header on requests made from the appleupdates.py code. Everything else (catalog, manifest, package downloading) reports something like `"Python/2.7.5 CFNetwork/673.6 Darwin/13.4.0 (x86_64) (VMware7%2C1)"` to the server. For certain server-side stuff, it can be useful to have something like `managedsoftwareupdate/2.3.1.2535` prepended.

### DIFF
--- a/code/client/munkilib/fetch.py
+++ b/code/client/munkilib/fetch.py
@@ -25,6 +25,7 @@ Created by Greg Neagle on 2011-09-29.
 import calendar
 import errno
 import os
+import sys
 import re
 import shutil
 import subprocess
@@ -39,6 +40,7 @@ import munkicommon
 from gurl import Gurl
 
 from Foundation import NSHTTPURLResponse
+import FoundationPlist
 
 
 # XATTR name storing the ETAG of the file when downloaded via http(s).
@@ -97,8 +99,10 @@ def header_dict_from_list(array):
     header_dict = {}
     machine = munkicommon.getMachineFacts()
     darwin_version = os.uname()[2]
-    header_dict["User-Agent"] = ("managedsoftwareupdate/%s Darwin/%s (%s) (%s)"
-        % (machine['munki_version'], darwin_version,
+    python_version = "%d.%d.%d" % (sys.version_info[0],sys.version_info[1],sys.version_info[2])
+    cfnetwork_version = FoundationPlist.readPlist("/System/Library/Frameworks/CFNetwork.framework/Resources/Info.plist")['CFBundleShortVersionString']
+    header_dict["User-Agent"] = ("Python/%s CFNetwork/%s managedsoftwareupdate/%s Darwin/%s (%s) (%s)"
+        % (python_version, cfnetwork_version, machine['munki_version'], darwin_version,
            machine['arch'], machine['machine_model']))
 
     if array is None:
--- a/code/client/munkilib/fetch.py
+++ b/code/client/munkilib/fetch.py
@@ -93,8 +93,15 @@ def writeCachedChecksum(file_path, fhash=None):
 def header_dict_from_list(array):
     """Given a list of strings in http header format, return a dict.
     If array is None, return None"""
+    header_dict = {}
+    machine = munkicommon.getMachineFacts()
+    darwin_version = os.uname()[2]
+    header_dict["User-Agent"] = ("managedsoftwareupdate/%s Darwin/%s (%s) (%s)"
+        % (machine['munki_version'], darwin_version,
+           machine['arch'], machine['machine_model']))
+
     if array is None:
-        return array
+        return header_dict
     header_dict = {}
     for item in array:
         (key, sep, value) = item.partition(':')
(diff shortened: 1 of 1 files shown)


code/client/munkilib/fetch.py
```diff
@@ -93,8 +93,15 @@ def writeCachedChecksum(file_path, fhash=None):
 def header_dict_from_list(array):
     """Given a list of strings in http header format, return a dict.
     If array is None, return None"""
```
RichardLafontaine: that isn't the case anymore...

mkuron:
Thanks, fixed

code/client/munkilib/fetch.py
```diff
@@ -92,9 +92,17 @@ def writeCachedChecksum(file_path, fhash=None):
...
+    machine = munkicommon.getMachineFacts()
+    darwin_version = os.uname()[2]
+    header_dict["User-Agent"] = ("managedsoftwareupdate/%s Darwin/%s (%s) (%s)"
+        % (machine['munki_version'], darwin_version,
+           machine['arch'], machine['machine_model']))
+
     if array is None:
-        return array
+        return header_dict
     header_dict = {}
```
gregneagle: So if array is not None, you throw away the User-Agent header? Is that your intent?

gregneagle:
"For certain server-side stuff, it can be useful to have something like managedsoftwareupdate/2.3.1.2535 prepended."  Can you elaborate?

mkuron:
I'm preparing some code for #536. As redirects require 2.4.0 or higher, my server-side PHP script will need to fall back to including the manifest instead of redirecting to it for Munki 2.3 and lower. Right now, there's no way for the PHP script to detect the Munki version.

I just fixed the he `array is None` mistake.

gregneagle:
Why add additional complexity to your PHP script? Why not just always return the desired manifest instead of redirecting to it in some cases?

mkuron:
#536: I want the ManagedInstallReport to show the manifest that was actually used, as opposed to the computer's host name. If I don't do that, MunkiWebAdmin (or other reporting solutions) won't be able to make the connection between the computer and the manifest, unless I added the same computer to manifest mapping logic to the reporting.

gregneagle:
While I have no objections to this change, I still think it should go up for quick discussion on munki-dev in case a change to the user-agent causes assumptions other Munki admins have made to break.

(IOW, if this was opt-in, I would not worry about it at all; but since you are changing the user-agent for everyone, I think it only polite to give a heads-up)

keeleysam:
I think it would be fine to append the `managedsoftwareupdate/VERSION` component, but as written it also removes the current Python and CFNetwork components.  Could you change this so it just adds to the UA instead of completely replacing it?

gregneagle:
I think Sam has a good point; it's possible that other Munki admins are using the current user-agent for something. Ideal would be to read the user-agent being set by the NSURLConnection code and then just add the managedsoftwareupdate/VERSION bit.

mkuron:
Sounds like a good idea.
Unfortunately, it seems like there is no way to get the default user agent from an NSURLConnection (besides running your own webserver in a separate thread, making a request, and getting the user agent from there). As there doesn't appear to be an API to get the CFNetwork version, I'll have to read `CFBundleShortVersionString` from /System/Library/Frameworks/CFNetwork.framework/Resources/Info.plist and get the Python version from `sys.version_info`. Going this route essentially freezes the structure of the user agent the way OS X does it today. If they ever change it, we'll either have to make that same change in Munki as well, or stick to the old structure for backwards-compatibility.

mkuron:
Ok, changed that.

keeleysam:
I thought about this for a bit and I think it would probably be more useful to send an actual HTTP header instead of messing with the user agent.  We already have the code to send additional headers that are added by the key `AdditionalHttpHeaders` in Munki's configuration, and if we sent a header such as `X-Munki-Version: 2.4.5223` it would also be easier to parse on the server side if that's what you want to accomplish.

Adding it as an actual header would also presumably not break anybody who currently does user agent parsing.

mkuron:
Well, that's the documented purpose of the User-Agent header: to tell the server about the user agent, i.e. the software making the request.
Did you find anybody who actually parses the old header to get information about the client? The CFNetwork and Python version numbers don't uniquely identify an OS X version anyway and thus are probably useless for everyone, and the Darwin version and architecture will continue to be at the end of the string. So if someone used a regex to extract the Darwin info from the string, it will continue work.

Also, we already broke backwards compatibility of the User-Agent header a year ago when we switched from curl to the NSURL stuff. The change proposed here is much less intrusive. The current User-Agent header is too much of an implementation detail that anybody should have ever relied on its contents: OS X could have at any point decided to change its contents or even drop it entirely.

gregneagle:
"Did you find anybody who actually parses the old header to get information about the client?"  Did anybody ask on munki-dev?

gregneagle:
(I think a user-agent that contained munki version and Darwin version would be fine, but please lets ask on munki-dev)

gregneagle:
Since there has been no movement on this, I'm going to merge it and then edit it to simplify the user-agent. If people disagree with the changes they can discuss on munki-dev, preferably before the next official release.

PR #540, opened 2015-11-14
Minor internal text updates

keeleysam:
Update the download link for git that the build scripts print; use proper OS X name.

Also, perhaps the multiple copies of FoundationPlist.py within the code should be replaced with symlinks within git.


keeleysam:
I fixed these problems

gregneagle:
Preferences/README.md still looks pretty bad... IMHO

gregneagle:
...and now there are merge conflicts, too... Seems a lot of work to resolve those for this...

PR #547, opened 2015-11-24
Add deploy percent functionality

natewalck:
Adds deploy percent functionality to munki.

This will allow you to set a deploy percent in the pkginfo for a given item, which would ensure it gets deployed to that percent of the total clients.

get_machine_shard has chunks set to 100, so machines will get distributed equally between 1 and 100.

Pkginfo Key Example:

```
<key>deploy_percent</key>
<integer>20</integer>
```

This would deploy the given package to 20% of the clients. This seeds a md5 hash using the serial number, so the shards are more or less evenly distributed among a given set of clients.

This setting is to allow a ramped deployment of a given package. You would set this percent to a number you are comfortable with. The package will be roughly be deployed to that percent of the fleet. Once you are confident the package deployment is working as intended, you could remove the key and all clients would get it or you could increase the percent and watch the results, then keep increasing it until 100 (or just remove it).

An example deployment might be to set deploy_percent to 10 for 24 hours, increase it to 20 for another 24 hours and then finally remove the key to deploy it to the entire fleet.

Another thing we should consider is making the percent a truly random portion of the fleet. Instead of the same n% of machines getting the new packages each time, we should seed the md5 using serial + package name + version. This would allow a truly random % of the fleet to get the package rather than the same % each time.

Once we agree (or not) on the functional code, I'll get it tested on our fleet, update makepkginfo and add documentation for this feature.

### DIFF
--- a/code/client/munkilib/munkicommon.py
+++ b/code/client/munkilib/munkicommon.py
@@ -2398,6 +2398,29 @@ def getIntel64Support():
     else:
         return False
 
+def get_machine_deploy_percent(serial=None, salt=None, chunks=100):
+    """
+    get_machine_deploy_percent(serial=None, salt=None, chunks=100)
+    Returns the machine's unique deploy percent number. If deploy_percent is
+    set manually in ManagedInstalls.plist, use that value, otherwise generate
+    the deploy percent.
+    serial => Pass a serial for machine to get its deploy percent
+    salt   => Pass a salt to generate the hash
+    chunks => Pass an int to set number of chunks.
+    """
+    if pref('deploy_percent'):
+        return pref('deploy_percent')
+    else:
+        md5 = hashlib.md5()
+        if salt:
+            serial = str(serial) + str(salt)
+        md5.update(serial)
+        digest = md5.hexdigest()
+        number = int(digest, 16)
+        deploy_percent = number % int(chunks)
+        return int(deploy_percent) + 1
+
+
 MACHINE = {}
 def getMachineFacts():
     """Gets some facts about this machine we use to determine if a given
@@ -2411,6 +2434,9 @@ def getMachineFacts():
         MACHINE['munki_version'] = get_version()
         MACHINE['ipv4_address'] = get_ipv4_addresses()
         MACHINE['serial_number'] = hardware_info.get('serial_number', 'UNKNOWN')
+        MACHINE['deploy_percent'] = get_machine_deploy_percent(
+            MACHINE.get('serial_number', 'UNKNOWN')
+        )
 
         if MACHINE['arch'] == 'x86_64':
             MACHINE['x86_64_capable'] = True
--- a/tests/code/client/munkilib/munkicommon_test.py
+++ b/tests/code/client/munkilib/munkicommon_test.py
@@ -1,0 +1,48 @@
+#!/usr/bin/python
+"""
+munkicommon_test.py
+
+Unit tests for munkicommon functions
+
+"""
+
+import mock
+import unittest
+
+import munkicommon
+
+class MunkicommonGetMachineDeployPercentTest(unittest.TestCase):
+    def setUp(self):
+        """Setup tests with """
+        self.managedinstalls = {
+        }
+
+        self.machine_facts = {
+            'serial_number': 'C02Q32Y5GCN4',
+        }
+
+        self.machine_conditions = {}
+
+    def test_deploy_percent_function(self):
+        """Make sure percent is correct with a known serial and result"""
+        assert munkicommon.get_machine_deploy_percent(
+            self.machine_facts['serial_number']
+        ) == 31
+
+    @mock.patch('munkicommon.pref', return_value=100)
+    def test_deploy_percent_override(self, munkicommon_mock):
+        """Mock munkicommon.pref to return a deploy percent setting.
+        This should return that setting back rather than generating
+        a machine deploy percent (which would be 31 if generated)
+        """
+        assert munkicommon.get_machine_deploy_percent(
+            self.machine_facts['serial_number']
+        ) == 100
+
+
+def main():
+    unittest.main(buffer=True)
+
+
+if __name__ == '__main__':
+    main()


gregneagle:
Would like to see a post on munki-dev explaining the motivation and use-case here. Slack is a walled garden that isn't indexed by Google.

natewalck:
Added the ability to use the value of 'shard' from ManagedInstalls.plist if present, otherwise generate a shard for the machine based upon serial #.

jessepeterson:
There are cases where it's, well, not the case, but I've always thought of manifests as the place where "what is determined to be installed on which machines" is specified. Therefore I'd expect that anything that has to do with on which machines something is installed gets specified in a manifest. I.e. to more granularly control which catalogs, clients, manifests get a `deploy_percent` setting would it be more appropriate for this to be a manifest key? Or perhaps similar to how versions are specified in the manifst like `Firefox-20.0` so something like `Firefox-20.0-36%` or something?

natewalck:
We had originally considered manifests, but if you want to push an update,
it makes it far more annoying (as you have to specify version).

Manifests are definitely the place to specify it if it has never been
installed before, but for updates, it seems best suited to the pkginfo for
the version you want to push.

natewalck:
Here are the distribution numbers for this shard code, based upon fleet size. This shows how many nodes exist within each range (1-10%, 11-20%, etc).

Small Fleet - 100 Machines
-----------
1-10 9
11-20 9
21-30 11
31-40 14
41-50 10
51-60 14
61-70 6
71-80 9
81-90 8
91-100 10


Medium Fleet - 1000 Machines
-----------
1-10 114
11-20 102
21-30 101
31-40 103
41-50 91
51-60 109
61-70 87
71-80 107
81-90 89
91-100 97


Large Fleet - 10000 Machines
-----------
1-10 1010
11-20 1015
21-30 1005
31-40 998
41-50 959
51-60 981
61-70 990
71-80 1020
81-90 1026
91-100 996


Jumbo Fleet - 50000 Machines
-------
1-10 4989
11-20 4950
21-30 5041
31-40 4975
41-50 5036
51-60 5035
61-70 5043
71-80 4954
81-90 5006
91-100 4971

jessepeterson:
Hmm, good point re: updates. You're talking about `update_for` that 'magically' get picked up by Munki regardless of manifest inclusion? Yeah, that is an interesting one. I would almost consider it a better separation of concepts to have to include a separate `_install` item for the update that you want to control distribution for.

code/client/munkilib/updatecheck.py
```diff
@@ -1030,6 +1030,24 @@ def compare_version_keys(a, b):
...
+                    munkicommon.display_debug1(
+                        'Machine deploy percent is %s', MACHINE['shard'])
+                    if (MACHINE['shard'] > deploy_percent):
+                        # skip this one, go to the next
+                        munkicommon.display_debug1(
+                            'Item %s, version %s requires %s deploy percent. '
+                            "Our machine deployment percent is %s."
+                            % (item['name'], item['version'],
+                               item['deploy_percent'],
+                               MACHINE['shard']))
```
gregneagle: I think this message is confusing. I understand the intent, but I think you want something more explicit.
"Item %s, version %s, is set to deploy to machines with a shard value of %s or lower. Our shard value is %s" or similar would give the admin a clearer idea of what is happening.

mikedodge04: I agree

code/client/munkilib/updatecheck.py
```diff
@@ -1030,6 +1030,25 @@ def compare_version_keys(a, b):
...
+                        item['name'], item['version'], deploy_percent)
+                    munkicommon.display_debug1(
+                        'Machine deploy percent is %s', MACHINE['shard'])
+                    if (MACHINE['shard'] > deploy_percent):
+                        # skip this one, go to the next
+                        munkicommon.display_debug1(
+                            'Item %s, version %s is set to deploy to machines '
+                            'with a shard value of %s or lower. '
+                            'Our shard value is %s.'
+                            % (item['name'], item['version'],
```
gregneagle: Great. Now maybe a consistent terminology in lines 1036-1040. I think shard vs deploy_percent will be confusing.

Maybe even change the name of the "deploy_percent" key to "minimum_shard_value" or something similar.

gregneagle: Though of course, that might confuse people for other reasons. /shrug

mikedodge04: I feel like most people will not understand what shards are, and wouldn't use it. Where as deploy_percent makes it more accessible to a larger audience

gregneagle: Then let's stop calling it shard anywhere and figure out a better, consistent term. "deployment_phase"?

gregneagle: deploy_sequence, rollout_level, rollout_index

mikedodge04: What about deploy_percent_range, deploy_percent ?

keeleysam: It's not a range though - it's zero through whatever is specified.

In this or any similar form it should be called deploy percentage, as it's
not shards. Shards would be akin to A/B testing through defined groups which
this does not do.

On Tuesday, November 24, 2015, Mike Dodge <notifications@github.com> wrote:

> In code/client/munkilib/updatecheck.py
> <https://github.com/munki/munki/pull/547#discussion_r45786503>:
>
> > @@ -1030,6 +1030,25 @@ def compare_version_keys(a, b):
> >                          rejected_items.append(reason)
> >                          continue
> >
> > +                if item.get('deploy_percent', ''):
> > +                    deploy_percent = item['deploy_percent']
> > +                    munkicommon.display_debug1(
> > +                        'Considering item %s, version %s '
> > +                        'with %s deploy percent required',
> > +                        item['name'], item['version'], deploy_percent)
> > +                    munkicommon.display_debug1(
> > +                        'Machine deploy percent is %s', MACHINE['shard'])
> > +                    if (MACHINE['shard'] > deploy_percent):
> > +                        # skip this one, go to the next
> > +                        munkicommon.display_debug1(
> > +                            'Item %s, version %s is set to deploy to machines '
> > +                            'with a shard value of %s or lower. '
> > +                            'Our shard value is %s.'
> > +                            % (item['name'], item['version'],
>
> What about deploy_percent_range, deploy_percent ?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/munki/munki/pull/547/files#r45786503>.
>


-- 
Samuel Keeley

gregneagle: I think we need to distinguish between the target deploy percentage and the value on the current machine used to see if we are in the lucky group. Can't call them both "deploy_percent" and not confuse people.

mikedodge04: *SHRUGS* 
Another way to look at it is, we already ask munki users to learn a lot of terminology like manifests and catalogs, what's one more term... As long as we have a really good definition of what shards are, it should be fine. Right?

natewalck: deploy_percent and target_deploy_percent ?

arubdesu:
> deploy_percent and target_deploy_percent ?

Sounds good to me.
BTW, what would it take to add the client-side randomization, assuming it could be part of this PR? Something needs to keep it consistent across runs for that same product/version until it applies whatever the applicable config change is.

natewalck:
Updated to standardize terminology.

natewalck:
Added a test for this new functionality to make sure the generated deploy_percent is overridden if it is set in ManagedInstalls.plist.

Also laid down the framework for possible future tests for munkicommon functions.

natewalck:
Resurrecting this PR as I think it would still be useful in its current form.

I think it would be useful to get some basic deploy percent functionality in Munki. The munki-dev thread got off on various bunny trails and made the functionality more complex than was necessary.

I think this solution is the minimum viable product for deploy percents and anything more complex can definitely be done external to Munki.

keeleysam:
Here's what I'm thinking now, many months later.

We've been doing this as part of installable conditions with an admin provided condition, and you can also use also use that condition in manifests for conditional items.

Having it as a built in condition sounds like it would be useful, but I think that having it be a separate pkginfo key makes it less flexible.  We already have the code to evaluate conditions, and then you can use it identically between manifests and pkginfos.

weswhet:
I would appreciate this as a "built in" feature as well.

natewalck:
Admin provided conditionals work for new items (which is something we do currently).

For updating *existing* items, the deploy_percent needs to be in the pkginfo as in this implementation (There is one other alternative, but is is far more messy and out of scope for this PR. Read the munki-dev thread if you are interested).

This is most useful as a painless way to push *updates* to existing installs for a certain percent of a fleet. This is already flexible as you can set the deploy percent yourself if you'd like, so that is always an option. Accounting for everyone's edge cases is out of scope for this PR and can be handled via the other method mentioned in munki-dev if it is truly needed.

keeleysam:
You can use admin provided conditionals for updates to existing items currently, just use the same code that you would use in the manifest and put it in the `installable_condition` pkginfo key; it doesn't need to be a separate one and that is actually more flexible as you can chain different conditions.

What I'm saying is have `percent` as a built in condition, but then you don't need to add anything else.

natewalck:
Ah, I see. So keep the shard generating data and the override, but ditch the purpose-built pkginfo key.

gregneagle:
@keeleysam Could you walk through an admin workflow of using a "shard_percent" "fact" and installable_conditions to do a phased deployment? I think I understand the concept and even a possible workflow, but it would be good to have it detailed.

Perhaps it would make sense to integrate https://github.com/munki/munki-facts with Munki proper and modularize this stuff...

natewalck:
I like the idea of integrating munki-facts quite a bit. That would make adding new stuff in the future far easier.

vmiller:
I don't really have an opinion on phased deployment, but I do like the idea
of integrating munki-facts.

Vaughn

erikng:
I too would like to see munki facts in munki proper.

grahamgilbert:
Our procedure is as follows for new items:

In the manifest, add a conditional item:

```
<dict>
	<key>condition</key>
	<string>deploy_percent &lt;= 25</string>
	<key>managed_installs</key>
	<array>
	    <string>some_item</string>
	</array>
</dict>
```

And for items that have older versions already existing in the repo, add the following the it item's pkgsinfo:

```
<key>installable_condition</key>
<string>deploy_percent &lt;= 75</string>
```

Adding the `installable_condition` to items that don't have an older version available to the client will obviously result in warnings that the item isn't available.

natewalck:
Changed the code to only set the machine fact. This still allows an override, which I think is still needed (and was specifically requested). Having this as a default fact is useful and the additional logic code was completely unnecessary.

So it could go either here or moved to munki-facts, but I'm not certain how I'd offer an override in munki-facts (which would be the only possible downfall of putting it there).

natewalck:
Also, we may consider switching this to shard using UUID as all machines have a UUID, but not all machines have a serial # (blank board).

natewalck:
Although, after looking at UUID a bit more, I'm not convinced it is necessary (for the extra code it would require). In our fleet, we have < 10 machines without serial numbers, so it does not affect the distribution to any statistically significant degree.

natewalck:
No longer needed I gues :)

PR #549, opened 2015-11-30
setup MSC logging before making logging calls

juofmgoblue:


gregneagle:
Interesting no-one noticed that for months...

juofmgoblue:
From what I can tell, the only logging call effected was "launched", and prior to this change it was merely omitted (no crash/etc.).  All other logging calls still work as expected, as they came after the setup call.  I don't know anyone other than ourselves using MSC GUI logs, otherwise maybe someone would have complained earlier.

PR #553, opened 2015-12-08
Added try block to catch malformed Info.plist.

### LINKED ISSUE #552: MSC displays blank "Updates" tab when calling 'convertIconToPNG' when a malformed Info.plist file is present
When an item is being considered for listing in the Managed Software Center **Updates** tab, `MunkiItems.py` will attempt to retrieve its icon, or absent an existing one attempt to create one. To do this it calls the `convertIconToPNG()` function which locates an existing on-disk version of the item and converts its icon from `.icns` to PNG by way of the `CFBundleIconFile` key in the application's `Info.plist` file. This process errors out with an `AttributeError` exception if the `Info.plist` in the application bundle is present but not actually plist-formatted and thus not in the `NSDictionary` family, but of some other non-compatible class. Additional sanity-checking is needed on the `Info.plist` file's object type so that we return `False` if it is not of a compatible type (NSDictionary) instead of continuing with a broken plist file.

bruienne:
convertIconToPNG() would cause MSC Updates tab to hang when processing applications with a malformed Info.plist file. This wraps `info.get('CFBundleIconFile', app_name)` in a try block that returns `False` if the `get()` fails. Fixes #552.

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center/MunkiItems.py
+++ b/code/apps/Managed Software Center/Managed Software Center/MunkiItems.py
@@ -244,28 +244,32 @@ def convertIconToPNG(app_name, destination_path, desired_size):
     except FoundationPlist.FoundationPlistException:
         info = {}
     try:
-        icon_filename = info.get('CFBundleIconFile', app_name)
-    except AttributeError:
+        try:
+            icon_filename = info.get('CFBundleIconFile', app_name)
+        except AttributeError:
+            icon_filename = app_name
+        icon_path = os.path.join(app_path, 'Contents/Resources', icon_filename)
+        if not os.path.splitext(icon_path)[1]:
+            # no file extension, so add '.icns'
+            icon_path += u'.icns'
+        if os.path.exists(icon_path):
+            image_data = NSData.dataWithContentsOfFile_(icon_path)
+            bitmap_reps = NSBitmapImageRep.imageRepsWithData_(image_data)
+            chosen_rep = None
+            for bitmap_rep in bitmap_reps:
+                if not chosen_rep:
+                    chosen_rep = bitmap_rep
+                elif (bitmap_rep.pixelsHigh() >= desired_size
+                      and bitmap_rep.pixelsHigh() < chosen_rep.pixelsHigh()):
+                    chosen_rep = bitmap_rep
+            if chosen_rep:
+                png_data = chosen_rep.representationUsingType_properties_(
+                    NSPNGFileType, None)
+                png_data.writeToFile_atomically_(destination_path, False)
+                return True
+    except Exception:
         return False
-    icon_path = os.path.join(app_path, 'Contents/Resources', icon_filename)
-    if not os.path.splitext(icon_path)[1]:
-        # no file extension, so add '.icns'
-        icon_path += u'.icns'
-    if os.path.exists(icon_path):
-        image_data = NSData.dataWithContentsOfFile_(icon_path)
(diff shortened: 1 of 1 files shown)


code/apps/Managed Software Center/Managed Software Center/MunkiItems.py
```diff
@@ -243,7 +243,10 @@ def convertIconToPNG(app_name, destination_path, desired_size):
             os.path.join(app_path, 'Contents/Info.plist'))
     except FoundationPlist.FoundationPlistException:
         info = {}
-    icon_filename = info.get('CFBundleIconFile', app_name)
+    try:
+        icon_filename = info.get('CFBundleIconFile', app_name)
+    except AttributeError:
+        return False
```
gregneagle: If there is an AttributeError, why not continue with icon_filename = app_name?

jamesez:
The particular issue we ran across was that the app's info.plist was really just a string (it had a key and string, like a dict, but without any plist/dict wrapping on it).  FoundationPlist deserialized it as an NSString subclass, so calling `get` on it returns an exception.

Actually I suggest the try block be higher up - there are other untrapped errors that could cause similar behavior (eg, NSData might not be able to read the guessed icon file for whatever reason, NSImage might not be able to get an imageRep, might not be able to write the destination file).  Any of those should cause MSC to show the generic package icon.

gregneagle:
OK, but bad Info.plist does not mean "no useable icon". It might in the instance you saw, but it doesn't mean that generically.

bruienne:
So how about this, catch the bad plist case and just pass on `app_name` to `icon_filename` but have any other issues in reading, converting or writing the icon file result in `False`?

```
def convertIconToPNG(app_name, destination_path, desired_size):
    '''Converts an application icns file to a png file, choosing the
    representation closest to (but >= than if possible) the desired_size.
    Returns True if successful, False otherwise'''
    app_path = os.path.join('/Applications', app_name + '.app')
    if not os.path.exists(app_path):
        return False
    try:
        info = FoundationPlist.readPlist(
            os.path.join(app_path, 'Contents/Info.plist'))
    except FoundationPlist.FoundationPlistException:
        info = {}
    try:
        try:
            icon_filename = info.get('CFBundleIconFile', app_name)
        except AttributeError:
            icon_filename = app_name
        icon_path = os.path.join(app_path, 'Contents/Resources', icon_filename)
        if not os.path.splitext(icon_path)[1]:
            # no file extension, so add '.icns'
            icon_path += u'.icns'
        if os.path.exists(icon_path):
            image_data = NSData.dataWithContentsOfFile_(icon_path)
            bitmap_reps = NSBitmapImageRep.imageRepsWithData_(image_data)
            chosen_rep = None
            for bitmap_rep in bitmap_reps:
                if not chosen_rep:
                    chosen_rep = bitmap_rep
                elif (bitmap_rep.pixelsHigh() >= desired_size
                      and bitmap_rep.pixelsHigh() < chosen_rep.pixelsHigh()):
                    chosen_rep = bitmap_rep
            if chosen_rep:
                png_data = chosen_rep.representationUsingType_properties_(
                    NSPNGFileType, None)
                png_data.writeToFile_atomically_(destination_path, False)
                return True
    except Exception:
        return False

    return False
```

gregneagle:
Do any of the Cocoa methods you've wrapped in a general try/except Exception block actually raise exceptions on failure?

But this generally seems fine.

gregneagle:
Were you going to implement the code you showed in your comment?

bruienne:
Sorry, completely missed it didn't actually push my updated commit. Implemented the code as it is in my earlier comment.

PR #557, opened 2015-12-20
Add option to set NSMutableURLRequest timeoutInterval property

MichalMMac:
NSMutableURLRequest timeoutInterval property was originally set with magic constant (10 seconds) in gurl.py. I propose to set timeoutInterval from Munki preferences making it easily configurable by user.

Proposed key is called **HTTPConnectionTimeoutInterval** and is set to default value of 10 seconds in munkicommon.py.

This is result of my testing [described](https://groups.google.com/forum/#!topic/munki-discuss/irW_64WJOhM) in munki-discuss google group. Setting bigger value for HTTPConnectionTimeoutInterval should increase download survivability in bad network conditions.

Please verify my changes. I am new to Munki codebase and it is possible I've missed something.

### DIFF
--- a/code/client/munkilib/fetch.py
+++ b/code/client/munkilib/fetch.py
@@ -156,7 +156,8 @@ def get_url(url, destinationpath,
                'additional_headers': header_dict_from_list(custom_headers),
                'download_only_if_changed': onlyifnewer,
                'cache_data': cache_data,
-               'logging_function': munkicommon.display_debug2}
+               'logging_function': munkicommon.display_debug2,
+               'connection_timeout': munkicommon.pref('HTTPConnectionTimeoutInterval')}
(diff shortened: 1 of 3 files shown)


natewalck:
How many HTTP related options are currently available? Would it make sense
to migrate them to a dictionary of HTTP options as more are exposed via
settings?

Nate

gregneagle:
For my own future support sanity, I think I'll just change the hard-coded timeout value to 60 seconds. We can revisit this in the future if needed, but I suspect it will be fine.

gregneagle:
Closing this in favor of https://github.com/munki/munki/commit/976c842224c3d61c20f55a72f03bc8ad75e90069

erikng:
Can't wait to try this out. Thanks for the commit.

PR #559, opened 2016-01-07
Fixed manifestutil traceback if no subcommands are given

poundbangbash:
In manifestutil if no subcommands are given manifestutil exits with a
trackback.  I added a check to see if the number of subcommands is < 1
and to list the available subcommands if so.


gregneagle:
Thank you for being one of the three manifestutil users in the universe!

poundbangbash:
Ha. We use it as the only mechanism to manage munki besides raw text editing. Learn the tool and use the defaults and they'll always be available. ;)

PR #562, opened 2016-01-29
Automatic Apple Updates

keeleysam:
This adds support for automatic unattended installation of Apple updates.  It is intended to cut down on the need for creating pkginfos just to add unattended_install keys.
    
Setting `AutomaticAppleUpdates` to true will cause all Apple updates which do not require a logout or restart to be treated as unattended, getting the blocking applications list from the Apple distribution.
    
 It is currently limited to just Yosemite and newer, though it can likely be extended down to Mavericks.  The behavior is similar to when "Install OS X updates" is set to true in the App Store's preferences.

### DIFF
--- a/code/client/munkilib/appleupdates.py
+++ b/code/client/munkilib/appleupdates.py
@@ -253,10 +253,12 @@ class AppleUpdates(object):
             if rewrite_pkg_urls and 'URL' in package:
                 package['URL'] = self.RewriteURL(package['URL'])
             if 'MetadataURL' in package:
-                package['MetadataURL'] = self.RewriteURL(package['MetadataURL'])
+                package['MetadataURL'] = self.RewriteURL(
+                    package['MetadataURL'])
         distributions = product['Distributions']
         for dist_lang in distributions.keys():
-            distributions[dist_lang] = self.RewriteURL(distributions[dist_lang])
+            distributions[dist_lang] = self.RewriteURL(
+                distributions[dist_lang])
 
     def RewriteCatalogURLs(self, catalog, rewrite_pkg_urls=False):
         """Rewrites URLs in a catalog to point to our local replica.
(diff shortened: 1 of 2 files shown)


gregneagle:
Tests aren't passing:

```
% sudo ./runtests.sh 
  File "./code/client/munkilib/appleupdates_test.py", line 23, in <module>
    import appleupdates
  File "/Users/Shared/munki-git/munki/code/client/munkilib/appleupdates.py", line 1664
    item.get('RestartAction', 'None')='None' and
                                     ^
SyntaxError: invalid syntax
..................
----------------------------------------------------------------------
Ran 18 tests in 0.001s

OK
```

gregneagle:
appleupdates.py line 1664 should be
`item.get('RestartAction', 'None') == 'None' and`

Odd your PyLinting didn't catch that...

Have you actually _run_ this code?

gregneagle:
I'd like to extend the automatic install back to whatever OS X version it was added to -- is that only 10.10 and not earlier?

keeleysam:
I resolved the syntax error - it's all working now.

The option for automatic installs was added in Yosemite, but I would feel comfortable extending it back to Mavericks as its catalogs have those keys and softwareupdate through the MAS on that version exhibits the same behavior with blocking applications.

Further back than that the behavior is also the same on Mountain Lion, but aside from printer driver updates those aren't supported anymore so the impact would be very low.

gregneagle:
Let's change the pref name to 'UnattendedAppleUpdates' or to be consistent with the other usages of "unattended",

weswhet:
@keeleysam Does this take into account the auto blocking applications that the App Store would normally use?

gregneagle:
Sam: please write this up and post to munki-dev and munki-discuss

PR #574, opened 2016-03-16
Fixed wobbling progress spinner

hjuutilainen:
On a non-retina machine, the progress spinner animation on the updates view has always been wobbling for me. In this pull request I've updated the progress-spinner graphic to be an exact square (40x40px) by cutting 1px of empty, transparent space from both sides. I've then changed the references to it in the base.css to be 20x20px instead of 21x20px.

Seems to fix the wobbling for me but I've only tested this on 10.11 (both retina and non-retina).


gregneagle:
Yay! I looked at this last week as part of all the other fixes, but didn't put much energy into it.

PR #575, opened 2016-03-16
Strip whitespace from raw_input

filipp:
To prevent catalog error when (for example) dragging a folder to the
Terminal window to set the repo path.

### DIFF
--- a/code/client/munkiimport
+++ b/code/client/munkiimport
@@ -61,10 +61,12 @@ def raw_input_with_default(prompt='', default=''):
 
     if default:
         prompt = '%s [%s]: ' % (prompt, default)
-        return raw_input(prompt).decode('UTF-8') or default.decode('UTF-8')
+        result = raw_input(prompt).decode('UTF-8') or default.decode('UTF-8')
     else:
         # no default value, just call raw_input
-        return raw_input(prompt + ": ").decode('UTF-8')
+        result = raw_input(prompt + ": ").decode('UTF-8')
+
+    return result.strip()
 
 
 def makeDMG(pkgpath):
--- a/code/client/munkiimport
+++ b/code/client/munkiimport
@@ -731,10 +731,21 @@ def configure():
 def init_repo(path):
     """Initializes a directory with the repo subdirs"""
     path = os.path.abspath(path)
+
     if not os.path.exists(path):
-        os.mkdir(path)
+        try:
+            os.mkdir(path)
+        except OSError as e:
+            print('Failed to initialize repo at %s (%s)' % (path, e))
+            return
+
     for d in REPO_DIRS:
-        os.mkdir(os.path.join(path, d))
+        full_path = os.path.join(path, d)
+        try:
+            os.mkdir(full_path)
+        except OSError as e:
+            print('Failed to create %s directory (%s)' % (d, e))
+            return
 
     pref('repo_path', path)
 
--- a/code/client/munkiimport
+++ b/code/client/munkiimport
@@ -107,7 +107,7 @@ def repoAvailable():
         mountRepoCLI()
     if not os.path.exists(REPO_PATH):
         return False
-    for subdir in ['catalogs', 'manifests', 'pkgs', 'pkgsinfo']:
+    for subdir in REPO_DIRS:
         if not os.path.exists(os.path.join(REPO_PATH, subdir)):
             print >> sys.stderr, "%s is missing %s" % (REPO_PATH, subdir)
             return False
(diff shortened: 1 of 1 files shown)


filipp:
I messed up with the branching. There's actually two things in here - strip whitespace from raw_input_with_default and --init command for munkiimport (just creates the necessary folders and setup up the repo path)

gregneagle:
Convince me. Your code as-is always strips leading and trailing spaces from all input during --configure. What if someone _wants_ a space? (Yes, I know that seems weird and unlikely, but silently stripping spaces without telling the user also seems like it could confuse people)

code/client/munkiimport
```diff
@@ -691,12 +691,17 @@ def cleanupAndExit(exitcode):
...
     try:
         _prefs = FoundationPlist.readPlist(PREFSPATH)
     except FoundationPlist.NSPropertyListSerializationException:
         return None
+
+    if value:
+        _prefs[prefname] = value
+        return FoundationPlist.writePlist(_prefs, PREFSPATH)
+
     if prefname in _prefs:
```
gregneagle: Yuck. Sometimes the function returns a value and sometimes it returns nothing. Seems like this should be two different functions -- one that gets, and one that sets.

code/client/munkiimport
```diff
@@ -723,6 +728,17 @@ def configure():
...
 
 
+def init_repo(path):
+    """Initializes a directory with the repo subdirs"""
+    path = os.path.abspath(path)
+    if not os.path.exists(path):
+        os.mkdir(path)
+    for d in REPO_DIRS:
+        os.mkdir(os.path.join(path, d))
+
```
gregneagle: Probably want to add some error handling to handle when os.mkdir() fails.

code/client/munkiimport
```diff
@@ -760,6 +777,8 @@ def main():
 
     parser = PassThroughOptionParser(usage=usage, epilog=epilog)
 
+    parser.add_option('--init', action='store_true',
+                      help='Initialize a munki repository.')
```
gregneagle: What is the implied behavior if --init is called against an existing path? "Initialize" sounds like it might clear out the contents. Pretty sure it only creates any missing directories, so the help text might be more explicit.

gregneagle: I am a bit conflicted about this option; people might do something silly like `munkiimport --init ~` and be confused about the results...

code/client/munkiimport
```diff
@@ -800,6 +819,10 @@ def main():
         configure()
         exit(0)
 
+    if options.init:
+        init_repo(arguments[0].strip())
```
gregneagle: `munkiimport --init "  munki_repo"` would do something the user would not expect!

I've seen people use the "pre-pend spaces so the thing sorts to the top" trick; why prevent them from doing so just because you would never do that?

code/client/munkiimport
filipp: :) When you get a value, it is returned, when you set it, it is set and nothing is returned. Could also always just return the value, but I don't need to get back what I just set.

code/client/munkiimport
filipp: Just to pretty-print the already self-explanatory OSError? OK.

gregneagle: People don't like to see unhanded exceptions and Python tracebacks.

code/client/munkiimport
filipp: OK. It works as I personally would expect - populate an existing directory or create it first. It might indeed be a good idea to return a string saying "New repository created at %s"

code/client/munkiimport
filipp: Fair enough. :) I just spent 30 minutes trying to figure out why munki wouldn't work when I simply dragged the folder to the Terminal window. I looked at the options and none of them seemed like they *should* have leading or trailing whitespace. Maybe just an rstrip() would be better.

code/client/munkiimport
filipp: I personally would take the traceback rather than an opaque message, but you're probably right. :)

gregneagle: People will report it as a bug in `munkiimport` instead of a mistake on their part.

code/client/munkiimport
gregneagle: There's a difference between doing an rstrip() of raw_input during --configure and stripping spaces from the _already parsed arguments_ the shell has handed you. I think you should not do the latter at all.

code/client/munkiimport
filipp: Good point. Added some exception handling.

code/client/munkiimport
```diff
@@ -731,10 +731,21 @@ def configure():
...
+            return
+
     for d in REPO_DIRS:
-        os.mkdir(os.path.join(path, d))
+        full_path = os.path.join(path, d)
+        try:
+            os.mkdir(full_path)
+        except OSError as e:
+            print('Failed to create %s directory (%s)' % (d, e))
+            return
```
gregneagle: Do we also want to follow UNIX convention and exit non-zero?

filipp: Sure :)

gregneagle:
If you are still interested in getting these changes into munkiimport, can you update the PR so it will apply to the current master?

gregneagle:
`munkiimport` has changed dramatically since this PR was written. Closing this.

PR #576, opened 2016-03-17
Log more details about apple_item presence.

bruienne:
Following up after updating the wiki re: SUS behavior when an `apple_item` is present.

This PR logs the names of Munki items that were marked as an `apple_item` either automatically (due to their `installs` or `receipts` array containing an item that starts with `com.apple`) or by the admin. This should help in troubleshooting situations where SUS checks are effectively blocked due to the user postponing updates and thus also postponing SUS checks. With this additional logging the admin will be able to identify the Munki item causing SUS checks to be skipped and take appropriate action, as outlined in the wiki [here](https://github.com/munki/munki/wiki/Apple-Software-Updates-With-Munki).


gregneagle:
I'll merge this, but I think I'll change the logging message to fit my editorial sensibilities and reduce future questions to something like: "Treating %s as an apple_item - this may block Apple SUS updates"

PR #577, opened 2016-03-17
Add new preference to control Munki Status window order.

bruienne:
This PR attempts to implement a new optional preference key `ShowProgressWindowOverScreenSaver` which when set to `True` causes the Munki Status application to appear above both screen savers and the Policy Banner app window.

Screen savers and the Policy Banner both use the same window level of 999, or `NSScreenSaverWindowLevel`. Issues have been reported by admins in organizations that use the built-in Policy Banner application to display their organization's AUP at the login screen. Because Munki Status uses `NSScreenSaverWindowLevel - 1` for its window's level this means that it will be obscured by the Policy Banner window, causing possible confusion with users. Unfortunately there is no middle ground between the window level of screen savers and the Policy Banner so the choice was made to name the key for its most intrusive effect: causing Munki Status to appear above a running screen saver. Admins who wish to work around the Policy Banner grabbing focus will have to consider whether they may have configurations where a screen saver would be running when Munki Status might also be called, i.e. at the login window.

The result with the key set to `True` and Policy Banner active:

pb ms

Discussion regarding this new preference and the modified behavior it causes is welcomed.


gregneagle:
In my testing, MunkiStatus did not appear in front of the screensaver. I'd rather not at yet another preference and just handle this automagically.

bruienne:
Using `self.window.setLevel_(NSScreenSaverWindowLevel + 1)` should take care of that, as far as I can tell the levels are arbitrary, all the way up to `kCGCursorWindowLevelKey` at level `2,147,483,630` I'm fine with not using a pref, would you instead detect running Policy Banner and adjust window accordingly?

gregneagle:
I would set the level to 1000, I would check if PolicyBanner exists, and I would do it only on affected OS X versions

bruienne:
This is testing successfully right now for me. The preference key code was removed and a check for files triggering Policy Banner is done, and a check for OS version is performed. Right now only 10.11 clients with active Policy Banner configurations should see the window level of Munki Status raised to `1000`.

code/apps/MunkiStatus/MunkiStatus/MSUStatusWindowController.py
```diff
@@ -156,7 +171,14 @@ def initStatusSession(self):
             if consoleuser == None or consoleuser == u"loginwindow":
                 # needed so the window can show over the loginwindow
                 self.window.setCanBecomeVisibleWithoutLogin_(True)
-                self.window.setLevel_(NSScreenSaverWindowLevel - 1)
+                # Check if we're to appear above screen saver or Policy Banner
+                #   and our Darwin major version is in the list
+                if havePolicyBanner and clientOS in affectedOSVersions:
+                    # Always appear above screen saver or Policy Banner window
+                    self.window.setLevel_(NSScreenSaverWindowLevel + 1)
```
gregneagle: Still think this should be just NSScreenSaverWindowLevel

bruienne:
The case this accounts for is the Policy Banner appearing after Munki Status has started. The way the Policy Banner works is that it re-appears after a user has clicked "Agree" but didn't log in, such as when Munki Status is active. The Policy Banner would then re-appear shortly after and obscure Munki Status once more since the most recent window with the same window level wins. By pushing it above `NSScreenSaverWindowLevel` it would always remain visible. I recognize that the whole Policy Banner issue is icky, but given the choice between having Munki Status appear above everything or having to communicate to users that sometimes it may be covered up, and sometimes it may not, I'd rather go with guaranteed visibility.

gregneagle:
1000 > 999

PolicyBanner is at 999
MunkiStatus can be at 1000

No need to be at 1001.

Starting to think we are talking about different things here as there are only so many ways I can say this...

bruienne:
I fully understand the math. I know that code suggests PB runs at 999 but when testing with `self.window.setLevel_(NSScreenSaverWindowLevel)` the Policy Banner window still spawns covering Munki Status if it is running. Since it re-spawns every 60 seconds it is easy to imagine a case where a running Munki Status session gets covered up by the Policy Banner.

It wasn't my intention to flip the script on the core issue mid-stream but after testing showed 1000 wasn't enough I went with 1001 which gets us this PR's intended behavior.

gregneagle:
Now that you've clarified that point, and since that doesn't match my observations from brief testing a few days ago, I'll try to find time to retest soon.

I really dislike the idea of putting MunkiStatus _over_ the screensaver unless it is absolutely necessary.

gregneagle:
Please test MunkiStatus as found here: https://github.com/munki/munki/tree/policy-banner

This addresses a few issues:

1) The NSWindowLevel is adjusted if running on 10.11+ and a PolicyBanner is present.
2) The translucent backdrop window is also moved in front of the PolicyBanner, preventing the user from clicking on the policy banner (or the Accept button)
3) In the case where the Policy Banner is (re-)displayed after the MunkiStatus window appears, the next time the MunkiStatus window is updated, it is again moved to the front.

bruienne:
After testing this in a few scenarios (perform logout and install which triggers PB immediately; delayed PB invocation after MunkiStatus was invoked) everything seems to be working okay here and we're fine with it for our org. It would be good to get a few more eyes on this from others with Policy Banners, especially now that 10.11.4 seems to blow up the PB window even larger which could possibly obscure MunkiStatus where previously it may not have.

 👍 LGTM

gregneagle:
Closing this since the 2.6.0 release addresses this issue.

PR #578, opened 2016-03-17
Update updatecheck.py

wesdoniak:
fix client_resources/site_default.zip error. ClientResourcesFilename is being assigned the ManifestName resource_name.

   
 `2016/03/17 20:37:33 [error] 24315#0: *54456 open() "/usr/local/munki_repo/client_resources/osx_managed.zip" failed (2: No such file or directory), client: 207.231.170.7, server: munki.domain.com, request: "GET /munki_repo/client_resources/osx_manage
d.zip HTTP/1.1", host: "munki.domain.com"`

### DIFF
--- a/code/client/munkilib/updatecheck.py
+++ b/code/client/munkilib/updatecheck.py
@@ -2901,8 +2901,7 @@ def download_client_resources():
             resources_name += '.zip'
         filenames.append(resources_name)
     else:
-        filenames.append(munkicommon.report['ManifestName'] + '.zip')
-    filenames.append('site_default.zip')
+        filenames.append('site_default.zip')
 
     resource_base_url = (
         munkicommon.pref('ClientResourceURL') or


erikng:
What does this change do? I have some expectations at my site that the Client Resources file is being assigned this name.

erikng:
After looking at this code, this would break me. I don't set a ClientResourceName, but rely on the previous method and URL rewrites. This was the method I was told to use after a rather lengthy discussion on munki-dev.

wesdoniak:
Shouldn't we just ClientResourceFilename and ClientResourceURL to set these values? The current config sets 2 default values if a ClientResourceFilename is not specified, this is causing errors (outlined) in the log snippet. If we want to default to ManifestName or site_default.zip, then there should be added logic to determine if ManifestName.zip exists on the repository.

erikng:
I have 300 sites with different types of users. Rather than setting those values or creating the .zips for each site, I dynamically rewrite the URL based on the request sent via the client. If a site wants customization, I change the way the rewrite is handled and create their respective zip files.

wesdoniak:
Understood. However, if you are standardizing your sites by manifest name, it makes sense to set the ClientResourceFilename to match manifest (rather than relying on a URL rewrite). Not really a big deal.. web server error logs just tend to annoy me due to monitoring.

erikng:
Well that's the thing - each client still has a unique manifest. I had suggested a new "base file name " key to do that exact standardization you refer to, but it was shot down by many people.

gregneagle:
No, this code is behaving precisely as designed.

If the client has ClientResourceFilename defined, it requests a zip with that name; if that fails (or if ClientResourceFilename is not defined), it requests a zip with the same name as the primary manifest; if that fails, it requests a zip with the name "site_default.zip".

This behavior is documented here: https://github.com/munki/munki/wiki/Client-Customization#making-client-customization-archives-available-to-clients

Your change would change the documented behavior, breaking anyone using it/depending upon it.

It's not clear to me what problem you are attempting to address other than failed GET requests on the web server, which are expected, since the only way to determine if "foo.zip" is available on the Munki server is to attempt to GET it. (I suppose we could do a HEAD request, but that's not more efficient when the resource exists, and the same overhead when it doesn't).

gregneagle:
"If we want to default to ManifestName or site_default.zip, then there should be added logic to determine if ManifestName.zip exists on the repository." <-- if you know of a way to determine if a file exists on a remote web server other than simply trying to GET it, I would be pleased to know what that is.

wesdoniak:
In my case we are using site_default.zip to deploy client customizations to MSC. Based on your response and looking at the code closer, I do see the underlining concerns. My only question is why should there be an additional GET request if site_default.zip is found on the remote server. Seems like the logic would be; if site_default.zip exists then don't check for ['ManifestName'] + '.zip'. 

As I mentioned; not really an issue other than web-server errors (the only error being generated in my case) each time a client checks-in. 

Cheers,

gregneagle:
"In my case we are using site_default.zip to deploy client customizations to MSC"

So am I.

"My only question is why should there be an additional GET request if site_default.zip is found on the remote server. Seems like the logic would be; if site_default.zip exists then don't check for ['ManifestName'] + '.zip'."

Because that's not the defined order!

The order is: ask for the zip named by ClientResourceFilename (if defined).
If not found, ask for the zip named the same as the current manifest.
If not found, ask for the site_default.zip.

<manifestname>.zip is _more specific_ than site_default. Therefore we must try to ask for that _first_.

This is precisely analogous to how Munki requests _manifests_ for the current client:

Ask for ClientIdentifer (if defined)
Ask for full hostname
Ask for short hostname
Ask for serial number
Ask for site_default

If we asked for site_default earlier, clients would never discover their more-specific manifest.

Since you agree this is behaving as expected, I'm going to close this. If in the future, you have questions about Munki's expected behavior, consider asking them on the munki-discuss or munki-dev lists before writing code.

PR #580, opened 2016-03-21
Allow the use of middleware

waderobson:
**If** the `Middleware` preference is set, munki will hand off the requested URL to the middleware. The middleware can then output additional headers, a new URL or both. 

An exception is thrown if the middleware reaches a 5-second timeout or if the middleware does not exit cleanly.
 
Based off the discussion [here](https://groups.google.com/forum/#!topic/munki-dev/cPgn1lm90U4)


code/client/munkilib/middleware.py
```diff
@@ -0,0 +1,92 @@
...
+
+class Timeout(Exception):
+    """Timeout exception for middleware"""
+    pass
+
+def kill_proc(proc, timeout):
+    timeout["value"] = True
+    proc.kill()
+
+def cmd_with_timeout(cmd, timeout_sec):
```
gregneagle: munkicommon already has a subclass of Popen that supports timeouts; can you use that?
https://github.com/munki/munki/blob/master/code/client/munkilib/munkicommon.py#L150

(I'm not actually convinced you need to have a timeout for this, but if you are, it would be nice to use existing/tested code)

gregneagle:
Almost there.

I think there is too much logic defined in middleware.py. I think it should just look for the existence of a file  relative to itself; IOW, the admin would be responsible for providing the actual middleware as a file at /usr/local/munki/middleware_local.py (or similar). The module at /usr/local/munki/middleware.py would simply see if that file exists, and if it does, load it and call a predefined function within.

I'll write up an example (tomorrow if I can find the time).

I wonder if we need to support multiple middleware modules (meaning admin can pass the options through more than one middleware module); I suspect it's not worth the effort/complexity.

bochoven:
Maybe it would be better to use a middleware design pattern that allows for chaining middleware, the preference for middleware should then be an array.

It may also be beneficial to also allow to override the Response object (to translate a certain response to something that munki understands. 

Some code that is currently in the core like the Client Certificate authentication code could then be moved out of the core into to a middleware module.

waderobson:
Ok yeah, I kinda went crazy. 

> The module at /usr/local/munki/middleware.py would simply see if that file exists, and if it does, load it and call a predefined function within

What about searching for a folder /usr/local/munki/middleware/ and if found inserting it into the path importing a predetermined pkg, in this case middleware.py, and running the main fuction.
Running this from fetch;
`os.path.exists('../middleware/')`
`os.sys.path.insert(0,'../middleware/')`
`import middleware` 
`options = middleware.main(options)`

_Kinda_ like the way conditionals are run. 
Supporting files could be included in the /usr/local/munki/middleware/ folder then.

Is that more along the lines of what you're thinking @gregneagle

gregneagle:
Unless we are planning to support chained middleware, I don't really see the need for a middleware directory; we could just keep it simple and look for a single file in /usr/local/munki (though do not hard-code that path; instead it should be relative to the munkilib directory)

I was thinking something dynamic along these lines: https://github.com/autopkg/autopkg/blob/master/Code/autopkglib/__init__.py#L506-L612

gregneagle:
@bochoven: Can you come up with some examples that would justify the added complexity you propose?

I don't see a clean way to extract client certificate authentication out of NSURLConnection/NSURLSession...

natewalck:
I think having it broken out into its own directory makes a lot of sense.
This is especially useful if you want to have libraries that support your
middleware goals (like an AWS library or some other custom library for an
internal system). Beats cluttering up the main munki directory.

It seems like a small abstraction that gives more flexibility. How much
added complexity does it really tack on?

Nate

gregneagle:
I'm not referring to a middleware directory as added complexity.

I'm referring to @bochoven 's proposals for changing middleware; middleware to alter _responses_ as well as _requests_, and factoring _out_ client certificate authentication.

These seem to add complexity.

natewalck:
Ahhh yes. That does indeed.

Let me take a step back then. What is the overall goal of his middleware
change?

Swapping in your own client certificate auth code would bypass a large
chunk of munki code. For this specific feature, is there a feature that
would be beneficial to add to munki rather than bypassing it and using
different code altogether?

I do like the idea of a more modular munki, but it seems like this could
add a large amount of complexity and make the code harder to follow in its
current state. (Not to mention supporting it).

Nate

bochoven:
I haven't looked deeply at the current codebase, but middleware could/should take care of all authentication mechanisms (basic auth, client certificates, etc.). I would think that it could simplify the core by moving optional parts to middleware.

gregneagle:
@bochoven Definitely not interested in re-architecting the HTTP/HTTPS code yet again unless there is a very compelling need. NSURLConnection/NSURLSession have their own hooks for handling authentication challenges. gurl.py can handle Basic and Digest by providing username and password; client cert auth is handled by Apple's code. Not at all interested in opening that can of worms.

Again: do you have specific examples that lead you to proposing this?

bochoven:
@gregneagle Looking deeper into Gurl.py I see that it won't be easy to add in middleware in the way I proposed. I don't have a particular example/need at the moment that would warrant a code rewrite. The main reason I proposed the change to the current PR is that it would simplify adding new functionality in the future.

waderobson:
> I was thinking something dynamic along these lines: https://github.com/autopkg/autopkg/blob/master/Code/autopkglib/__init__.py#L506-L612
@gregneagle Thanks! I will take a crack at this tonight.

@natewalck My use case is enabling s3/CloudFoo. Unique headers/query params are required to authenticate to these services.

@bochoven all I'm interested in getting back is the content and headers, for debugging which can be viewed with `managedsoftwareupdate -vvvv`.

waderobson:
Just checking back in here. I have made the changes to allow searching for files that start with middleware*.py and importing them under a common namespace "middleware". I hope that was what you were getting at @gregneagle. 

I've been testing with s3 and now google cloud storage and so far no issues. The only thing I've been wondering about is how best to disable the middleware when appleupdates.py scrapes swscan.apple.com for catalogs. For now, I'm just checking the URL with the middleware for swscan.apple.com.

I appreciate the feedback and the patience. I am aware I'm not the worlds greatest python developer.

gregneagle:
Haven't forgotten about this; have some other things on my plate right now and will circle back to this as I get time.

gregneagle:
Current code looks really good. I have one concern. I don't know how "expensive" all the lookup and importing is. As the code stands now, it's done for every single URL request. It feels better/more efficient to do the lookup and import of the middleware module one time (perhaps for the first request, or when fetch.py is imported); then cache that state so you don't have to do it for every. single. subsequent request.

waderobson:
That makes sense. I'll take a crack at.

waderobson:
Ok have a look and let me know what you think. I've moved the searching and importing to the top of fetch. It should now search and import the middleware once, when fetch is imported.

gregneagle:
Time to start writing documentation and creating examples!

gregneagle:
I made a few subsequent commits to make things safer: I could easily crash Munki by adding a middleware.py file that did not have a 'process_request_options' function. The updates make sure a candidate "middleware*.py" file has a callable 'process_request_options' before 'really' importing it, and there are debug and warning messages to make it easier for people to debug the import of their middleware.

I'd like you to start some initial documentation and point to a working example middleware. Here's a really, really simple (and useless) one:

`middleware.py`
```python
def process_request_options(options):
    print options
    return options
```

waderobson:
Awesome!!. Thanks Greg!  I will work on documentation. I do have a working example [here](https://github.com/waderobson/s3-auth) and I have a google cloud storage one working as well just not posted. Now that the dust has settled I can focus on better documentation.

PR #587, opened 2016-04-06
Change Font CSS to include -apple-system

chrisgrande:
Automatically uses system font on El Capitan or newer.

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center/WebResources/base.css
+++ b/code/apps/Managed Software Center/Managed Software Center/WebResources/base.css
@@ -49,7 +49,7 @@ option {
 }
 
 body, button, input {
-    font-family: "Helvetica Neue", "Helvetica", "Arial";
+    font-family: -apple-system, "Helvetica Neue", "Helvetica", "Arial";
     font-smooth: always;
     -webkit-font-smoothing: antialiased
 }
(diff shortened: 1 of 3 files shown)


gregneagle:
I actually tried this. 1) I didn't like the look, and 2) Surprisingly, the Mac App Store does not do this. It uses Helvetica and Helvetica Neue.

chrisgrande:
I can't speak for you liking it of course, but San Fransisco has much better spacing at smaller sizes. Also the Mac App Store on El Capitan does seem to use it, see attached. 
app store fonts

gregneagle:
So ask for people on munki-dev and munki-discuss to try out your change and vote.

I looked very hard at similar views in the App Store on my non-Retina display and I thought the font looked a lot more like Helvetica than San Fran, but I could be wrong.

gregneagle:
(and though I can see a difference in your example picture, I think the extra spacing between letters in the middle example looks worse than the bottom example. Just my opinion, of course)

chrisgrande:
I'll post to the groups and see what others think, my basic idea behind the switch is that it will automatically use the system font that has been hinted for screen use and will continue to be up to date as Apple changes it – I also like San Fransisco as a System font. (In other news I am glad someone else updated the layout, I started kicking the CSS and just didn't have the time to finish)

gregneagle:
I wonder if the Mac App Store CSS changed in recent weeks. Looking at it today, it seems clearly (to me) to be San Francisco now, but a month ago I could have sworn it was Helvetica Neue.

chrisgrande:
I think they have been cleaning up a lot of the stores (iTunes, Mac) in the recent weeks - I've noticed a bunch of little spacing and alignment issues going away...

Issue #598, opened 2016-04-20
No shebang in conditional crashes Munki


waderobson:
Hello. I ran into this yesterday when I sent out a half-baked conditional to some of the fleet. 
To recreate this remove the top line of a conditional script. IOW no `#!/bin/bash` or `#!/usr/bin/python`.

```
Starting...
    Performing preflight tasks...
ERROR: Unexpected error in updatecheck:
Traceback (most recent call last):
  File "/usr/local/munki/managedsoftwareupdate", line 1048, in <module>
    main()
  File "/usr/local/munki/managedsoftwareupdate", line 742, in main
    updatecheckresult = updatecheck.check(client_id=options.id)
  File "/usr/local/munki/munkilib/updatecheck.py", line 2973, in check
    munkicommon.getConditions()
  File "/usr/local/munki/munkilib/munkicommon.py", line 2454, in getConditions
    utils.runExternalScript(conditionalscriptpath))
  File "/usr/local/munki/munkilib/utils.py", line 131, in runExternalScript
    stderr=subprocess.PIPE)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 8] Exec format error
```

I was able to catch it by doing this. 
https://gist.github.com/waderobson/36461224cd8ec8dad60e4315c7c4faf0#file-gistfile1-txt-L29-L40

Should I create a PR?

gregneagle:
Yes.

gregneagle:
Addressed in #599.

PR #599, opened 2016-04-20
Catch missing conditional shebang.

waderobson:
Fixing #598.

Tested. Output looks like this:
```
Starting...
    Performing preflight tasks...
/usr/local/munki/conditions/wan_ip.py can not be interpreted
Checking for available updates...
```

Have also tested that:
- You still receive this message when not executable
```
/usr/local/munki/conditions/wan_ip.py not executable
```
- The conditional still works.

### DIFF
--- a/code/client/munkilib/utils.py
+++ b/code/client/munkilib/utils.py
@@ -123,13 +123,18 @@ def runExternalScript(script, allow_insecure=False, script_args=()):
         cmd = [script]
         if script_args:
             cmd.extend(script_args)
-        proc = subprocess.Popen(cmd, shell=False,
-                                stdin=subprocess.PIPE,
-                                stdout=subprocess.PIPE,
-                                stderr=subprocess.PIPE)
-        (stdout, stderr) = proc.communicate()
-        return proc.returncode, stdout.decode('UTF-8', 'replace'), \
-                                stderr.decode('UTF-8', 'replace')
+        proc = None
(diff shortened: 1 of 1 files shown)


code/client/munkilib/utils.py
```diff
@@ -123,13 +123,18 @@ def runExternalScript(script, allow_insecure=False, script_args=()):
...
-        (stdout, stderr) = proc.communicate()
-        return proc.returncode, stdout.decode('UTF-8', 'replace'), \
-                                stderr.decode('UTF-8', 'replace')
+        proc = None
+        try:
+            proc = subprocess.Popen(cmd, shell=False,
+                                    stdin=subprocess.PIPE,
+                                    stdout=subprocess.PIPE,
+                                    stderr=subprocess.PIPE)
+        except OSError:
```
gregneagle: We should catch OSError _and_ IOError, and we should return the captured error and not make an assumption about what it is. Something like:

```python
        except (OSError, IOError), err:
            raise RunExternalScriptError(
                'Error %s when attempting to run %s' % (unicode(err), script))
```

waderobson:
You're right. Your snippet works perfectly. Here's the new output:
```
Starting...
    Performing preflight tasks...
Error [Errno 8] Exec format error when attempting to run /usr/local/munki/conditions/wan_ip.py
Checking for available updates...
```

PR #602, opened 2016-04-22
Refactor IOKit/PowerManager "no idle sleep" assertion source, in prep for reuse.

juofmgoblue:
This is especially helpful when downloading large packages and/or while
on slow network connections, avoiding "inactivity" causing machines to sleep
and package downloads to fail.

### DIFF
--- a/code/client/munkilib/installer.py
+++ b/code/client/munkilib/installer.py
@@ -41,6 +41,7 @@ from removepackages import removepackages
 # No name 'Foo' in module 'Bar' warnings. Disable them.
 # pylint: disable=E0611
 from Foundation import NSDate
+# pylint: enable=E0611
 
 # lots of camelCase names       
 # pylint: disable=C0103
--- a/code/client/munkilib/installer.py
+++ b/code/client/munkilib/installer.py
@@ -42,6 +42,9 @@ from removepackages import removepackages
 # pylint: disable=E0611
 from Foundation import NSDate
 
+# lots of camelCase names       
+# pylint: disable=C0103
+
 # initialize our report fields
 # we do this here because appleupdates.installAppleUpdates()
 # calls installWithInfo()
--- a/code/client/munkilib/installer.py
+++ b/code/client/munkilib/installer.py
@@ -31,6 +31,7 @@ import adobeutils
 import launchd
 import munkicommon
 import munkistatus
+import powermgr
 import profiles
 import updatecheck
 import FoundationPlist
@@ -40,41 +41,6 @@ from removepackages import removepackages
 # No name 'Foo' in module 'Bar' warnings. Disable them.
 # pylint: disable=E0611
 from Foundation import NSDate
-# stuff for IOKit/PowerManager, courtesy Michael Lynn, pudquick@github
-from ctypes import c_uint32, cdll, c_void_p, POINTER, byref
-from CoreFoundation import CFStringCreateWithCString
-from CoreFoundation import kCFStringEncodingASCII
-from objc import pyobjc_id
-# pylint: enable=E0611
-
-# lots of camelCase names
-# pylint: disable=C0103
-
-libIOKit = cdll.LoadLibrary('/System/Library/Frameworks/IOKit.framework/IOKit')
-libIOKit.IOPMAssertionCreateWithName.argtypes = [
-    c_void_p, c_uint32, c_void_p, POINTER(c_uint32)]
-libIOKit.IOPMAssertionRelease.argtypes = [c_uint32]
-
-def CFSTR(py_string):
-    '''Returns a CFString given a Python string'''
-    return CFStringCreateWithCString(None, py_string, kCFStringEncodingASCII)
-
-def raw_ptr(pyobjc_string):
-    '''Returns a pointer to a CFString'''
-    return pyobjc_id(pyobjc_string.nsstring())
-
-def IOPMAssertionCreateWithName(assert_name, assert_level, assert_msg):
-    '''Creaes a PowerManager assertion'''
-    assertID = c_uint32(0)
-    p_assert_name = raw_ptr(CFSTR(assert_name))
-    p_assert_msg = raw_ptr(CFSTR(assert_msg))
-    errcode = libIOKit.IOPMAssertionCreateWithName(
-        p_assert_name, assert_level, p_assert_msg, byref(assertID))
-    return (errcode, assertID)
-
-IOPMAssertionRelease = libIOKit.IOPMAssertionRelease
-# end IOKit/PowerManager bindings
-
 
 # initialize our report fields
 # we do this here because appleupdates.installAppleUpdates()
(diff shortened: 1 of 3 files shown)


gregneagle:
Is this wise? Laptop users might be annoyed if large packages run their battery to exhaustion. A failed partial download should be resumed on the next attempt.

bochoven:
I'm with @gregneagle here, people don't get feedback as to why their machine does not go to sleep when they expect it to. It might be different if the download is a result of a user initiated install request though..

juofmgoblue:
We've actually had users request this, due to repeated download failure when they weren't watching closely and touching the trackpad every few minutes.

I can see an argument where this may not be desirable for background downloads, but the fact that even attended/GUI downloads are impacted by idle seems like poor UX.  Unfortunately I don't see a way to easily surface the runtype to updatecheck.

Usability vs manageability tradeoffs aren't easy.  Perhaps it's better as a config, but I understand there's desire to keep the number of available knobs low as well.

nmcspadden:
What's the behavior of other common download platforms? Does the App Store completely preclude sleep? What about Safari / Firefox / Chrome downloads? I don't know 100% for sure, but I suspect they don't prevent idle sleep from happening even if a download is going.

I feel like consistency makes a bit more sense here, even if this is admin-provided. Given that Munki already has pretty robust download resume functionality, I personally lean towards letting the machine manage its sleep settings as expected, and resuming a download should the network connection drop.

juofmgoblue:
App Store has Power Nap support and will download a "few minutes" per cycle per https://support.apple.com/en-us/HT204032

juofmgoblue:
One possibility would be to selectively assert no sleep only when on AC Power, using something like this: https://github.com/munki/munki/blob/master/code/apps/Managed%20Software%20Center/Managed%20Software%20Center/munki.py#L549

erikng:
I think the AC Power no sleep is a very smart idea.

flammable:
We do that - as part of our initial imaging, we disable sleep for AC power:

`/usr/bin/pmset -c sleep 0`

Otherwise, in some of our labs, the machine would go to sleep during the bootstrapping phase - particularly in the labs that have a large amount of Adobe packages.

This issue hasn't affected us since.  If we could limit it to AC power, I think it'd be worth merging into Munki.

grahamgilbert:
I'm with @flammable - I would say this would be a useful option (and it should definitely be an option that the Munki admin needs to opt into) when on AC power, but this should never be an option on battery power.

rmanly:
Also with Mike. I've done the same here in various ways over the years. When deploying to labs it is pretty frustrating if they sleep while downloading large packages.

gregneagle:
So lets collect some thoughts:

1) Seems to be a consensus that doing a no-sleep assertion when on AC power is low-risk and acceptable.

2) When an end-user uses Managed Software Center to manually check for updates or to install "self-serve" items (optional_installs), the "manualcheck" option is set: https://github.com/munki/munki/blob/master/code/client/managedsoftwareupdate#L479. So it would be possible to also do a no-sleep assertion when a user is interacting with MSC.app

hfike:
1 and 2 are logical and if I were new to the project, I would expect that behavior. Number 3 I think makes sense as well. If I click install, I want it to install right then. My expectation are that the computer will do the job, not go to sleep in the middle.

An optional preference to change the behavior of the no-sleep on battery could be useful for bootstrapping in certain situations. If during your bootstrapping process for laptops you drop the ManagedInstall.plist with the "no-sleep-on-battery" preference true, you could change it later. Since MBP's generally ship with enough power to bootstrap without AC, some could find this useful.
I have a laptop cart that is on charge generally, but when I reimage them, I generally just plug them into the network on battery, instead of pulling the chargers out of the cart.

arubdesu:
I'm of the opinion that 'the mail must go through' - safe sleep can help customers recover from running out of power. Maybe that's me being BOFH-y, but maybe not. I see aggressive app nap utilization from things like Outlook, so users (with laptops) will be trained to expect things to happen with lids closed and in bags.

In lieu of taking the PR as-is, it seems like the upsides most people want can be enabled without an admin pref if we also cover the case of 'checkandinstallatboot' (in addition to a GUI run, and on-AC-power runs).

flammable:
If we could check for "custom" too, that'd handle running
managedsoftwareupdate via ARD.

Mike

-- 

mike@mikesolin.com / @flammable
http://mikesolin.com

On Apr 23, 2016, at 6:13 PM, Allister Banks <notifications@github.com>
wrote:

I'm of the opinion that 'the mail must go through' - safe sleep can help
customers recover from running out of power. Maybe that's me being BOFH-y,
but maybe not. I see aggressive app nap utilization from things like
Outlook, so users (with laptops) will be trained to expect things to happen
with lids closed and in bags.

In lieu of taking the PR as-is, it seems like the upsides most people want
can be enabled without an admin pref if we also cover the case of
'checkandinstallatboot' (in addition to a GUI run, and on-AC-power runs).

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/munki/munki/pull/602#issuecomment-213841992>

juofmgoblue:
I've rescoped this pull request to refactoring powermgr for reuse, reverting package download portion and leaving functionality in line with head.  I will follow up in a future pull request including selective use of this within download_installeritem().  Please review.

code/client/munkilib/installer.py
```diff
@@ -40,41 +41,6 @@
...
 from Foundation import NSDate
-# stuff for IOKit/PowerManager, courtesy Michael Lynn, pudquick@github
-from ctypes import c_uint32, cdll, c_void_p, POINTER, byref
-from CoreFoundation import CFStringCreateWithCString
-from CoreFoundation import kCFStringEncodingASCII
-from objc import pyobjc_id
-# pylint: enable=E0611
-
-# lots of camelCase names
-# pylint: disable=C0103
```
gregneagle: After moving this code into powermgr, we don't have any more camelCase names?

code/client/munkilib/powermgr.py
```diff
@@ -0,0 +1,74 @@
...
+powermgr.py
+munki module to toggle IOKit/PowerManager idle sleep assertions.
+"""
+
+# stuff for IOKit/PowerManager, courtesy Michael Lynn, pudquick@github
+from ctypes import c_uint32, cdll, c_void_p, POINTER, byref
+from CoreFoundation import CFStringCreateWithCString
+from CoreFoundation import kCFStringEncodingASCII
+from objc import pyobjc_id
+# pylint: enable=E0611
```
gregneagle: Flipside of the issue in installer.py...

juofmgoblue: Fixed, I think...

juofmgoblue:
PTAL.  I think I got the pylint enable/disable comments back where they should be.

cwindus:
I'm curious as to what the outcome of this has been? I've been running into bootstrapping desktop machines with munki 2.8.2.2855 where the machine just sleeps in the middle of downloading packages.

I can see the assertions being created during installs, but they're not being created during downloads of packages.

I've often thought it would be handy to have a bootstrap argument that would ensure no sleep during downloads and installs, and also remove the 'Stop' button from the login window status bar.

PR #616, opened 2016-06-07
Centrify repo backend

centrify-keithr:
We would like to be able to use the munki command-line tools such as munkiimport to access a repository stored on our cloud servers. Rather than make Centrify-specific changes to the munki code, this change adds the ability to configure the munki command-line tools to use an external program to access the repository. We will create a Centrify-specific program that accesses the repository on our cloud servers. Implementing it in this way also allows for the creation of external programs to use other mechanisms to access the repository, such as webdav.

### DIFF
--- a/code/client/iconimporter
+++ b/code/client/iconimporter
@@ -29,12 +29,13 @@ from optparse import OptionParser
 from munkilib import munkicommon
 from munkilib import FoundationPlist
 from munkilib import iconutils
+from munkilib import Repo
 
 
-def generate_png_from_copy_from_dmg_item(install_item, repo_path):
-    dmgpath = os.path.join(
-        repo_path, 'pkgs', install_item['installer_item_location'])
-    mountpoints = munkicommon.mountdmg(dmgpath)
+def generate_png_from_copy_from_dmg_item(install_item, repo):
+    dmgpath = repo.join('pkgs', install_item['installer_item_location'])
+    handle = repo.open(dmgpath, 'r')
+    mountpoints = munkicommon.mountdmg(handle.local_path)
     if mountpoints:
         mountpoint = mountpoints[0]
         apps = [item for item in install_item.get('items_to_copy', [])
@@ -43,9 +44,9 @@ def generate_png_from_copy_from_dmg_item(install_item, repo_path):
             app_path = os.path.join(mountpoint, apps[0]['source_item'])
             icon_path = iconutils.findIconForApp(app_path)
             if icon_path:
-                png_path = os.path.join(
-                    repo_path, u'icons', install_item['name'] + u'.png')
-                result = iconutils.convertIconToPNG(icon_path, png_path)
+                png_path = repo.join(u'icons', install_item['name'] + u'.png')
+                png_handle = repo.open(png_path, 'w')
+                result = iconutils.convertIconToPNG(icon_path, png_handle.local_path)
                 if result:
                     print_utf8(u'\tWrote: %s' % png_path)
                 else:
@@ -57,15 +58,15 @@ def generate_png_from_copy_from_dmg_item(install_item, repo_path):
         munkicommon.unmountdmg(mountpoint)
 
 
-def generate_pngs_from_installer_pkg(install_item, repo_path):
+def generate_pngs_from_installer_pkg(install_item, repo):
     icon_paths = []
     mountpoint = None
     pkg_path = None
-    item_path = os.path.join(
-        repo_path, u'pkgs', install_item['installer_item_location'])
+    pkg_repo = None
+    item_path = repo.join( u'pkgs', install_item['installer_item_location'])
     if munkicommon.hasValidDiskImageExt(item_path):
-        dmg_path = item_path
-        mountpoints = munkicommon.mountdmg(dmg_path)
+        handle = repo.open(item_path, 'r')
+        mountpoints = munkicommon.mountdmg(handle.local_path)
         if mountpoints:
             mountpoint = mountpoints[0]
             if install_item.get('package_path'):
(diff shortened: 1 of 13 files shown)


gregneagle:
That's a pretty serious set of changes. I can't promise I'll have the time to properly review them for a while. It would be helpful for you to post something on munki-dev so that other interested parties will be aware and can take a look.

code/client/munkilib/CommandRepo.py
```diff
@@ -0,0 +1,311 @@
+<<<<<<< HEAD
```
clburlison: You seem to have quite a few errors in this file from a git merge.

code/client/munkilib/FileRepo.py
```diff
@@ -0,0 +1,246 @@
+<<<<<<< HEAD
```
clburlison: Same git merge issues from `code/client/munkilib/CommandRepo.py`

pehavens:
Thanks for the comment Greg. This is Peter Havens from Centrify. Met you a couple times at some of the Mac conferences. Wanted to give you a heads up on what the changes are about, and possibly see if it would make sense to discuss them, and see if it makes sense as an addition / option to the project.  The purpose of the modifications we’re proposing is to enable us to use munki tools such as munkiimport to access a repository stored on our cloud servers. Rather than make Centrify-specific changes to the munki code, we’ve added the ability to configure the munki command-line tools to use an external program to access the repository. We will then create a Centrify-specific external program that accesses the repository on our cloud servers. Implementing it in this way also allows for the creation of external programs to use other mechanisms to access the repository, such as webdav. I will ask Keith to post something in munki-dev as well so others can take a look.

Thanks!

-- Peter

gregneagle:
One quick concern I have is that I and others need `makecatalogs` to be able to run on Linux; currently `makecatalogs` has no hard dependencies other than Python 2.6+ standard library.

groob:
@centrify-keithr @pehavens 
I'm interested in a generic mechanism to access a remote munki repository, but I'm not sure about this particular implementation.

Here's roughly what I'm doing:
I've [patched](https://github.com/munki/munki/compare/master...groob:netimport) a few `munkiimport` method to shell out to a separate cli command which interacts with the [munki api server](https://github.com/munki/mwa2/wiki/API-notes). The patched version works, although it's not abstracted enough for a pull request. 

Would you mind documenting how the changes in this PR would allow for different backend solutions to be implemented? What will the interface(RPC or otherwise) look like? If we're shelling out to an external command, what are the flags/stdin methods it should expect? 
What about things like API credentials/authentication with the backend?

groob:
Looking briefly at `CommandRepo`, and `Repo`, I'm not quite clear how the abstraction would allow for someone to write an API backend, an S3 backend, a SQL backend etc. 

My main concern is that this PR adds a lot of changes to munki but does not yet provide a good generic abstraction for dealing with a repository which is quite different than files on a local or mounted filesystem.

centrify-keithr:
On Jun 6, 2016, at 6:22 PM, Greg Neagle <notifications@github.com<mailto:notifications@github.com>> wrote:


One quick concern I have is that I and others need makecatalogs to be able to run on Linux; currently makecatalogs has no hard dependencies other than Python 2.6+ standard library.

This shouldn’t change that, at least not for the file-based repositories that are currently supported. If the linux machine is configured to use an external command to access the repository, that command must exist on the machine, and run on linux, of course.

centrify-keithr:
On Jun 6, 2016, at 7:52 PM, Victor <notifications@github.com<mailto:notifications@github.com>> wrote:


Looking briefly at CommandRepo, and Repo, I'm not quite clear how the abstraction would allow for someone to write an API backend, an S3 backend, a SQL backend etc.

My main concern is that this PR adds a lot of changes to munki but does not yet provide a good generic abstraction for dealing with a repository which is quite different than files on a local or mounted filesystem.

My goal was narrower than that, since the repository we need access to does look mostly like files on a remote filesystem. I haven’t put a lot of thought into what would be required to support repositories that are significantly different. One obvious thing would be changing CommandRepo to call the external command for things like “join” and “dirname” rather than assuming it treats paths within the repository as UNIX-like.

Our command will use a client certificate from a local keychain, so it doesn’t need a way to pass authentication through from munkiimport et. al., but that could be needed for other backends.

Do you have thoughts on what other changes would be required to make the abstraction sufficiently generic?

Would you mind documenting how the changes in this PR would allow for different backend solutions to be implemented? What will the interface(RPC or otherwise) look like? If we're shelling out to an external command, what are the flags/stdin methods it should expect?
What about things like API credentials/authentication with the backend?

The external command gets the repo URL (from the repo_url configuration property) as the first command-line argument, an operation to perform as the second, and the arguments to that operation as the remaining arguments. It does not receive anything via stdin; some of the commands output to stdout (e.g. listdir, walk). The UNIX zero/nonzero exit status convention is used to communicate success or failure. As mentioned above, there is no provision for passing credentials to the command, but that can be changed.

The operations are the ones implemented in CommandRepo.py; I can write a brief description of each one if you like.

code/client/munkilib/CommandRepo.py
centrify-keithr: On Jun 6, 2016, at 5:55 PM, Clayton Burlison <notifications@github.com<mailto:notifications@github.com>> wrote:


In code/client/munkilib/CommandRepo.py<https://github.com/munki/munki/pull/616#discussion_r65995187>:

> @@ -0,0 +1,311 @@
> +<<<<<<< HEAD


You seem to have quite a few errors in this file from a git merge.

Weird. I didn’t get any messages about conflicts from the git command. I’ve pushed a fix for this.

ygini:
Why not using a simple pre and post action?

Before and munkiimport/makecatalog, a tool of your choice can be called to update a working copy of your remote storage (pull last version before, push changes after).

That's what I do with git based storage. 

Like that, you do intermediate work on your system and talk with the server only when things are done. It should save bandwidth and ensure atomicity of munki repo.

code/client/munkiimport
```diff
@@ -441,11 +394,16 @@ class CatalogDBException(Exception):
 def make_catalog_db():
     """Returns a dict we can use like a database"""
 
-    all_items_path = os.path.join(REPO_PATH, 'catalogs', 'all')
-    if not os.path.exists(all_items_path):
+    all_items_path = repo.join('catalogs', 'all')
+    handle = None
```
jgstew: Does handle really need to be declared like this first?

code/client/munkiimport
```diff
@@ -664,12 +622,16 @@ def make_catalogs():
...
     if not VERBOSE:
         print 'Rebuilding catalogs at %s...' % REPO_PATH
-    proc = subprocess.Popen([makecatalogs_path, REPO_PATH],
-                            bufsize=-1, stdout=subprocess.PIPE,
+    cmd = [makecatalogs_path]
+    if REPO_URL:
+        cmd.append('--repo-url')
+        cmd.append(REPO_URL)
+    cmd.append(REPO_PATH)
+    proc = subprocess.Popen(cmd, bufsize=-1, stdout=subprocess.PIPE,
```
jgstew: Can you explain and document this change better?

code/client/iconimporter
```diff
@@ -202,13 +209,24 @@ def main():
...
+        repo_url = pref('repo_url')
+        if not repo_url:
+            print_err_utf8("Need to specify a URL for the repo!")
+            exit(-1)
+        else:
+            print_utf8("Using repo url: %s" % repo_url)
+
+    # Make sure the repo exists
+    repo = Repo.Open(repo_path, repo_url)
+    if not repo.available():
```
jgstew: This change needs better explanation and documentation.

gregneagle:
Closing this as it's been replaced by #689

PR #621, opened 2016-06-21
Adding the ability to Perform FileVault Authorized Restarts. 

weswhet:
This adds the ability to perform AuthRestarts via the fdesetup command. It requires two Preferences be defined PerformAuthRestarts(bool) and RecoveryKeyFile which is a path to a plist where the recovery key is stored that will be used to perform the restart. The RecoveryKeyFile needs to contain a Key labeled RecoveryKey(this is the default name provided by fdesetup's outputplist feature) with a value containing either the individual recovery key or a password of an account enabled in FileVault. This feature works extremely well for major OS X upgrades as it now mimics that of the AppStore. I'll happily write the full documentation for this feature. 

munki-dev discussion. https://groups.google.com/forum/#!topic/munki-dev/iYBW5qM9LXM

cc @keeleysam @grahamgilbert

### DIFF
--- a/code/client/managedsoftwareupdate
+++ b/code/client/managedsoftwareupdate
@@ -279,6 +279,17 @@ def doRestart():
     if not consoleuser or consoleuser == u'loginwindow':
         # no-one is logged in or we're at the loginwindow
         time.sleep(5)
+        #check to see if we should perform an AuthRestart
+        if (munkicommon.pref('PerformAuthRestarts')
+                and munkicommon.pref('RecoveryKeyFile')):
+            munkicommon.log('Starting Authorized Restart...')
+            # try to perform an auth restart
+            munkicommon.perform_auth_restart()
+            time.sleep(2)
+            # if we got to here then the auth restart failed notify that it did
+            # then perform a normal restart
+            munkicommon.display_warning(
+                'Authorized Restart Failed... Performing normal restart')
         dummy_retcode = subprocess.call(['/sbin/shutdown', '-r', 'now'])
     else:
         if munkicommon.munkistatusoutput:
--- a/code/client/managedsoftwareupdate
+++ b/code/client/managedsoftwareupdate
@@ -286,10 +286,10 @@ def doRestart():
             # try to perform an auth restart
             munkicommon.perform_auth_restart()
             time.sleep(2)
-            # if we got to here then the auth restart failed notify that it did
+            # if we got to here then the auth restart failed, notify that it did
             # then perform a normal restart
             munkicommon.display_warning(
-                'Authorized Restart Failed... Performing normal restart')
+                'Authorized Restart Failed. Performing normal restart...')
         dummy_retcode = subprocess.call(['/sbin/shutdown', '-r', 'now'])
     else:
         if munkicommon.munkistatusoutput:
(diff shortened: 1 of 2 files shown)


gregneagle:
I assume it is up to the admin to securely get/create/handle the RecoveryKeyFile, since it's really dangerous to have around.

code/client/managedsoftwareupdate
```diff
@@ -279,6 +279,17 @@ def doRestart():
     if not consoleuser or consoleuser == u'loginwindow':
         # no-one is logged in or we're at the loginwindow
         time.sleep(5)
+        #check to see if we should perform an AuthRestart
```
gregneagle: Don't you also need to check to see if FileVault is enabled? What happens if moron admin specifies these keys, yet FileVault is not enabled/active on the boot disk?

weswhet:
Yes, I'm open to adding something that makes it more secure. i.e root/read only but I figured it would be best done by the admin.

code/client/managedsoftwareupdate
weswhet: The supports_auth_restarts function should handle that.

gregneagle:
I think the only thing needed is to make it abundantly clear in the documentation for this feature how insecure having the RecoveryKeyFile around is.

gregneagle:
What happens when authrestart is attempted with a RecoveryKeyFile that contains a key that is not actually valid or a password that is not actually valid?

weswhet:
Agreed on the documentation. If the key isn't valid fdesetup returns an Authentication Error message.

gregneagle:
And then what? I think it's really important that if the authrestart attempt fails, that it falls back to a "normal" restart.

weswhet:
That's the default behavior yes.

gregneagle:
and you've tested it? This looks like it _should_ work, but looks can be deceiving.

```python
        if (munkicommon.pref('PerformAuthRestarts')
                and munkicommon.pref('RecoveryKeyFile')):
            munkicommon.log('Starting Authorized Restart...')
            # try to perform an auth restart
            munkicommon.perform_auth_restart()
            time.sleep(2)
            # if we got to here then the auth restart failed notify that it did
            # then perform a normal restart
            munkicommon.display_warning(
                'Authorized Restart Failed... Performing normal restart')
        dummy_retcode = subprocess.call(['/sbin/shutdown', '-r', 'now'])
```

weswhet:
Yes, I've tested all major scenarios including that off PerformAuthRestarts and RecoveryKeyFile are defined correctly but the RecoveryKey isn't valid. It proceeds to perform a normal restart.

weswhet:
This is what it looks like when you PerformAuthRestarts enabled on a non encrypted machine.. `fdesetup supportsauthrestart` returns true even when filevault is turned off. Still gives an appropriate error and performs a normal restart. 
screen shot 2016-06-21 at 5 13 28 pm

weswhet:
Thoughts on moving this into a separate branch so I can do wider testing with the autobuilds?

gregneagle:
That sounds like a good idea. I'll see if I can manage it.

gregneagle:
Pulled to new branch: https://github.com/munki/munki/tree/fv-auth-restart

gregneagle:
I'm going to close this PR; we can decide later if/when to merge the fv-auth-restart branch to master.

PR #636, opened 2016-07-12
Added ignore for line length as we don't enforce this currently

natewalck:
Leave out line length as we don't care about that.


natewalck:
Also added a proper exit code if there are lint errors so the build status shows correctly.

PR #641, opened 2016-07-28
Add support for `makepkginfo` and new Adobe installers

timsutton:
This adds initial support for the new Adobe "HyperDrive" installers released mid-June 2016, for `makepkginfo`. CCP packages built that include any of these new product versions include this new installer engine that's used for only those new versions. Since a CCP package can contain a mix of many different applications, many "monolithic" CCP packages made today will contain both the legacy "RIBS" and the newer installer formats.

Some discussion happened [earlier on munki-dev](https://groups.google.com/d/topic/munki-dev/5KzSwVIDbp4/discussion) on this support, and I believe @foigus, @clburlison, and @bochoven have tested this code (as have I used it to build pkginfos for all the new installers).

What this code does currently:

- generates `installs` items for the newer applications by assembling a special metadata path, very similar to how the `Uninstall.db` paths work from CS5 and up
- if the package contains a single app, extract sane `version` and `display_name` keys from the package metadata

I haven't found it necessary to add any code to the installer process itself. Because the new installers log to new (and unpredictable, thanks!) locations I've not made attempts to attach to these logs to follow installation progress. However, the HyperDrive installers are so much faster (i.e. a 2 GB app installed in 30 seconds) that I think it's less of an issue. Not ruling out the idea but just didn't want to add a lot more code that would need to be thoroughly tested just for adding progress info.

In the future I think it would be also helpful to extract `installed_size` by counting the relevant payloads in the metadata, as well as extracting actual apps to build `installs` application types if possible - the latter seeming easier now than it did with the old RIBS data. However I figured it would be better to get some initial functional support reviewed and brought in first, and deal with those details later.

This brings Adobe creative packages support in pretty much to parity with where it was before (with the added advantage of having better `version` info generated wherever possible, compared to prior CCP package support).

### DIFF
--- a/code/client/munkilib/adobeutils.py
+++ b/code/client/munkilib/adobeutils.py
@@ -1284,7 +1284,7 @@ def getAdobeCatalogInfo(mountpoint, pkgname=""):
                 product_saps = [
                     prod['SAPCode'] for
                     prod in option_xml_info['products']
-                    if prod['MediaType'] == 'Product'
+                    if prod.get('MediaType') == 'Product'
                 ]
                 for app_info in [app for app in hd_app_infos
                                  if app['SAPCode'] in product_saps]:
--- a/code/client/munkilib/adobeutils.py
+++ b/code/client/munkilib/adobeutils.py
@@ -1264,11 +1264,12 @@ def getAdobeCatalogInfo(mountpoint, pkgname=""):
                     os.path.join(dirpath, 'HD'), sap_code)
                 hd_app_infos.append(hd_app_info)
 
-            # installs keys will be populated if we have either RIBS
+            # 'installs' array will be populated if we have either RIBS
             # or HD installers, which may be mixed together in one
-            # CCP package
-            if mediasignatures or hd_app_infos:
-                installs = []
+            # CCP package.
+            # Acrobat Pro DC doesn't currently generate any useful installs
+            # info if it's part of a CCP package.
+            installs = []
 
             # media signatures are used for RIBS (CS5 to CC mid-2015)
             if mediasignatures:
@@ -1319,7 +1320,8 @@ def getAdobeCatalogInfo(mountpoint, pkgname=""):
                             installitem['type'] = 'file'
                             installs.append(installitem)
 
-            cataloginfo['installs'] = installs
+            if installs:
+                cataloginfo['installs'] = installs
 
             return cataloginfo
 
--- a/code/client/munkilib/adobeutils.py
+++ b/code/client/munkilib/adobeutils.py
@@ -1251,17 +1251,6 @@ def getAdobeCatalogInfo(mountpoint, pkgname=""):
                 mediasignature = cataloginfo['adobe_install_info'].get(
                     "media_signature")
                 mediasignatures = [mediasignature]
-            if mediasignatures:
-                # make a default <key>installs</key> array
-                uninstalldir = "/Library/Application Support/Adobe/Uninstall"
-                installs = []
-                for mediasignature in mediasignatures:
-                    signaturefile = mediasignature + ".db"
-                    filepath = os.path.join(uninstalldir, signaturefile)
-                    installitem = {}
-                    installitem['path'] = filepath
-                    installitem['type'] = 'file'
-                    installs.append(installitem)
 
             # Determine whether we have HD media as well in this installer
             hd_metadata_dirs = [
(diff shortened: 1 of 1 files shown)


timsutton:
Since Adobe has done updates of a few of these new versions, it looks like this path generated by `makepkginfo` isn't quite correct.

For example, the new release of Premiere Pro CC 2015 (10.4) `makepkginfo` generates a `pimx` file

`PPRO_10_3_0_AdobePremierePro10AllTrial_10.4.0.30.pimx`

..but installing the package makes this:

`PPRO_10_4_0_AdobePremierePro10AllTrial_10.4.0.30.pimx`

@clburlison reported a similar difference with MUSE:

`MUSE_2015_2_1_AdobeMuse2015.2.0-core_2015.2.1.21.pimx` <-- installed on disk
`MUSE_2015_2_0_AdobeMuse2015.2.0-core_2015.2.1.21.pimx` <-- generated by makepkginfo

It seems to me like assembling the path with the "base version" at the beginning we actually need to use the "product version".

timsutton:
And sadly it's not even that simple, because `BaseVersion` in the Application.json files are more specific (`10.4.0.30`) than `baseVersion` in .ccp and .pimx files (`10.4.0`). And we weren't already parsing .ccp files.

Rather than having to parse another file altogether, however, it seems we can extract the shorter `baseVersion` string from the directory containing all this package metadata (`PPRO10.4.0`).

clburlison:
Thanks @timsutton the latest commit fixed Muse, Premiere, and XD (IE- the only HD updates at this time). Now we just have to hope Adobe doesn't break it in the future.

gregneagle:
Keep working on this stuff -- I won't start looking at merging it until after the 2.8.0 official release happens. (IOW I'm not ignoring you)

gregneagle:
"Now we just have to hope Adobe doesn't break it in the future." It's far more likely they will then they won't.

timsutton:
@clburlison Not the only ones - Bridge and ACR happened a couple weeks ago and recently:

https://twitter.com/foigus/status/761284204621475840

foigus:
An "All Apps" package from the Enterprise Cloud Packager (website) dies on the version of makepkginfo in this branch:

```
$ /usr/local/munki/makepkginfo /Users/macadmin/Desktop/All\ Apps/Build/All\ Apps_Install.pkg 
  File "/usr/local/munki/makepkginfo", line 1166, in <module>
    main()
  File "/usr/local/munki/makepkginfo", line 872, in main
    catinfo = getCatalogInfoFromPath(item, options)
  File "/usr/local/munki/makepkginfo", line 107, in getCatalogInfoFromPath
    cataloginfo = adobeutils.getAdobeCatalogInfo(pkgpath)
  File "/usr/local/munki/munkilib/adobeutils.py", line 1318, in getAdobeCatalogInfo
    if prod['SAPCode'] == app_info['SAPCode']][0]
KeyError: 'SAPCode'
```

[Application_dot_json.zip](https://github.com/munki/munki/files/409392/Application_dot_json.zip)
[optionXML_xml.txt](https://github.com/munki/munki/files/409394/optionXML_xml.txt)

Gerk:
I can second what @foigus just posted.

```
./makepkginfo /Users/mark/Downloads/All\ Apps/Build/All\ Apps_Install.pkg 
  File "./makepkginfo", line 1166, in <module>
    main()
  File "./makepkginfo", line 872, in main
    catinfo = getCatalogInfoFromPath(item, options)
  File "./makepkginfo", line 107, in getCatalogInfoFromPath
    cataloginfo = adobeutils.getAdobeCatalogInfo(pkgpath)
  File "/Users/mark/code/munki/code/client/munkilib/adobeutils.py", line 1318, in getAdobeCatalogInfo
    if prod['SAPCode'] == app_info['SAPCode']][0]
KeyError: 'SAPCode'
```

Gerk:
The above issues is resolved with this pull request:  https://github.com/munki/munki/pull/649

timsutton:
This filter logic seems to me like it might be assuming that there's only one major version of a given application in a package. I wonder what happens if someone would build a package containing say, Photoshop CC 2015.5 and CC 2015.0, or CC 2014. Seems like a rare case, but still..

timsutton:
As suspected, this logic has the possibility of generating the wrong metadata for HD installers for packages containing multiple versions of apps with the same SAPCode. I'll see if we can narrow this filter down to also test for the "major" product versions.

PR #644, opened 2016-07-28
fix SecureManagedInstallsPreferences

maximermilov:
after munki@93b0d07
CFPreferencesCopyAppValue will return old value (set by old client), so
CFPreferencesCopyValue with matching user/host values should be used instead.

### DIFF
--- a/code/client/munkilib/munkicommon.py
+++ b/code/client/munkilib/munkicommon.py
@@ -57,6 +57,7 @@ from Foundation import NSDate, NSMetadataQuery, NSPredicate, NSRunLoop
 from Foundation import CFPreferencesAppSynchronize
 from Foundation import CFPreferencesCopyAppValue
 from Foundation import CFPreferencesCopyKeyList
+from Foundation import CFPreferencesCopyValue
 from Foundation import CFPreferencesSetValue
 from Foundation import kCFPreferencesAnyUser
 from Foundation import kCFPreferencesCurrentUser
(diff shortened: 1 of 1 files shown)


gregneagle:
Changing from CFPreferencesCopyAppValue to CFPreferencesCopyValue breaks managed preferences.

maximermilov:
How about overriding __getitem__/__contains__ in SecureManagedInstallsPreferences ?

maximermilov:
I just updated pull request.

gregneagle:
No, this still doesn't address the managed preferences issue.

So let's back up: what are you trying to address here? If the recent change breaks something for you, please describe it so I can understand. (I don't really understand "CFPreferencesCopyAppValue will return old value (set by old client)"), but this might be the whole point here:

ByHost vs non-ByHost preferences. The code should have _never_ been writing to the ByHost preferences. But it did. With the recent change, it now (correctly) writes to the "normal" (Any Host) preferences. But the downside is that any values previously written to the ByHost preferences "override" those written at the Any Host level.

So it well may be that this fix is not worth the trouble it causes; even though the original code was doing the wrong thing -- it may just have to be something we document and live with.

And if that's the case, why is the fix not to just go back to the way it was before? Why change to yet another implementation?

maximermilov:
> any values previously written to the ByHost preferences "override" those written at the Any Host level.

exactly.
even if preferences at the Any Host level written after preference at the ByHost level, CFPreferencesCopyAppValue will return old value.

CFPreferencesCopyValue called with kCFPreferencesAnyHost returns desired value.

gregneagle:
But you have not addressed my last point: why would we not just undo my change and revert to the prior behavior?

maximermilov:
We use SecureManagedInstallsPreferences to store temporary values between preflight/postflight.
no need in migration script in our use case.

> why would we not just undo my change and revert to the prior behavior?

Both fixes will work.

gregneagle:
Reverted change here: https://github.com/munki/munki/commit/49457d5aa2e57255bb8fac8e11031929cf80999b

Added doc strings to explain the unintuitive behaviors:
https://github.com/munki/munki/commit/40309f57069afbe83c5e1045a5344f374b381939

No reason to move forward with the PR.

PR #647, opened 2016-08-08
Opt-in behavior to ignore system proxies.

### LINKED ISSUE #646: Add preference to bypass system proxy
I'd like for my Munki clients to ignore system proxy settings and connect directly to my Simian server.

This should be a straightforward change since [gurl.py](https://github.com/munki/munki/blob/master/code/client/munkilib/gurl.py) is using NSURLSession:

```
from CFNetwork import kCFNetworkProxiesHTTPSEnable
[...]
configuration.setConnectionProxyDictionary_({ kCFNetworkProxiesHTTPSEnable: False })
```

I suspect many (most?) people won't want this configuration though, and would prefer their munki connections go through the (managed) proxy, so I propose exposing a default-off flag ("NoProxy", perhaps?) in [ManagedInstalls.plist](https://github.com/munki/munki/blob/master/Preferences/ManagedInstalls.plist). 

Let me know if this change sounds good and I'll send a PR.

verycarefully:
Fix #646

### DIFF
--- a/code/client/munkilib/fetch.py
+++ b/code/client/munkilib/fetch.py
@@ -184,9 +184,12 @@ def get_url(url, destinationpath,
         cache_data = gurl_obj.get_stored_headers()
         del gurl_obj
 
+    ignore_system_proxy = munkicommon.pref('IgnoreSystemProxies')
+
     options = {'url': url,
                'file': tempdownloadpath,
                'follow_redirects': follow_redirects,
+               'ignore_system_proxy': ignore_system_proxy,
                'can_resume': resume,
                'additional_headers': header_dict_from_list(custom_headers),
                'download_only_if_changed': onlyifnewer,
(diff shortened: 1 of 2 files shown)


gregneagle:
This looks good; I won't merge until after the 2.8.0 release, which (knock on wood) should happen later this week.

gregneagle:
In the meantime, could you address the linting/PEP-8 errors noted here: https://travis-ci.org/munki/munki/builds/150682883

gregneagle:
It occurs to me that we'll want to document that this feature only works on OS versions that support NSURLSession, that is 10.9+.

natewalck:
As for the lint errors, you can run them manually by following these directions:

https://github.com/munki/munki/tree/master/tests#getting-started

That is the exact script that runs in travis CI, so you'll know if it is happy or not.

gregneagle:
@verycarefully Can you please add documentation for this new feature to the wiki?

verycarefully:
Done: appended an entry to https://github.com/munki/munki/wiki/Preferences

Thanks!

PR #674, opened 2016-10-17
update makecatalogs to function standalone with python 3

keeleysam:
this updates makecatalogs to be compatible with python 3.5 while retaining current functionality on python 2.6 and 2.7.  

note - this only works if makecatalogs is standalone, because foundationplist and the other imports from munkilib aren't updated yet.

### DIFF
--- a/code/client/makecatalogs
+++ b/code/client/makecatalogs
@@ -41,8 +41,8 @@ except ImportError:
         LOCAL_PREFS_SUPPORT = True
     except ImportError:
         # maybe we're not on an OS X machine...
-        print >> sys.stderr, ("WARNING: FoundationPlist is not available, "
-                              "using plistlib instead.")
+        sys.stderr.write("WARNING: FoundationPlist is not available, "
+                         "using plistlib instead.\n")
         import plistlib
         LOCAL_PREFS_SUPPORT = False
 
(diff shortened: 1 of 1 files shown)


keeleysam:
(i need to do a little more work on unicode handling for filenames, this is functional but unicode filenames trip up on python2 right now.)

gregneagle:
"(i need to do a little more work on unicode handling for filenames, this is functional but unicode filenames trip up on python2 right now.)"

I can't merge this if it causes regressions under Python 2.6 or Python 2.7. You are probably the only person trying to run this under Python 3.x -- I'd rather you stay broken than break everyone else!

code/client/makecatalogs
```diff
@@ -41,8 +41,8 @@ except ImportError:
         LOCAL_PREFS_SUPPORT = True
     except ImportError:
         # maybe we're not on an OS X machine...
-        print >> sys.stderr, ("WARNING: FoundationPlist is not available, "
-                              "using plistlib instead.")
+        sys.stderr.write("WARNING: FoundationPlist is not available, "
+                         "using plistlib instead.")
```
gregneagle: With the .write method I think you'll also need to add a trailing `\n'.

keeleysam:
yeah this is WIP, currently testing out the error handling.   it works if you don't have any though :)

code/client/makecatalogs
```diff
@@ -70,10 +70,13 @@ except ImportError:
...
-            path = unicode(path, 'utf-8')
-        elif type(path) is not unicode:
-            path = unicode(path)
+        try:
+            if isinstance(path, str):
+                path = str(path, 'utf-8')
+            elif not isinstance(path, str):
+                path = str(path)
+        except TypeError:
+            path = str(path)
```
gregneagle: I don't see how this works on both Python2 and Python3.

keeleysam:
okay i've tested errors with missing files, unicode filenames, unicode strings, etc and haven't identified any regressions.

gregneagle:
Great. Make sure you (or some people in the community) test on 10.6 as well. I don't want a repeat of the Munki tools 2.8.1 release, which broke the Munki tools on 10.6 and 10.7 (at least).

keeleysam:
Installing snow leopard right now to verify.

keeleysam:
i have verified this change on stock 10.6.8 both standalone and with munkilib, operation does not change.

gregneagle:
Don't we need a `from __future__ import print_function` to use `print('foo')` as you do now?

gregneagle:
@keeleysam Can you either update to work with `makecatalogs` in Munki3dev, or abandon and close this?

PR #675, opened 2016-10-18
Only attempt to download icons with hashes

### LINKED ISSUE #670: Don't attempt to download icon where no icon hash present
Currently managedsoftwareupdate tries to download icons for all items, and caches them based on the icon hash in the pkginfo/catalog.

Since #444 makecatalogs generates these hashes automatically.

Munki shouldn't try to download icons when there is no hash, as if there is not one present and you're using a modern makecatalogs there almost certainly is not actually an icon to download and you generate a bunch of requests for items which always 404.

The question is, should this be a change in behavior or should it need yet another munki preference?

keeleysam:
Fixes #670

Currently munki attempts to download icons for every optional item.  If you're using `makecatalogs` from Munki 2.2 or later, icon hashes are added for every item where an icon is present.

This change has munki skip requests for packages where no icon_hash is present, saving a lot of requests and time particularly for those who have a lot of packages without icons.

### DIFF
--- a/code/client/munkilib/updatecheck.py
+++ b/code/client/munkilib/updatecheck.py
@@ -2886,7 +2886,8 @@ def download_icons(item_list):
                 munkicommon.display_error(
                     'Could not create %s' % icon_subdir)
                 return
-        if pkginfo_icon_hash != xattr_hash:
+        # only fetch icons that exist and don't match our cache.
+        if pkginfo_icon_hash != xattr_hash and pkginfo_icon_hash is not None:
             item_name = item.get('display_name') or item['name']
             message = 'Getting icon %s for %s...' % (icon_name, item_name)
             try:


erikng:
This wouldn't address the issue I presented. An admin can specify an icon in a pkginfo manually and a hash will not be generated.

gregneagle:
Sure it will -- when `makecatalogs` generates the catalogs.

erikng:
Ah I'm in idiot. It generates it in the catalog not the pkginfo.

gregneagle:
I like this idea, but there are some issues.

pkg icons do _not_ have to live in the icons directory next to the pkginfo directory. Clients can set IconURL and get the icons from _any_ URL, perhaps even on a different web server.

This change would prevent that -- since makecatalogs would not find the icon, it would not generate a hash. Right now that's not an issue, since clients will still request the icon even if the hash is not defined.

But with this change, the client will never request the icon, even though the admin has made it available via an alternate URL.

gregneagle:
The rationale behind this PR is addressed a different way here: https://github.com/munki/munki/commit/bccb7523ac109b2ff73ab4dc1ed9df1042a6b7ab.

PR #681, opened 2016-11-14
Version ordering option

franklefebvre:
Hi Greg,

Munki uses the LooseVersion class, which behaves in an unusual way when comparing versions for alpha and beta builds with the standard notation (1.0, 1.0a1, 1.0b1).
I added an option to managedsoftwareupdate, to change the way prereleases are compared to releases. When managedsoftwareupdate is invoked with the --semanticversionordering option, the releases above are ordered this way: 1.0a1, 1.0b1, 1.0.
This option is disabled by default, so that scripts relying on the current behaviour are not impacted.
Feel free to add this contribution to a future release of Munki if you think it can be useful.

Best regards,
Frank

### DIFF
--- a/code/client/managedsoftwareupdate
+++ b/code/client/managedsoftwareupdate
@@ -469,7 +469,7 @@ def main():
     p.add_option('--auto', '-a', action='store_true',
                  help='Used by launchd LaunchAgent for scheduled runs. '
                  'No user feedback or intervention. All other options '
-                 'ignored.')
+                 'ignored except --semanticversionordering.')
     p.add_option('--logoutinstall', '-l', action='store_true',
                  help='Used by launchd LaunchAgent when running at the '
                  'loginwindow.')
@@ -498,6 +498,9 @@ def main():
                  'packages.')
     p.add_option('--munkipkgsonly', action='store_true',
                  help='Only check/install Munki packages, skip Apple SUS.')
+    p.add_option('--semanticversionordering', action='store_true',
+                 help='Order prereleases before releases when sorting packages, '
+                 'e.g. 1.2b1 before 1.2.')
     p.add_option('--version', '-V', action='store_true',
                  help='Print the version of the munki tools and exit.')
 
@@ -612,6 +615,7 @@ def main():
     # set munkicommon globals
     munkicommon.munkistatusoutput = options.munkistatusoutput
     munkicommon.verbose = options.verbose
+    munkicommon.semantic_version_ordering = options.semanticversionordering
 
     # Set environment variable for verbosity
     os.environ['MUNKI_VERBOSITY_LEVEL'] = str(options.verbose)
--- a/code/client/munkilib/munkicommon.py
+++ b/code/client/munkilib/munkicommon.py
@@ -1321,13 +1321,23 @@ class MunkiLooseVersion(version.LooseVersion):
             cmp_list.append(0)
         return cmp_list
 
+    def _tag(self, version_list):
+        # attach a tag to each element so that strings are ordered before numbers
+        return [(0 if isinstance(e, StringType) else 1, e) for e in version_list]
+
+    def _normalize(self, version_list, max_length):
+        if semantic_version_ordering:
+            return self._tag(self._pad(version_list, max_length))
+        else:
+            return self._pad(version_list, max_length)
+
     def __cmp__(self, other):
         if isinstance(other, StringType):
             other = MunkiLooseVersion(other)
 
         max_length = max(len(self.version), len(other.version))
-        self_cmp_version = self._pad(self.version, max_length)
-        other_cmp_version = self._pad(other.version, max_length)
+        self_cmp_version = self._normalize(self.version, max_length)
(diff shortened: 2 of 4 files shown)


gregneagle:
Please see https://github.com/munki/munki/wiki/Contributing-to-Munki

I can no longer accept PRs from people for whom we do not have a signed CLA on file.

(Note that having a CLA on file is not a guarantee that PRs will be merged... )

gregneagle:
(But I think something along these lines could be very useful...)

timsutton:
I also would love to see something like this for certain applications that use alpha characters in version strings, but would imagine it to be more widely usable as a pkginfo key - this way you could tell Munki to use this alternate logic consistently but for a limited set of applications/installers.

franklefebvre:
Tim: I like your idea of being able to choose the sort ordering on a per-application basis, and I am willing to implement something like this.
However, unless I missed something, there is one pkginfo file per version. Therefore a pkginfo key may not be the best option for that setting, as it would lead to ambiguous behaviour when there are conflicting settings for a given application.
Any suggestions?

gregneagle:
Making this different version comparison active only when Munki is called with a different command-line option would make this feature unusable by most people since the launchd jobs that call the tools don't have that command-line option.

Making it a global preference might work, but could break the non-zero number of products that have letters in their version numbers that are not indicative of "development/alpha/beta" status.

Making it a pkginfo option seems the most reasonable. There really is no chance of "ambiguous behaviour" as Munki considers only one pkginfo item at any time.

(and I dislike the name "semanticversionordering". I'd prefer one that referred to "prerelease versioning" or something similar)

timsutton:
I interpret @franklefebvre to mean about "ambiguous behaviour" the question of how Munki should behave to order versions of multiple items of the same `name` by their `version` keys, if they don't all have this key set.

I think what @gregneagle is referring to is the logic by which Munki would compare something like the version of some `installs` item vs. a found version on disk, i.e. comparing `CFBundleShortVersionString` of  `1.0` in a pkginfo against a `1.0b` version already found on the machine.

If that's the case, these are two different phases of Munki's checking logic.

If this proposal means the former (and perhaps in addition to the latter), I think if Munki _were_ able to order its catalog items for a given `name` using `version` according to a different algorithm, then that "prerelease versioning" key should be present in every single item with that `name`. If it happens to not be, fall back to the default version comparison mechanism and log a warning that the key was added to at least one but not all eligible versions of that item.

gregneagle:
Ah, yes -- two different, but related, problems:

1) How does Munki order its internal list of versions for a product when the version numbers contain letters that have possible _meaning_, and:

2) How does Munki compare the version it is considering against the version found on the startup disk.

timsutton:
As an aside, I'm wondering whether this normalization of the version info would allow things like Camera Raw versions to be compared properly. Those are always of the form `9.1f300`, or `9.1.1f400`, and which LooseVersion "gets wrong" depending on how many numeric components there are before the alpha character.

Duncan McCracken talked about using `pkg_resources` in two of his "Which Version is This?" talks this year, and one thing he liked about it was that it seemed to pad out to the alpha characters properly:

```python
In [1]: from distutils.version import LooseVersion as LV

In [2]: LV('9.1f300') < LV('9.1.1f400')
Out[2]: False

In [3]: from pkg_resources import parse_version

In [4]: parse_version('9.1f300') < parse_version('9.1.1f400')
Out[4]: True
```

Looks like that also does "alpha" and "beta"-like identifiers:

```python
In [9]: parse_version('1.0b') < parse_version('1.0')
Out[9]: True

In [10]: parse_version('1.0b') < parse_version('1.0c')
Out[10]: True

In [11]: parse_version('1.0a') < parse_version('1.0b')
Out[11]: True
```

`pkg_resources` seems to be present on the Python 2.6 that ships with Snow Leopard in addition to the 2.7 for all versions since.

gregneagle:
"development" is lower than "beta", so

```
>>> parse_version('1.0b1') > parse_version('1.0d1')
True
```

seems to do the right thing. But is there an agreed-upon "meaning" for all the possible letters?

gregneagle:
FWIW, parse_version parses version strings against PEP 440 specifications: https://www.python.org/dev/peps/pep-0440/

franklefebvre:
Thanks Greg! I didn't know about parse_version. It seems to do exactly what's needed. And since setuptools is already included in OS X 10.6, there won't be any additional requirements for Munki 2.

So here is what I suggest:

- Add a "version_scheme" key to the pkginfo dict, with a string value in order to allow for future extension. Currently supported values will be "default" and "parse_version". (you may come up with better names, English is not my first language.)

- When looking for a new version for a given package, Munki will extract the version_scheme information for every version of that package in the pkginfo file.
    - If all version_scheme values are identical and correspond to a supported scheme, Munki will use this scheme for version comparison.
    - Otherwise, it will log a warning and revert to the default behavior.

If you agree to this proposal, I can implement it soon.

gregneagle:
Closing this old PR with conflicts and no current proponent.

PR #684, opened 2016-12-16
Support for pkginfo key "script_environment"

choules:
We have used munki a lot for configuration management during the last months. This often involved writing a lot of scripts: installcheck_scripts for checking on current settings, postinstall_scripts for setting the new config and uninstall_scripts for resetting default values.

In each of these scripts most of the time the same variables where needed, for example the key of a plist dictionary and the corresponding value. This resulted in heavy copying and pasting and those usual errors like forgetting to paste a value.

With support for a "script_environment" key in pkginfo files, we would be able to set some basic variables for use in all embedded scripts of a pkginfo.

I have duplicated the behavior of "installer_environment", just to be consistent. And of course, especially for OnDemand items it could be helpful to switch to the context of "CURRENT_CONSOLE_USER".

I have tested this build inside a virtual machine and afterwards on some of my production clients, without any issues.


gregneagle:
Why add a new dict and more code to makepkginfo? Why not just use the installer_environment?

choules:
I thought of "installer_environment" to be specifically for Apple's "installer" and did not want to mix this with any extension for embedded scripts.

One might as well want to use different environment variables for the installer and for the embedded scripts.

And, especially with "makepkginfo" the option parser suggested a clear line between installer and script functionality.

But, of course, we might as well use the same pkginfo item for both cases. But then I would suggest to rename the key to "environment", which might break backward compatibility... :-/

gregneagle:
There's precedent for renaming keys: unattended_install and unattended_uninstall were originally force_install and force_uninstall; Munki supported both names for a long time (and may still).

I'm not convinced install_environment would be confusing for the one or two people on the planet who would be using it...

choules:
Alright, I will update the branch by renaming the install_environment key to "environment". I will try to keep a backward compatibility for the time being.

erikng:
If you are reusing code, couldn't you also install modules on the machines and call them? You can find many examples of this in other munki tools (Sal, MunkiReport) and design your code around this.

It's completely extensible to you and requires zero code changes to munki.

On Dec 16, 2016, at 11:25 AM, Matthias Choules <notifications@github.com<mailto:notifications@github.com>> wrote:


Alright, I will update the branch by renaming the install_environment key to "environment". I will try to keep a backward compatibility for the time being.

-

choules:
@gregneagle I have renamed the "install_environment" key to "environment" and used this for scripts as well. Tests went well.

@erikng The code I was reusing was different with each pkginfo. Writing a module for each item would be kind of overkill. Or did I get you wrong?

erikng:
I'm not sure the kinds of things you are writing across software layers, but you could write a function that takes inputs.

gregneagle:
Closing this old PR that has conflicts and no current proponents.

PR #685, opened 2016-12-20
Enable plugins to munki that will allow writes to the munki repo to be customized.

ryanyu91:
**Changed Files:**
code/client/munkiimport
code/client/iconimporter
code/client/makecatalogs
code/client/manifestutil
code/client/munkilib/iconutils.py

**Added Files:**
code/client/munkilib/FileRepo.py
code/client/munkilib/Repo.py

**Reason For Changes:**
 
The purpose for this change is to enable plugins to munki that will allow writes to the munki repo to be customized.
 
**Changes:**
 
The methods used to write to the munki repo are the target of this modification. A plugin can create new methods for overwriting the default behavior of writing to the munki repo.
 
The default behavior is retained in the absence of a plugin. This is accomplished via the introduction of FileRepo.py which continues to simply write changes to the munki repo. A plugin can be introduced to change this default behavior and allow munki repo writes to be redirected as desired by the plugin author.
 
In order to accomplish this we have refactored all of the os.path.* methods, as well as the mount, unmount and available methods from the following tools:  (iconimporter, makecatalogs, munkiimport, manifestutil, iconutils), and put them into a different python script to be used as a library or common code module. By default, these are now in the FileRepo.py module. The plugin can be used to overwrite this common code module. 
 
In addition to the above described changes, we also made the following changes to support this plugin
concept:
           •	Added the ability to add a plugin either via the command line or via munkiimport –configure
           •	The code will look for custom plugins in the /usr/local/munki/munkilib/plugins/ directory
           •	If plugin is found in the plugins directory, munkiimport --configure will give the option to type                   in the plugin name. For example, if you plugin is Foo.py, you could specify a plugin name of Foo in munkiimport --configure 
           •	If no plugin is found the FileRepo.py module will be used as the default common code module for writing to the munki repo
           •	Add the ability to set the plugin via the --plugin option for the following tools (makecatalogs, munkiimport, manifestutil)

**Testing**
Tested with munkiimport on local filesystem/network shares/our own custom plugins
Tested manifestutil on local filesystem/network
Verified through regression testing, all features of munkiimport, makecatalogs, iconimporter, manifestutil from before all still work (local filesystem/network shares)

### DIFF
--- a/code/client/iconimporter
+++ b/code/client/iconimporter
@@ -29,6 +29,7 @@ from optparse import OptionParser
 from munkilib import munkicommon
 from munkilib import FoundationPlist
 from munkilib import iconutils
+from munkilib import Repo
 
 # PyLint cannot properly find names inside Cocoa libraries, so issues bogus
 # No name 'Foo' in module 'Bar' warnings. Disable them.
@@ -36,11 +37,11 @@ from munkilib import iconutils
 from Foundation import CFPreferencesCopyAppValue
 # pylint: enable=E0611
 
-def generate_png_from_copy_from_dmg_item(install_item, repo_path):
+def generate_png_from_copy_from_dmg_item(install_item, repo):
     '''Generate a PNG from a disk image containing an application'''
-    dmgpath = os.path.join(
-        repo_path, 'pkgs', install_item['installer_item_location'])
-    mountpoints = munkicommon.mountdmg(dmgpath)
+    dmgpath = repo.join('pkgs', install_item['installer_item_location'])
+    handle = repo.open(dmgpath, 'r')
+    mountpoints = munkicommon.mountdmg(handle.local_path)
     if mountpoints:
         mountpoint = mountpoints[0]
         apps = [item for item in install_item.get('items_to_copy', [])
@@ -49,9 +50,9 @@ def generate_png_from_copy_from_dmg_item(install_item, repo_path):
             app_path = os.path.join(mountpoint, apps[0]['source_item'])
             icon_path = iconutils.findIconForApp(app_path)
             if icon_path:
-                png_path = os.path.join(
-                    repo_path, u'icons', install_item['name'] + u'.png')
-                result = iconutils.convertIconToPNG(icon_path, png_path)
+                png_path = repo.join(u'icons', install_item['name'] + u'.png')
+                png_handle = repo.open(png_path, 'w')
+                result = iconutils.convertIconToPNG(icon_path, png_handle.local_path)
                 if result:
                     print_utf8(u'\tWrote: %s' % png_path)
                 else:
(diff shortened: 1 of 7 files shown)


code/client/manifestutil
```diff
@@ -417,77 +421,14 @@ def manifest_rename(source_manifest_name, dest_manifest_name,
...
-        return
-    cmd = ['/sbin/umount', repo_path]
-    return subprocess.call(cmd)
-
-
 def cleanup_and_exit(exitcode):
     """Give the user the chance to unmount the repo when we exit"""
     result = 0
-    if WE_MOUNTED_THE_REPO:
+    if repo.mounted:
```
gregneagle: Careful that we differentiate between "is the repo mounted" and "did _we_ mount the repo". If the repo was already mounted before we run, we should not bother the user and ask if the repo should be unmounted.

code/client/munkilib/FileRepo.py
```diff
@@ -0,0 +1,171 @@
...
+"""
+FileRepo
+Created by Centrify Corporation 2016-06-02.
+Implementation for accessing a repo via direct file access, including
+a remote repo mounted via AFP, SMB, or NFS.
+"""
+
+from collections import namedtuple
+import sys
+sys.path.append("/usr/local/munki/munkilib")
```
gregneagle: None of the other Munki Python code is tied to a specific install location -- it all works with relative imports. I'd like that to continue to be the case. This also helps when testing -- I can leave the "production" versions of the Munki tools installed in /usr/local/munki, but test new code anywhere in working space.

code/client/iconimporter
```diff
@@ -29,18 +29,19 @@ from optparse import OptionParser
...
 
-def generate_png_from_copy_from_dmg_item(install_item, repo_path):
+def generate_png_from_copy_from_dmg_item(install_item, repo):
     '''Generate a PNG from a disk image containing an application'''
-    dmgpath = os.path.join(
-        repo_path, 'pkgs', install_item['installer_item_location'])
-    mountpoints = munkicommon.mountdmg(dmgpath)
+    dmgpath = repo.join('pkgs', install_item['installer_item_location'])
+    handle = repo.open(dmgpath, 'r')
+    mountpoints = munkicommon.mountdmg(handle.local_path)
```
gregneagle: Does this actually work? IOW, can you open a disk image for reading and then mount it? I'd think mountdmg would be unhappy with a file that was already open...

code/client/munkilib/FileRepo.py
```diff
@@ -0,0 +1,171 @@
...
+                self.repo_mode = mode
+                self.file = open(self.repo_path, mode)
+                self.local_path = self.repo_path
+
+            def read(self):
+                return self.file.read()
+
+        return RepoFile(self, os.path.join(self.path, path), mode)
+
+    def mount(self):
```
gregneagle: This function has completely removed the functionality added to allow fileshares to be mounted under /Volumes in 10.12. This was added here: https://github.com/munki/munki/commit/e9bdb65ade01294aa702ebbf40a3e050bd0a244b and here: https://github.com/munki/munki/commit/70135813715cc3bd262fd036352fb5c680720c0c

code/client/munkilib/Repo.py
```diff
@@ -0,0 +1,61 @@
...
+def Open(path, url, plugin):
+    #looks for plugin in /usr/local/munki/munkilib/plugins (installation of munki)
+    if plugin == None or plugin == "":
+        #default is FileRepo if no plugin is specified in configuration or options.
+        module = imp.load_source('FileRepo', '/usr/local/munki/munkilib/FileRepo.py')
+        import_class = getattr(module, "FileRepo")
+        parent = import_class
+    else:
+        module = imp.load_source(plugin, '/usr/local/munki/munkilib/plugins/' + plugin + ".py")
+        import_class = getattr(module, plugin)
```
gregneagle: Again I dislike all the hard-coding of install locations here -- let's get this working with relative locations. Perhaps also the FileRepo module should be in the plugins directory like any other repo plugin

ryanyu91:
Made changes in new commit: https://github.com/ryanyu91/munki/commit/cfca75aca75fc73c70b16d760395971a06eaa045

Commit based on comments by Greg Neagle on what to fix: munki#685 (https://github.com/munki/munki/compare/master...ryanyu91:master)
IconImporter:
	- Not opening DMG and then mounting it anymore, directly mounting like how it was before
ManifestUtil:
	- Checking if the repo is mounted as well as if we (munki) mounted it. Only this will display prompt whether we want to unmount or not
FileRepo.py, Repo.py:
	- not hardcoding import path anymore
	- FileRepo - added 10.12 mounting fileshares code

Issue #687, opened 2017-01-06
Official Support for loading launchd's without reboot


erikng:
Given the likelihood of imaging being deprecated in leu of DEP provisioning, I am wondering if we should have official support for the launchdaemons/agents to be loaded during installation.

Pros: 
- DEP bootstrap with munki can immediately function without additional packages/postinstall scripts that are custom to every company.
- Other traditional workflows can reduce a reboot.
- Reboot no longer needed through GUI installs
- Less need for autopkg recipes to deploy each sub munki pkg as the entire package can be deployed without issue. 

Cons:
- Will require a new launchd package
- Will more than likely require pre/post install for launchctl scripts and account for potential race conditions.
- This could be a breaking change for people who touch munki's trigger file during provisioning and expect a restart.
- Further documentation

gregneagle:
This is far trickier than you might think: a postinstall script that unloaded/loaded the Munki launchdaemons would kill a Munki session that was trying to install an updated set of Munki launchdaemons...

Feel free to think about it and experiment, but it's incredibly low on my list of priorities.

nmcspadden:
At an initial first pass, I'd be happy if this was limited just to "first time installs" - i.e. no Munki launchds are already loaded (and therefore wouldn't have any potential problems killing existing Munki processes by loading them).

erikng:
Perhaps we only handle it for the first installation then and leave it alone.

gregneagle:
Also, you note "Reboot no longer needed through GUI installs" -- rebooting also makes sure the updated launch _agents_ get loaded; without a reboot (or logout of all sessions) you'll need to handle the unload/load of the launch agents in all GUI sessions, again being careful to not kill anything in use by Munki...

gregneagle:
"At an initial first pass, I'd be happy if this was limited just to "first time installs" --

We can detect if there are any Munki launch daemons running or loaded, but if there are, then we need to restart instead. You can't decide you need to restart in a package postinstall script, you have to decide _before_ the install starts -- indeed, even before a preinstall script runs. It would have to be done in a distribution script. Not impossible (I think), but not trivial...

gregneagle:
FWIW, I have several non-imaging workflows that install the Munki tools as part of building or deploying a machine. If we get limited to only using MDM+DEP we have a lot more problems to deal with, one major one being the likelihood that Munki's primary developer stops working on Munki :-)

gregneagle:
[...]

PR #690, opened 2017-01-10
make DEP munki mpkg

erikng:
This creates a new munki mpkg for use with DEP.

Fixes https://github.com/munki/munki/issues/687.

This package must be run on 10.7 and higher.


erikng:
In hindsight, I should have added an initial commit with the mpkg script so we could diff the changes I made from the original. I am going to rebase and issue a new PR.

PR #693, opened 2017-01-17
Consistently use /usr/bin/python shebang.

ab:
A `#!/usr/bin/env python` can cause problems where scripts are unable to
load munkilib from the python path. Whatever we use, we should use
consistently across all scripts.

For example, a homebrew python at `/usr/local/bin/python` can cause
messages like this:

    The following errors occurred while building catalogs:

    WARNING: FoundationPlist is not available, using plistlib instead.

### DIFF
--- a/code/apps/Managed Software Center/msu_testing/generate_msu_test_data.py
+++ b/code/apps/Managed Software Center/msu_testing/generate_msu_test_data.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python
 # encoding: utf-8
 """
 generate_msu_test_data.py
--- a/code/client/makecatalogs
+++ b/code/client/makecatalogs
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python
 # encoding: utf-8
 #
 # Copyright 2009-2016 Greg Neagle.
--- a/code/client/repoclean
+++ b/code/client/repoclean
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python
 # encoding: utf-8
 #
 # Copyright 2016 Greg Neagle.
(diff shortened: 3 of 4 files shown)


gregneagle:
The use of `#!/usr/bin/env python` for these file was not arbitrary.

As long as munkilib is in the same directory as makecatalogs and repoclean, the Python interpreter path will make no difference in whether or not those tools can import munkilib.

makecatalogs is built so it relies on no OS X-specific modules and can run, for example, on Linux or Windows (so it can run on the box hosting the Munki repo), and should not rely on a specific /usr/bin/python install. The intention is the same for repoclean -- it should be able to be run on the box hosting the Munki repo.

getmanifest.py is an example of a web CGI and should use whatever Python is available on the webserver, which again could be a Linux or Windows box.

I'm inclined to leave these files as-is.

ab:
Right, sorry, the specific issue is not being able to load `foundation` in the homebrew-installed `/usr/local/bin/python` due to not having any `pyobjc` package installed.

If this behavior is intentional, I'd suggest removing the `WARNING: FoundationPlist is not available, using plistlib instead.` Or, if some action is recommended, then the warning should be made actionable.

natewalck:
@ab In that case, since you have brew python higher in your path, I'd call makecatalogs using the system python rather than depending upon your PATH.

ab:
The real issue here is that warning message:
```
WARNING: FoundationPlist is not available, using plistlib instead.
```

Based on your replies it seems that this is not actually problematic. But the user has no way of knowing what the impact is or what to do about it other than source diving.

gregneagle:
...or asking in one of the many suoport venues, as you've done.

gregneagle:
See https://github.com/munki/munki/commit/4b4bbb8aaced9b6f8c8f8369a8f01134ec30011c

gregneagle:
I would take a PR that caused the note/warning not to be printed at all if `makecatalogs` is running on a platform other than macOS/OS X.  On macOS/OS X the user should at least be aware that there is a difference from "normal" operation here.

PR #699, opened 2017-02-09
Add serial number to README.md

bochoven:


gregneagle:
I forgot that directory and files even existed. I wonder if it might be better to just delete the entire folder to eliminate duplicate work between the wiki and the README in the Preferences folder. (This folder pre-dates the wiki and was intended as a bit of "getting started" information.

bochoven:
Yes, delete it if you want. I was just making sure there aren't different 'truths' about Munki's behaviour.

bochoven:
I think the low barrier of updating the wiki weighs up against the practice of having documentation following releases.

gregneagle:
Removed Preferences directory here: https://github.com/munki/munki/commit/81160a1741efaa4ed16f124860cfb32e6ae8a00a

PR #703, opened 2017-02-18
Add -s option to sign package

erikng:
This PR adds -s option to sign both the DEP package script and the normal script.

`./make_munki_mpkg.sh -s "Developer ID Installer: Munki (U8PN57A5N2)"`
`./make_munki_mpkg_DEP.sh -s "Developer ID Installer: Munki (U8PN57A5N2)"`

I have tested without the option and with the option for both scripts. This should not be a breaking change to anyone's configuration.

### DIFF
--- a/code/tools/make_munki_mpkg.sh
+++ b/code/tools/make_munki_mpkg.sh
@@ -32,18 +32,19 @@ fi
 
 usage() {
     cat <<EOF
-Usage: `basename $0` [-i id] [-r root] [-o dir] [-c package]"
+Usage: `basename $0` [-i id] [-r root] [-o dir] [-c package] [-s cert]"
 
     -i id       Set the base package bundle ID
     -r root     Set the munki source root
     -o dir      Set the output directory
     -c package  Include a configuration package (NOT CURRENTLY IMPLEMENTED)
+    -s cert     Sign distribution package with signing certificate from keychain
 
 EOF
 }
 
 
-while getopts "i:r:o:c:h" option
+while getopts "i:r:o:c:s:h" option
 do
     case $option in
         "i")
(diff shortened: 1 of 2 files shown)


code/tools/make_munki_mpkg.sh
```diff
@@ -32,18 +32,19 @@ fi
...
 usage() {
     cat <<EOF
-Usage: `basename $0` [-i id] [-r root] [-o dir] [-c package]"
+Usage: `basename $0` [-i id] [-r root] [-o dir] [-c package] [-s cert]"
 
     -i id       Set the base package bundle ID
     -r root     Set the munki source root
     -o dir      Set the output directory
     -c package  Include a configuration package (NOT CURRENTLY IMPLEMENTED)
+    -s cert     Sign distribution package with signing certificate from keychain
```
gregneagle: Might want to better describe the needed input string; this is going to confuse people

erikng: How about now?

gregneagle: Still not sure about that. 'cert' and 'signing certificate from keychain' still seem ambiguous. What is required here is the _common name/CN_ of the 'signing certificate from keychain', no?

erikng: Correct.

How about this:
`-s CN    Sign distribution package with an installer certificate from keychain. Requires Common Name`

gregneagle: or `-s cert_cn      Sign distribution package with a Developer ID Installer certificate from keychain. Provide the certificate's Common Name`

erikng:
Thanks for merging. This will be extremely helpful for people who want to deploy a standard munki install through DEP (and of course for everyone else as well).

PR #707, opened 2017-02-24
Create power assertion for downloads

carlashley:
In pre-deployment testing of munki, I've noticed that "fresh" installs of macOS will go to sleep while downloading packages. It appears power assertions are created for a package installation, but aren't created for downloading packages.

This is an issue for fresh installs of macOS, especially where you are doing a 'first run' to get all software deployed to a machine. I've also seen this issue occur on laptops with aggressive energy saver settings.

Adding a power assertion during the download process fixed this issue in testing.

### DIFF
--- a/code/client/munkilib/updatecheck/download.py
+++ b/code/client/munkilib/updatecheck/download.py
@@ -31,6 +31,7 @@ from .. import fetch
 from .. import info
 from .. import munkihash
 from .. import osutils
+from .. import powermgr
 from .. import prefs
 from .. import reports
 
@@ -104,6 +105,8 @@ def download_installeritem(item_pl, installinfo, uninstalling=False):
     Returns True if the item was downloaded, False if it was already cached.
     Raises an error if there are issues..."""
 
+    # Create a power assertion to stop sleep during downloads
+    no_idle_sleep_assertion_id = powermgr.assertNoIdleSleep()
     download_item_key = 'installer_item_location'
     item_hash_key = 'installer_item_hash'
     if uninstalling and 'uninstaller_item_location' in item_pl:
@@ -151,6 +154,9 @@ def download_installeritem(item_pl, installinfo, uninstalling=False):
 
     dl_message = 'Downloading %s...' % pkgname
     expected_hash = item_pl.get(item_hash_key, None)
+
+    # Release power assertion
+    powermgr.removeNoIdleSleepAssertion(no_idle_sleep_assertion_id)
     return fetch.munki_resource(pkgurl, destinationpath,
                                 resume=True,
                                 message=dl_message,


MichalMMac:
I belive there was thread about this in one of Munki google group. Not sure outcome was in favor of implementing this change.

carlashley:
Yes, I've seen that discussion...

In my environment, I've had a number of users with munki on their own personal laptops with this modification, they're prompted to connect to AC; honestly, I think that's enough of a warning.

In my deployment scenarios, regardless of whether the device is a laptop or a desktop, the first run of munki is always done on AC.

It's a complete pain in the ass to have a lab full of machines stall during the first run because they went to sleep while downloading a package.

If you're using the `/Users/Shared/.com.googlecode.munki.checkandinstallatstartup` file to trigger your first run installation method (which I do, because I'm completely over building first run Launch Daemons and scripts, and don't care to use them anymore), there is no other way to stop the machine going to sleep.

gregneagle:
I'm not inclined to accept this as-is: I think during "normal" operation Munki should _not_ be creating power assertions during downloads. If you were to only create these assertions during a bootstrapping run, I'd be more likely to accept it.

gregneagle:
See also https://groups.google.com/d/topic/munki-dev/esrOL3zAzOU/discussion

gregneagle:
...which also references the discussion here: #602

gregneagle:
After reviewing the previous discussion, my mind hasn't changed here. I'd be more likely to accept a PR that did one of the following:

Did no-sleep assertions for downloads only when on AC power (not on battery)
Did no-sleep assertions for downloads only during "bootstrapping" and not during "normal" operation.

It seems either of these would address your issue without adversely affecting battery in "normal" operation.

flammable:
I like this idea!  I agree with Greg, though - if it could be limited to bootstrapping (or AC power), I think that would avoid some unexpected behavior.

gregneagle:
So does this mean you have no interest in modifying your PR?

carlashley:
I'm not in a position at present to do further testing on AC/battery power, and it occurs to me that it's probably best to do that sort of check within the power assertion module itself, so that the power assertion can be called anywhere within munki and have that logic built in.

As I mentioned in another reply, I've seen laptops with quite aggressive power savings with this issue as well, and even though munki attempts to resume downloading the file once the laptop is powered back up, it has failed more often than succeeded. Said laptops are personal laptops, and so I try to avoid muddling with peoples configuration choices.

Per some of the other conversations when this has popped up before, I guess I'm in the minority in that I'd rather see the package download properly first time every time, than deal with help desk calls where users want to know why `xyz.pkg` has failed to install for the fifth time that day, when it just boils down to keeping the machine awake for the 5-10min it will take to grab the package; which at least for my use cases is the likely scenario, given the bulk of software installs are done in a 'first flight' manner.

Perhaps creating a power assertion when the battery level is greater than n% might be viable, and perhaps even allow it to be customised per client within the `ManagedInstalls.plist` prefs?

PR #709, opened 2017-02-28
importing "os" so the launchctld function will actually work

weswhet:
added `import os`..

post install appears to work for DEP after this :)


gregneagle:
LGTM. Hope @erik is fine with it...

erikng:
Sorry about that. I think during a commit I removed the import os thinking it was no longer needed for another line I removed.

PR #717, opened 2017-03-08
Updated Swedish localization.

MagerValp:
Added Swedish localization for featured items and the system configuration problem alert, plus a few minor tweaks.


gregneagle:
This really should not be merged to master, but rather to the Munki3 branch. I'll try to figure that out!

MagerValp:
Oh dang it. I'll make a new PR.

Issue #428, opened 2014-12-05
munki 2 update button disappear in french with package "force install after date"  and "logout"


sebizare:
Hi Greg,
Thanks again for the great apps munki 2
It seems that the button "update" in munki 2 disappear in french version when there is a package with "force install after date"  and "logout". I test it with the last install (3 december  2014)
see the screenshot.
If the user change the windows size the button reappear
In english no problem. but it seems that's the length of the text in french

![en1](https://cloud.githubusercontent.com/assets/10085519/5313886/7fa2fc3e-7c6d-11e4-96ae-f30ffc01599c.png)
in french the button 
![fr_withforce_installdate_and_logout](https://cloud.githubusercontent.com/assets/10085519/5313890/895cd7b8-7c6d-11e4-8ee8-256fe4bca75e.png)
do you thinks it's possible to make the button to have the priority on the text ?

cordialy sebastien
https://www.youtube.com/watch?v=wCkerYMffMo

gregneagle:
Closing...

PR #718, opened 2017-03-08
Fix 'one or more' in fr strings for MSC, notifier

timsutton:
Re: https://groups.google.com/d/msg/munki-dev/7YvwIZbNGxI/xosTgMr4DgAJ


RichardLafontaine:
Hello Tim, I initially submitted that string in 2014 when I translated MSC, but it caused the issue #428. It was too long and the Update button didn't show because of that. It was fixed in #430. Sorry I was late to reply in the forum.

gregneagle:
Should this be closed in favor of https://github.com/munki/munki/pull/723 ?

timsutton:
Derp, guess so. I didn't realize that was the reason behind it.

PR #689, opened 2017-01-06
Enable plugins to munki that will allow writes to the munki repo to be customized.

ryanyu91:
NEW CHANGES

Commit based on comments by Greg Neagle on what to fix: munki#685 (master...ryanyu91:master)
IconImporter:

Not opening DMG and then mounting it anymore, directly mounting like how it was before
ManifestUtil:
Checking if the repo is mounted as well as if we (munki) mounted it. Only this will display prompt whether we want to unmount or not
FileRepo.py, Repo.py:
not hardcoding import path anymore
FileRepo - added 10.12 mounting fileshares code
Changed Files:
code/client/munkiimport
code/client/iconimporter
code/client/makecatalogs
code/client/manifestutil
code/client/munkilib/iconutils.py

Added Files:
code/client/munkilib/FileRepo.py
code/client/munkilib/Repo.py

Reason For Changes:

The purpose for this change is to enable plugins to munki that will allow writes to the munki repo to be customized.

Changes:

The methods used to write to the munki repo are the target of this modification. A plugin can create new methods for overwriting the default behavior of writing to the munki repo.

The default behavior is retained in the absence of a plugin. This is accomplished via the introduction of FileRepo.py which continues to simply write changes to the munki repo. A plugin can be introduced to change this default behavior and allow munki repo writes to be redirected as desired by the plugin author.

In order to accomplish this we have refactored all of the os.path.* methods, as well as the mount, unmount and available methods from the following tools: (iconimporter, makecatalogs, munkiimport, manifestutil, iconutils), and put them into a different python script to be used as a library or common code module. By default, these are now in the FileRepo.py module. The plugin can be used to overwrite this common code module.

In addition to the above described changes, we also made the following changes to support this plugin
concept:
•	Added the ability to add a plugin either via the command line or via munkiimport –configure
•	The code will look for custom plugins in the /usr/local/munki/munkilib/plugins/ directory
•	If plugin is found in the plugins directory, munkiimport --configure will give the option to type in the plugin name. For example, if you plugin is Foo.py, you could specify a plugin name of Foo in munkiimport --configure
•	If no plugin is found the FileRepo.py module will be used as the default common code module for writing to the munki repo
•	Add the ability to set the plugin via the --plugin option for the following tools (makecatalogs, munkiimport, manifestutil)

Testing
Tested with munkiimport on local filesystem/network shares/our own custom plugins
Tested manifestutil on local filesystem/network
[...]

PR #721, opened 2017-03-08
Plugins can Inherit FileRepo

ryanyu91:
Added __init__.py to denote that plugins directory contains packages and able to import modules from plugins directory.

Added a condition whether to show plugin option during munkiimport configure to also ignore __init__

Based on Pull Request https://github.com/munki/munki/pull/689#pullrequestreview-16492427

### DIFF
--- a/code/client/munkiimport
+++ b/code/client/munkiimport
@@ -820,7 +820,7 @@ def configure():
             plugins = [fn for fn in os.listdir(plugin_path)
               if any(fn.endswith(ext) for ext in included_extensions)]
             for plugin in plugins:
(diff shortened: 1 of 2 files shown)


gregneagle:
@ryanyu91 : https://github.com/munki/munki/blob/new_repo_abstraction/notes_about_this_branch.md

Ryan: please take a look and see if Centrify can work with this style of repo plugin. We can add features if there are things missing, but it seems a more flexible starting point.

gregneagle:
I'm closing this PR since the current implementation is quite different.

PR #724, opened 2017-03-10
Allow marking optional installs as featured in manifests.

MagerValp:


gregneagle:
I've merged this, but I'm going to change the section/key name to "featured_items".

gregneagle:
Renamed and documented: https://github.com/munki/munki/wiki/Featured-Items-in-Managed-Software-Center

PR #727, opened 2017-03-13
standardize format of all launchagents and launchdaemons

erikng:
This change simply lints all the launchagents and daemons.

I figure since we the launchd package is already updating, this is a perfect time.

By doing this, an admin can install the normal launchd's with a standard munki package, and then manage those same launchd's with a CM tool (say) chef without the CM tool thinking something is wrong with the files and attempting to unload/load them.

### DIFF
--- a/launchd/LaunchAgents/com.googlecode.munki.ManagedSoftwareCenter.plist
+++ b/launchd/LaunchAgents/com.googlecode.munki.ManagedSoftwareCenter.plist
@@ -2,27 +2,27 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
-  <key>Label</key>
-  <string>com.googlecode.munki.ManagedSoftwareCenter</string>
-  <key>LimitLoadToSessionType</key>
-  <array>
-    <string>Aqua</string>
-  </array>
-  <key>ProgramArguments</key>
-  <array>
-    <string>/usr/local/munki/launchapp</string>
-    <string>-a</string>
-    <string>/Applications/Managed Software Center.app</string>
-  </array>
-  <key>RunAtLoad</key>
-  <false/>
-  <key>KeepAlive</key>
-  <dict>
-     <key>PathState</key>
-     <dict>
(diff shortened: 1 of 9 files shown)


gregneagle:
I don't think what you mean by "lint" is what happens when you, say, run `plutil -lint foo.plist`

Can you be clearer about what you actually did so the process could be repeated in the future?

I get that you reformatted the plist files to some undefined standard. I'm wondering what that standard _is_, or what tool you used.

(Because even before you reformatted them, they were valid plists that launchd happily accepted. If this change is being preposed to make a specific tool happy, please document which tool that is and what specifically was done to make that tool happy.)

erikng:
I used the standard plist library from python to make these new files.
```python
old = plistlib.readPlist('/munki/launchd/LaunchDaemons/com.googlecode.munki.managedsoftwareupdate-manualcheck.plist')

plistlib.writePlist(old, '/munki/launchd/LaunchDaemons/com.googlecode.munki.managedsoftwareupdate-manualcheck.plist')
```

gregneagle:
Ok. Is there an example of any sort of _improvement_ this change makes? You seemed to imply that there might be some sort of issue with, say, Chef ("manage those same launchd's with a CM tool (say) chef without the CM tool thinking something is wrong with the files and attempting to unload/load them.")

erikng:
If you define a launchd resource as a hash, chef will interpret the hash in the same manner as the changes I made. The keys are sorted. Since the keys in the original files are not sorted, it interprets that the file is not configured correctly and will unload it, change the file contents and load it.

This is problematic as it could potentially kill munki. I have internally worked around this issue (and manage the launchd's with different filenames anyway) but figured this would potentially help others if they venture down this path.

Again, given that we are making changes to the package and require a reboot anyway, I figured it would make sense to sort the keys. Ruby and Python both want to sort the keys by default, so why should we argue with these systems?

nmcspadden:
+1 to having the launch daemons/agents act more like generated code. Since there's no expectation anyone will ever hand-edit them, it would be an overall config-management-friendliness improvement with no impact to existing users if they come, by default, in a 'sorted' form.

I agree with Erik only because there's already going to be a change to the launchds. This is a good time to piggyback on that change.

erikng:
Thanks for merging!

PR #728, opened 2017-03-13
Remove RunAtLoad from LaunchAgent plists

erikng:
Apologies for missing this, but I forgot one other change.

According to launchd's documentation, RunAtLoad is not needed. I ran into this before on another tool:

```     RunAtLoad <boolean>
     This optional key is used to control whether your job is launched once at the time the job is loaded.
     The default is false.
```

https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man5/launchd.plist.5.html

This key is not needed as by default it will be false.

### DIFF
--- a/launchd/LaunchAgents/com.googlecode.munki.ManagedSoftwareCenter.plist
+++ b/launchd/LaunchAgents/com.googlecode.munki.ManagedSoftwareCenter.plist
@@ -22,7 +22,5 @@
 		<string>-a</string>
 		<string>/Applications/Managed Software Center.app</string>
 	</array>
-	<key>RunAtLoad</key>
-	<false/>
 </dict>
 </plist>
--- a/launchd/LaunchAgents/com.googlecode.munki.MunkiStatus.plist
+++ b/launchd/LaunchAgents/com.googlecode.munki.MunkiStatus.plist
@@ -26,7 +26,5 @@
 	<array>
 		<string>/Applications/Managed Software Center.app/Contents/Resources/MunkiStatus.app/Contents/MacOS/MunkiStatus</string>
 	</array>
-	<key>RunAtLoad</key>
-	<false/>
 </dict>
 </plist>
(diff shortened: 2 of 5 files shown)


gregneagle:
Again: this is to make your config management tool happy?

gregneagle:
And: has it always been false by default? 10.5? 10.6?  There are certainly many historical artifacts.

erikng:
It has always been false. You can merge or not, but I'd love to see it come in. I should have remembered that and didn't.

gregneagle:
Why did you pick this particular LaunchAgent? They all have RunAtLoad=false.

I'm inclined to keep it as it makes it clearer to humans who read these what is going on. You didn't clarify if this causes issues with Chef.

erikng:
I have too many forks right now. You are right, they need to be removed as well.

I have removed them from the rest of the files and yes, it does impact chef.

gregneagle:
So:

1) Let do them all or not at all.
2) In the future, be sure to include a rationale for changes like this so I don't think it's just someone's OCD. Changes often have fallout; sometimes we forget to test the impact of changes. Understanding the rationale helps weigh the risk vs benefit of any particular change. (IOW, if nothing is broken...)

erikng:
1. My vote is to remove them as it's not necessary.

2. Fair enough. It's definitely not OCD. I'm trying to prevent users who get into config management get trumped up by the CM tools doing things to these plists.

I've been wanting to send a PR for this for about six months but knew it wasn't justified. I figure if we don't fix this now, we never will be able to.

gregneagle:
I do wonder that if Chef is so sensitive in this regard -- how on earth do you deal with third-party software that doesn't adhere to the Chef's narrow interpretation of a "good" launchd plist?

erikng:
One could _not_ manage third party software and chef would leave it alone, but yes, it does get a little interesting once you opt to manage them.

I think for a tool this important, if you care about managing munki, you care about the daemons and agents as well.

PR #733, opened 2017-03-17
Don't show status window until Setup Assistant has finished

MagerValp:
This resolves the issue with Munki trying to run before Setup Assistant has finished (if it hasn't been bypassed with .AppleSetupDone), but getting stuck waiting for the network, blocking the user from using the assistant to configure networking.

### DIFF
--- a/code/apps/MunkiStatus/MunkiStatus/MSUStatusWindowController.py
+++ b/code/apps/MunkiStatus/MunkiStatus/MSUStatusWindowController.py
@@ -174,6 +174,11 @@ class MSUStatusWindowController(NSObject):
 
     def initStatusSession(self):
         '''Initialize our status session'''
+        # Wait for Setup Assistant to finish before showing the window.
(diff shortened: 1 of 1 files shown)


gregneagle:
As discussed in Slack, this has the potential for machine-bricking badness:

By suppressing MunkiStatus, user can interact with the Setup Assistant, including creating a new account, and then when Setup Assistant completes, the user is auto-logged into the newly-created account.

This (we believe) kills any loginwindow LaunchAgents, including the one running Munki. If Munki is still bootstrapping, and especially if it is doing something like installing a macOS update when it is killed, that could lead to corrupt software included a corrupted and unbootable OS.

MagerValp:
Further experimentation confirms what you say and shows that this approach was wrong. There are still issues with how Munki interacts with the Setup Assistant, but it requires more thought to solve. I worked around it by a manual run before rebooting to Setup Assistant, and not using Munki's bootstrapping.

Issue #746, opened 2017-04-09
Missing Japanese localizations


gregneagle:
@taniguti and cc @arubdesu 

While preparing for an eventual Munki 3 release, I'm going through the localizations. I noticed two strings that must have been added to Managed Software Center at some point since the last localization pass that were never localized into most of the supported languages, including Japanese:

/* System configuration problem alert title */
"System configuration problem" = "System configuration problem";

/* System configuration problem alert detail */
"A systems configuration issue is preventing Managed Software Center from operating correctly. The reported issue is: " = "A systems configuration issue is preventing Managed Software Center from operating correctly. The reported issue is: ";

There are also new "Featured" and "Featured items" labels:

/* FeaturedLabel */
"Featured" = "Featured";

/* FeaturedItemsHeaderText */
"Featured items" = "Featured items";

Finally, there is some Notification Center notification text which probably needs a native speaker's eyes:

/* Multiple Update message */
"%@ pending updates" = "%@個のアップデートを保留しています";

/* One Update message */
"1 pending update" = "1つのアップデートを保留しています";

/* Details label */
"Details" = "詳細";

/* Forced Install Date summary */
"One or more items must be installed by %@" = "%@までに、一つまたそれ以上のアイテムをインストールしなければいけません";

/* Software updates available message */
"Software updates available" = "利用可能なソフトウェアアップデート";

If you could help me out with Japanese translations here, it would be appreciated.

-Greg

PR #747, opened 2017-04-09
Updated JP localization

nakanana:
issue #746
There was a difference in %s - %@ in Managed Software Center file.

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center/ja.lproj/Localizable.strings
+++ b/code/apps/Managed Software Center/Managed Software Center/ja.lproj/Localizable.strings
@@ -1,8 +1,8 @@
 ﻿/* Multiple Updates message */
-"%s pending updates" = "%s個のアップデートを保留しています";
+"%@ pending updates" = "アップデートを%@個保留しています";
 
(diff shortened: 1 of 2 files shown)


code/apps/Managed Software Center/Managed Software Center/ja.lproj/Localizable.strings
```diff
@@ -1,8 +1,8 @@
 ﻿/* Multiple Updates message */
-"%s pending updates" = "%s個のアップデートを保留しています";
+"%@ pending updates" = "アップデートを%@個保留しています";
```
gregneagle: Actually since Managed Software Update is a Cocoa-Python application, the string substitution is done in Python, and %s is the string placeholder, which is different from Objective-C's %@ object placeholder

gregneagle: munki-notifier, on the other hand, _is_ a "normal" Cocoa app and does use %@ for the string object placeholders.

gregneagle:
Thank you!

PR #755, opened 2017-04-26
fix: Don't print blank lines during munkiimport

clburlison:
This line was printing between 20-50 blank lines on each munkiimport

### DIFF
--- a/code/client/munkiimport
+++ b/code/client/munkiimport
@@ -82,7 +82,9 @@ def make_dmg(pkgpath):
         output = proc.stdout.readline()
         if not output and (proc.poll() != None):
             break
-        print output.rstrip('\n').decode('UTF-8')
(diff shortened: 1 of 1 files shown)


erikng:
I've been wondering what this was. I figured it was new functionality that I hadn't kept up with.

gregneagle:
Suggest maybe doing the strip then printing only if the line as a len > 0.

clburlison:
You're right I should have looked at why the output was being printed in the first place. I've adjusted the commit to check for an output length greater than zero.

I just verified in a clean VM with Munki 3 beta 5 and the blank line printing is still happening so it's not something specific to my admin machine. Looking back at the commit history I'm not sure what caused that change. However, it's now working. Thanks Greg

PR #757, opened 2017-04-26
Update nl locale for MSC

bochoven:


bochoven:
Sorry about 2 PR's I assumed GH would allow me to change more than one file in the web ui.

PR #758, opened 2017-04-26
fix: Add basic PEP8 back into travis ci

clburlison:
Gonna leave this here. If @natewalck / @keeleysam have comments post below. 

The current implementation doesn't require the repo to have a separate script just for running a basic pep8 test. Plus it's off of travis slower legacy system with the addition of `sudo: no`.

If this gets merged someone with repo access will need to re-enable the Travis CI integration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
+sudo: no
 language: python
 python:
   - "2.7"
 install:
   - pip install flake8 pep8
 script:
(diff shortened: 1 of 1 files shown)


natewalck:
Looks good to me! Also consider the workflow. How will people verify their lint changes work. One reason I used an external script was so they could copy paste the command. (To be fair, they can copy paste this command as well).

Issue #749, opened 2017-04-11
Missing Russian localizations


gregneagle:
/* System configuration problem alert title */
"System configuration problem" = "System configuration problem";

/* System configuration problem alert detail */
"A systems configuration issue is preventing Managed Software Center from operating correctly. The reported issue is: " = "A systems configuration issue is preventing Managed Software Center from operating correctly. The reported issue is: ";

There are also new "Featured" and "Featured items" labels:

/* FeaturedLabel */
"Featured" = "Featured";

/* FeaturedItemsHeaderText */
"Featured items" = "Featured items";

Finally, there is some Notification Center notification text which probably needs a native speaker's eyes:

/* Multiple Update message */
"%@ pending updates" = "%@ отложенных обновлений";

/* One Update message */
"1 pending update" = "1 отложенное обновление";

/* Details label */
"Details" = "Подробнее";

/* Forced Install Date summary */
"One or more items must be installed by %@" = "Один или несколько элементов должны быть установлены до %@";

/* Software updates available message */
"Software updates available" = "Обновления программного обеспечения, доступные";

timsutton:
I should be able to find someone who can take a look at this..

gregneagle:
Doesn't @groob speak Russian?

groob:
He speaks it, but didn't have computers growing up so he only learned computer interface details in English. 

...I'll try to send a PR for this.

gregneagle:
See https://github.com/munki/munki/wiki/Munki-3-Localization-needs for the current list.

timsutton:
I've also asked a couple people if they can take a look. If they don't, I'll ask a couple more :)

timsutton:
Looks like this one can now be closed (via #759 and #760).

PR #759, opened 2017-05-01
Munki3 ru updates

timsutton:
For #749.

Only missing the "Featured" / "Featured items". Apple uses Подборка for "Featured", but my weak grasp of Russian can't tell me whether that makes "Featured items" "Подборка элементы". @groob, would you have any suggestions (or any other feedback on the others)?

I also found one missing translation for `Removal Error message` and tried it myself.

Separately I noticed a `systemsadministrator` typo in the original English, but left that out of this PR.


gregneagle:
"Separately I noticed a systemsadministrator typo in the original English" I keep fixing that, and it keeps coming back. I think I finally realized why, though...

gregneagle:
And now that I fixed that error, this PR "has conflicts that must be resolved".

gregneagle:
And we're also missing the label for the Password field:

```
    /* Password label */ (jp and ru only)
        en: Password:
        jp: 
        ru:
```

PR #760, opened 2017-05-01
Add 'Featured'/'Featured items' ru strings

timsutton:
One more Russian update for #749, this just adds the 'Featured' items (the one I [suggested](https://github.com/munki/munki/pull/759#issue-225447860) was wrong, but this is correct according to a native speaker).


gregneagle:
Great! Can we get a localized label for the Password field?

timsutton:
That one should be in there too, unless I'm missing which one you're referring to: https://github.com/munki/munki/blob/Munki3/code/apps/Managed%20Software%20Center/Managed%20Software%20Center/ru.lproj/Localizable.strings#L257

gregneagle:
Thanks.

PR #772, opened 2017-05-11
fix icon_hash

keeleysam:


gregneagle:
I fixed it a different way here: https://github.com/munki/munki/commit/2e56c59b62871f83c60b604685ca1544f09badbf

See if that works for you. (I did it this other way as a step to the next set of changes...)

keeleysam:
yep, that produces the same results for me

PR #783, opened 2017-06-17
No hash hashing

keeleysam:
`repo.itemlist('icons')` will return `_icon_hashes.plist` if it exists, so remove that from our list so there's not a loop with a changing hash list on every `makecatalogs` run.


gregneagle:
I had considered naming the file .icon_hashes.plist, which would have prevented `repo.itemlist('icons')` from returning it, but I worried that some/many web servers might be configured to _not_ serve files that start with a period.

PR #784, opened 2017-06-27
Fixing postflight script arguements

weswhet:
I noticed that the postflight scripts I had always had 'osinstall' passed as the run type arguement. This is an attempted fix for the previous behavior.

### DIFF
--- a/code/client/managedsoftwareupdate
+++ b/code/client/managedsoftwareupdate
@@ -249,7 +249,8 @@ def doInstallTasks(do_apple_updates, only_unattended=False):
             reports.savereport()
             # install macOS
             try:
-                success = osinstaller.run(finishing_tasks=doFinishingTasks)
+                success = osinstaller.run(finishing_tasks=doFinishingTasks(
+                    runtype='osinstall'))
             except:
                 display.display_error(
                     'Unexpected error in munkilib.osinstaller:')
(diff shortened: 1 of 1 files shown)


code/client/managedsoftwareupdate
```diff
@@ -249,7 +249,8 @@ def doInstallTasks(do_apple_updates, only_unattended=False):
             reports.savereport()
             # install macOS
             try:
-                success = osinstaller.run(finishing_tasks=doFinishingTasks)
+                success = osinstaller.run(finishing_tasks=doFinishingTasks(
+                    runtype='osinstall'))
```
gregneagle: The prior version passes the function pointer (a pointer to the doFinishingTasks function). Your change passes the _result_ of _calling_ that function. IOW, `doFinishingTasks(runtype='osinstall')` is called, and its return value is passed to `osinstaller.run()`.  So the "finishing tasks" are run _before_ `osinstaller.run()` is called, and `osinstaller.run()` is being passed the result of a function instead of a function pointer. (Hopefully the result of `doFinishingTasks(runtype='osinstall')` is `None`, or we're likely to get a traceback.)

gregneagle:
See if my most recent commit solves this issue for you. Seems to here, though I haven't tested the behavior during an OS install.

weswhet:
yeah, that should do it! Thanks

PR #785, opened 2017-06-28
FR localization of "Requires macOS version %s"

RichardLafontaine:


gregneagle:
Can you move the new text between:

/* Install Required action text */
"Required" = "obligatoire";

/* Restart button title */
"Restart" = "Redémarrer";

Some other tools I use to check and diff the localized text will resort these anyway...

gregneagle:
Merci!

PR #793, opened 2017-08-10
LastRunWithNoErrors

mikedodge04:
LastSuccessfulRun to ManagedInstallReport

$ cat /Library/Managed\ Installs/ManagedInstallReport.plist |grep Last -A1
	<key>LastSuccessfulRun</key>
	<string>2017-08-10 20:34:20 +0000</string>

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center.xcodeproj/project.pbxproj
+++ b/code/apps/Managed Software Center/Managed Software Center.xcodeproj/project.pbxproj
@@ -9,6 +9,7 @@
 /* Begin PBXBuildFile section */
 		650B29A419B69FC800A5E946 /* MSCToolbar.py in Resources */ = {isa = PBXBuildFile; fileRef = 650B29A319B69FC800A5E946 /* MSCToolbar.py */; };
 		C00A4C57185FCEC9004EB3B7 /* FoundationPlist.py in Resources */ = {isa = PBXBuildFile; fileRef = C00A4C56185FCEC9004EB3B7 /* FoundationPlist.py */; };
+		C00F67571F016C9F00D9007D /* CocoaWrapper.py in Resources */ = {isa = PBXBuildFile; fileRef = C00F67561F016C9F00D9007D /* CocoaWrapper.py */; };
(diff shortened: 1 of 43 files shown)


gregneagle:
Two things:

1) A new feature should go into Munki3dev and not master

2) Your definition of "LastSuccessfulRun" might not match others' definitions. What problem are you trying to solve with this that isn't solved with a reporting console like Sal or MunkiReport or similar?

gregneagle:
one more thing:

If there are errors in this run, then LastSuccessfulRun is not set. Does this value get copied over from a previous run, then?

gregneagle:
(my point about your definition of "LastSuccessfulRun": you could have a bogus item in a manifest that caused an error every run because Munki could not find Foo in any available catalogs. Yet it successfully installs many other items, maybe for weeks. What useful/actionable info is given by setting LastSuccessfulRun?

mikedodge04:
I want a way to understand client side whether munki is in a good state. If LastSuccessfulRun is > 30days. I plan on forgetting the pkgs and reinstalling.

gregneagle:
You plan on forgetting what packages and reinstalling what?

mikedodge04:
all munki pkgs, and my local manifest :)

gregneagle:
That seems awfully Facebook-specific. Rarely are errors during a Munki run resolved by reinstalling the Munki packages on the client: they are almost always caused by server-side issues with a manifest or catalog items.

I think this key/value will cause a lot of confusion for most Munki admins.

You could do this for yourself with no code changes to Munki in a Munki postflight script.

gregneagle:
(also note the conflicts; you'll need to fix the PR in any case)

mikedodge04:
LastSuccessfulRun is still correct though. Would you be ok if we changed the key name? I want to update my opensource cpe_munki cookbook to have an remediate flag.

gregneagle:
Change the key name to LastRunWithNoErrors or something similar. "LastSuccessfulRun" to too open to individual admin opinion.

gregneagle:
As the "First-time contributor" badge reminded me: I need you to submit a CLA:
https://github.com/munki/munki/wiki/Contributing-to-Munki

mikedodge04:
/Library/Preferences/ManagedInstalls LastCheckDate instead

gregneagle:
...wondering why you reopened this, and when you'll be covered by the CLA...

mikedodge04:
My people are working on that as we speak :)  @natewalck  asked me to open it again

PR #794, opened 2017-08-23
An option to silently skip pkg check

ygini:
Add --skip-pkg-check (-s) option
This PR add a manual option to skip the pkg_check process in addition to PackageCompleteURL & PackageURL based detection.
Manual request for this option is usefull when a unique server contain data and metadata but both are push from different sources. (Like GIT + rsync)
The difference with the force option rely in the warning message that are avoided here and the check done on the metadata part aren't skiped.

### DIFF
--- a/code/client/makecatalogs
+++ b/code/client/makecatalogs
@@ -318,7 +318,7 @@ def main():
     parser.add_option('--force', '-f', action='store_true', dest='force',
                       help='Disable sanity checks.')
     parser.add_option('--skip-pkg-check', '-s', action='store_true', dest='skip_payload_check',
-                      help='Skip check of pkg infos. Useful'
+                      help='Skip checking of pkg existence. Useful'
                            'when pkgs aren\'t on the same server'
                            'as pkginfo, catalogs and manifests.')
     parser.add_option('--repo_url', '--repo-url',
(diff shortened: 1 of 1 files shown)


code/client/makecatalogs
```diff
@@ -315,13 +317,17 @@ def main():
                       help='Print the version of the munki tools and exit.')
     parser.add_option('--force', '-f', action='store_true', dest='force',
                       help='Disable sanity checks.')
+    parser.add_option('--skip-pkg-check', '-s', action='store_true', dest='skip_payload_check',
+                      help='Skip check of pkg infos. Usefull'
```
gregneagle: "Useful" not "Useful"

ygini:
What do you think about "Silently skip actual payload check but keep check related to pkginfo. Useful when manipulating munki infos without payload accessible locally (for example, munki repo on git with pkg content elsewhere)"

gregneagle:
Why not my original suggestion?

Skip checking of pkg existence. Useful when pkgs aren't on the same server as pkginfo, catalogs and manifests.

ygini:
Because I really need to take time to read the full sentence and catch all infos. Sorry for wasting your time. Doing multiple things in the same time isn't the best thing to do I suppose ^^

gregneagle:
Finally, this needs to be targeted at Munki3dev and not master. I can do that manually, I suppose.

ygini:
Question for that, the source code for makecatalogs in munki3dev (last change June 17) isn't the one used in the munki 3.0.3 package published on July 13.

Which one should I update? I can port the change I've did to the munki3dev implementation if you want.

gregneagle:
All the more reason to target Munki3dev, which will be the basis for the 3.1 release.

ygini:
PR for the other implementation sent. We can close this one.

PR #795, opened 2017-08-23
An option to silently skip pkg check (munki3dev)

ygini:
Add --skip-pkg-check (-s) option
This PR add a manual option to skip the pkg_check process in addition to PackageCompleteURL & PackageURL based detection.
Manual request for this option is usefull when a unique server contain data and metadata but both are push from different sources. (Like GIT + rsync)
The difference with the force option rely in the warning message that are avoided here and the check done on the metadata part aren't skiped.

### DIFF
--- a/code/client/makecatalogs
+++ b/code/client/makecatalogs
@@ -162,7 +162,7 @@ def verify_pkginfo(pkginfo_ref, pkginfo, pkgs_list, errors):
     return True
 
 
-def process_pkgsinfo(repo, icons, force=False):
+def process_pkgsinfo(repo, icons, force=False, skip_payload_check=False):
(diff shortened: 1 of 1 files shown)


gregneagle:
Looks good.

gregneagle:
@ygini I forgot: I need a signed Contributor License Agreement: https://github.com/munki/munki/wiki/Contributing-to-Munki  or I'll have to back out this change. Sorry!

ygini:
@gregneagle my memory was unclear on this, I was thinking that I've already did it in the past. Anyway, I've just sent it.

gregneagle:
And received. Thanks!

PR #797, opened 2017-08-24
Add uninstall-pkg command to managed_uninstalls in manifestutil

poundbangbash:
Added a command `uninstall-pkg` to manifestutil to move a pkg from a section to the managed_uninstalls section instead of having to run two commands: `remove-pkg` then `add-pkg`.

CLA submitted.

-Eric

### DIFF
--- a/code/client/manifestutil
+++ b/code/client/manifestutil
@@ -624,6 +624,65 @@ def add_pkg(repo, args):
         return 1 # Operation not permitted
 
 
+def uninstall_pkg(repo, args):
+    '''Moves a package from one section to managed_uninstalls in a manifest.'''
+    parser = MyOptionParser()
+    parser.set_usage(
+        '''uninstall-pkg PKGNAME --manifest MANIFESTNAME [--section SECTIONNAME]
+       Moves a package from one section to managed_uninstalls in a manifest. Package is removed from
+       managed_installs unless a different manifest section is specified with
+       the --section option''')
+    parser.add_option('--manifest',
+                      metavar='MANIFESTNAME',
+                      help='''name of manifest on which to operate''')
+    parser.add_option('--section', default='managed_installs',
+                      metavar='SECTIONNAME',
+                      help=('manifest section from which to move the '
+                            'package to. Defaults to managed_uninstalls.'))
+    try:
+        options, arguments = parser.parse_args(args)
+    except MyOptParseError, errmsg:
+        print >> sys.stderr, str(errmsg)
+        return 22 # Invalid argument
+    except MyOptParseExit:
+        return 0
+
+    if not options.manifest:
+        if len(arguments) == 2:
+            options.manifest = arguments[1]
+            del arguments[1]
+        else:
+            parser.print_usage(sys.stderr)
+            return 7 # Argument list too long
+    if len(arguments) != 1:
+        parser.print_usage(sys.stderr)
+        return 7 # Argument list too long
+    pkgname = arguments[0]
+
+    manifest = get_manifest(repo, options.manifest)
+    if not manifest:
+        return 2 # No such file or directory
+    if not options.section in manifest:
+        print >> sys.stderr, ('Section %s is not in manifest %s.'
+                              % (options.section, manifest))
+        return 1 # Operation not permitted
+    if pkgname not in manifest[options.section]:
+        print >> sys.stderr, ('Package %s is not in section %s '
+                              'of manifest %s.'
+                              % (pkgname, options.section, options.manifest))
+        return 1 # Operation not permitted
+    else:
+        manifest[options.section].remove(pkgname)
+        manifest['managed_uninstalls'].append(pkgname)
+        if save_manifest(repo, manifest, options.manifest,
+                         overwrite_existing=True):
+            print ('Removed %s from section %s and added to section managed_uninstalls of manifest %s.'
+                   % (pkgname, options.section, options.manifest))
+            return 0
+        else:
+            return 1 # Operation not permitted
+
+
 def remove_pkg(repo, args):
     '''Removes a package from a manifest.'''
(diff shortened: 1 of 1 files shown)


gregneagle:
While I understand the appeal and usefulness of this subcommand, it seems odd. There's no `install-pkg` subcommand to add an item to `managed_installs` and remove it from `managed_uninstalls`, for example.

Looking at the code, it also seems to error if there is no existing pkgname in `managed_installs` (or whichever section you specify) which seems to mean you need to check for its existence in `managed_installs` before using `uninstall-pkg`. That seems to reduce the usefulness/convenience of this subcommand. Am I misunderstanding?

poundbangbash:
I thought about creating the subcommand as `move-pkg` and specifying the sections to move from and to but as I thought about how I move packages around manifests 99% of the time it is to remove an install from the managed_installs and add to the managed_uninstalls.  I don't recall ever moving anything from being uninstalled back to being installed nor moving anything from being a managed_install to an optional_install. If those cases do arise the `remove-pkg` `add-pkg` method can be used.  This subcommand is a selfish one in that it takes fewer steps to do a common task. It's not a catch all for every scenario, just a handy one to have for the main use case.

The error if a package is not in a section to remove is the same as the `remove-pkg` subcommand. In fact, this uninstall subcommand code was copied from the remove-pkg code and modified. The attempt to remove the pkg stops the process and nothing is added to managed_uninstalls

> display-manifest uninstall-client
 catalogs:
     home
 included_manifests:
 managed_installs:
 managed_uninstalls:
> uninstall-pkg VLC --manifest uninstall-client
Package VLC is not in section managed_installs of manifest uninstall-client.
> display-manifest uninstall-client
 catalogs:
     home
 included_manifests:
 managed_installs:
 managed_uninstalls:


-Eric

On Aug 24, 2017, at 12:09 PM, Greg Neagle <notifications@github.com> wrote:

While I understand the appeal and usefulness of this subcommand, it seems odd. There's no install-pkg subcommand to add an item to managed_installs and remove it from managed_uninstalls, for example.
Looking at the code, it also seems to error if there is no existing pkgname in managed_installs (or whichever section you specify) which seems to mean you need to check for its existence in managed_installs before using uninstall-pkg. That seems to reduce the usefulness/convenience of this subcommand. Am I misunderstanding?

gregneagle:
I guess it seems to me as a convenience command you'd want the behavior to be:

1) Add to managed_uninstalls
2) _If_ it's in managed_installs (or other section if specified, excepting managed_uninstalls), remove it from there. If it's not in managed_installs, just exit.

poundbangbash:
I think I like the order that it is in now to verify that the item I'm intending to uninstall is removed from the proper section.  For those that want to just add to managed_uninstalls I'd use `add-pkg` to that --section.

Take this manifest for example:

 catalogs:
     home
 included_manifests:
 managed_installs:
 managed_uninstalls:
 optional_installs:
     VLC

If the admin chose the wrong section and attempted to make the change `uninstall-pkg VLC --manifest test --section managed_installs`, the order of operation suggested would place the item in the managed_uninstalls first, attempt to remove the item from managed_installs, then exit quietly because it is not in managed_installs. Since the item wasn't in the specified section the manifest would have the item in both the managed_uninstalls and still remain in the optional_installs causing a conflict.

-Eric

gregneagle:
That's not really a significant conflict, though: Munki knows what to do in that situation: ignore the optional_installs instance and pay attention to the managed_uninstalls instance.

So what if it printed a warning if the item is not in the expected section from which to remove it? Otherwise it just doesn't seem convenient enough to bother adding -- I know I'd probably never use it as currently implemented.

poundbangbash:
Sorry, I didn't see this reply last week.

I made changes to incorporate your suggestion however I'm still leery about the case where the wrong section could be provided and the pkg will still be placed in the managed_uninstalls section.  For example, below VLC is already in `managed_installs` and the admin specifies to uninstall from the `optional_installs` section.  The result is that a warning is given about it not being in `optional_installs`, leaving it in `managed_installs` and also being put in `managed_uninstalls`. The software would never attempt to be uninstalled since it's listed in `managed_installs`. I'm not sure how to address that elegantly.  One option would be to search the other sections for the pkg if it's not found in the section specified but then what would be the right thing to do with that information? 

```
> display-manifest uninstall-client
 catalogs:
     home
 included_manifests:
 managed_installs:
     GoogleChrome
     VLC
 managed_uninstalls:
 optional_installs:

> uninstall-pkg VLC --manifest uninstall-client --section optional_installs
WARNING: Package VLC is not in section optional_installs of manifest uninstall-client.
Added VLC to section managed_uninstalls of manifest uninstall-client.

> display-manifest uninstall-client
 catalogs:
     home
 included_manifests:
 managed_installs:
     GoogleChrome
     VLC
 managed_uninstalls:
     VLC
 optional_installs:
```

The way the code is submitted now would stop the command before adding the pkg to the managed_uninstalls section and output a warning causing the admin to go investigate why that command wasn't allowed.

-Eric

gregneagle:
Let's either simplify or abandon this. You wrote earlier: "99% of the time it is to remove an install from the managed_installs and add to the managed_uninstalls" so make it do just that, only that, and fail with an error if the item isn't present in managed_installs. Any other "move" can be done with the other verbs (`remove-pkg` and `add-pkg`)

code/client/manifestutil
```diff
@@ -723,7 +720,7 @@ def remove_pkg(repo, args):
         return 2 # No such file or directory
     if not options.section in manifest:
```
gregneagle: there is no options.section now, no?

poundbangbash:
How about `move-install-to-uninstall`?

gregneagle:
Works for me. Autocomplete at least means you don't have type the whole thing.

PR #800, opened 2017-09-01
fix: Correct issue with Auth Restart

### LINKED ISSUE #799: Munki 3 Auth Restart issue
### The issue
When doing an auth restart munki reads the `RecoveryKeyFile` which I had set as (yes this was an admin error and is now corrected):

```bash
            <key>RecoveryKeyFile</key>
            <data>
BAgw
            </data>
```
FoundationPlist is unable to process this data properly because it shouldn't exist and causes a exception in `authrestart/__init__.py`. This then halts the munki run.

### The fix?
Untested, fix is to add to include `NSInvalidArgumentException` in the exception located:

https://github.com/munki/munki/blob/98a489298536e24048244ca51d72f4c8c0fdb084/code/client/munkilib/authrestart/__init__.py#L124

This should cause munki to return a empty string and continue the run.

### Error log
[AuthRestartErr.txt](https://github.com/munki/munki/files/1265153/AuthRestartErr.txt)

(I'll test this tomorrow just ran out of time for today)

clburlison:
An invalid RecoveryKey can be set that causes FoundationPlist to throw
a ValueError. If this happens no restart happens. Instead we should log
the error and revert to a non-auth restart.

Tested that this does indeed Fix #799

### DIFF
--- a/code/client/munkilib/authrestart/__init__.py
+++ b/code/client/munkilib/authrestart/__init__.py
@@ -126,7 +126,7 @@ def get_auth_restart_key(quiet=False):
             display.display_error(
                 'We had trouble getting info from %s...', recoverykeyplist)
         return ''
-    except KeyError:
+    except (KeyError, ValueError):
         if not quiet:
             display.display_error(
                 'Problem with key: RecoveryKey in %s...', recoverykeyplist)


gregneagle:
I tested a few unexpected data types against FoundationPlist.readPlist() and it does look like ValueError is the appropriate exception:

```
$ python
Python 2.7.10 (default, Feb  7 2017, 00:08:15) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from munkilib import FoundationPlist
>>> FoundationPlist.readPlist(1)
2017-09-01 08:15:44.685 Python[98188:4850537] -[OC_PythonNumber getFileSystemRepresentation:maxLength:]: unrecognized selector sent to instance 0x7fae00d7db10
  File "<stdin>", line 1, in <module>
  File "munkilib/FoundationPlist.py", line 73, in readPlist
    plistData = NSData.dataWithContentsOfFile_(filepath)
ValueError: NSInvalidArgumentException - -[OC_PythonNumber getFileSystemRepresentation:maxLength:]: unrecognized selector sent to instance 0x7fae00d7db10
>>> FoundationPlist.readPlist([])
2017-09-01 08:15:54.277 Python[98188:4850537] -[OC_PythonArray getFileSystemRepresentation:maxLength:]: unrecognized selector sent to instance 0x7fae00f05260
  File "<stdin>", line 1, in <module>
  File "munkilib/FoundationPlist.py", line 73, in readPlist
    plistData = NSData.dataWithContentsOfFile_(filepath)
ValueError: NSInvalidArgumentException - -[OC_PythonArray getFileSystemRepresentation:maxLength:]: unrecognized selector sent to instance 0x7fae00f05260
>>> FoundationPlist.readPlist(buffer('hello'))
2017-09-01 08:17:10.300 Python[98188:4850537] -[OC_PythonData getFileSystemRepresentation:maxLength:]: unrecognized selector sent to instance 0x7fae00d98bf0
  File "<stdin>", line 1, in <module>
  File "munkilib/FoundationPlist.py", line 73, in readPlist
    plistData = NSData.dataWithContentsOfFile_(filepath)
ValueError: NSInvalidArgumentException - -[OC_PythonData getFileSystemRepresentation:maxLength:]: unrecognized selector sent to instance 0x7fae00d98bf0
```

Issue #802, opened 2017-09-12
managedsoftwareupdate restart flag is overridden by nopkg script.


chymb:
Hi,
I noticed that our deployment bootstrapping (which makes heavy use of nopkg scripts) was skipping a reboot and tracked it down to the following line. The "=" overrides a previous True value if the nopkg script requires no restart.

https://github.com/munki/munki/blob/91b43fdbe84b661c6d08bcb581186cc600f6b7b0/code/client/munkilib/installer/core.py#L306

Cheers,
Chris

gregneagle:
Wanted this fix to go into tomorrow's RC, so I did my own version of the fix:
https://github.com/munki/munki/commit/eb17f853fb7083d50c6ce778fe57bcb36bd9e86b

Thanks for finding this!

PR #803, opened 2017-09-12
Fixed bug that would clear restart requirement

chymb:
This patch fixes issue #802 which causes managedsoftwareupdate to skipping a reboot. The "=" overrides a previous True value for "restartflag" if the nopkg script requires no restart.

### DIFF
--- a/code/client/munkilib/installer/core.py
+++ b/code/client/munkilib/installer/core.py
@@ -303,7 +303,7 @@ def install_with_info(
                     retcode = -1
             # nopkg (Packageless) install
             elif installer_type == "nopkg":
-                restartflag = requires_restart(item)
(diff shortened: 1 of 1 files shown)


gregneagle:
Thanks for this. I know this is a minor change, but my organization requires a signed CLA from all contributors: https://github.com/munki/munki/wiki/Contributing-to-Munki

If you don't want to do that, that's fine, I can implement the required fix on my own.  :-)

gregneagle:
Wanted this fix to go into tomorrow's RC, so I did my own version of the fix:
https://github.com/munki/munki/commit/eb17f853fb7083d50c6ce778fe57bcb36bd9e86b

Thanks for finding this!

PR #809, opened 2017-10-12
Localized pkgsinfo support

bochoven:
This PR adds support for localising pkgsinfo files so the locale settings of the client machine will show the appropriate strings. This PR introduces a new pkgsinfo key `localized_strings`. The type is a dict of dicts. See the example below:

```xml
	<key>localized_strings</key>
	<dict>
		<key>nl</key>
		<dict>
			<key>category</key>
			<string>Wetenschappelijke Software</string>
			<key>description</key>
			<string>Nederlandse beschrijving</string>
			<key>display_name</key>
			<string>Nederlandse naam</string>
			<key>preinstall_alert</key>
			<dict>
				<key>alert_detail</key>
				<string>Er kan wat gebeuren</string>
				<key>alert_title</key>
				<string>Let op</string>
				<key>cancel_label</key>
				<string>Stop</string>
				<key>ok_label</key>
				<string>Ga door</string>
			</dict>
		</dict>
	</dict>
```

# Structure

The key(s) of the outer dict should match the desired locale, for instance `nl` matches the Dutch locale.
The inner dict should contain one or more of the following pkgsinfo keys: `category`,  `description` and `display_name`. The values should contain the localised strings.
There's also support for these dictionaries: `preinstall_alert`,  `preuninstall_alert`, `preupgrade_alert`.

# Language code

To get a list of language codes, you can use the following python snippet:

```python
from Foundation import NSLocale
print NSLocale.ISOLanguageCodes()
```

# Fallback language

To indicate the default language that the strings in the pkgsinfo are in, you should add the `fallback_locale` key to the `localized_strings` dict.
```xml
	<key>localized_strings</key>
	<dict>
		<key>fallback_locale</key>
		<string>en</string>
	</dict>
```

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center/CocoaWrapper.py
+++ b/code/apps/Managed Software Center/Managed Software Center/CocoaWrapper.py
@@ -29,7 +29,6 @@ from Foundation import (
     NSFileHandle,
     NSFileManager,
     NSInsetRect,
-    NSLocale,
     NSLocalizedString,
     NSLog,
     NSMakePoint,
--- a/code/apps/Managed Software Center/Managed Software Center/CocoaWrapper.py
+++ b/code/apps/Managed Software Center/Managed Software Center/CocoaWrapper.py
@@ -29,6 +29,7 @@ from Foundation import (
     NSFileHandle,
     NSFileManager,
     NSInsetRect,
+    NSLocale,
     NSLocalizedString,
     NSLog,
     NSMakePoint,
--- a/code/apps/Managed Software Center/Managed Software Center/MunkiItems.py
+++ b/code/apps/Managed Software Center/Managed Software Center/MunkiItems.py
@@ -534,6 +534,8 @@ class GenericItem(dict):
 
     def __init__(self, *arg, **kw):
         super(GenericItem, self).__init__(*arg, **kw)
+        if self.get('localized_strings'):
+            self.add_localizations()
         # now normalize values
         if not self.get('display_name'):
             self['display_name'] = self['name']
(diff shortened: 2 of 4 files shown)


gregneagle:
I really like the idea and the proposed layout of the localized strings in the pkginfo.

But I think using NSLocale.currentLocale() to decide which localization to use is too simplistic.

Instead you want to get a list of available localizations and then use something like NSBundle.preferredLocalizationsFromArray: to figure out the "best" localization. 

This mechanism takes into account the order of language preferences in System Preferences' Language and Region pane, so that if my first preferred language is Dutch, my second is German, and my third is English -- if there are no localized strings for Dutch but there are strings for German it should choose to display German instead of falling back to the non-localized strings. (and it's probably bad to assume the non-localized are English, but that's a different issue.)

Alternately you could use NSLocale.preferredLanguages() and iterate through those:

```
NSLocale.preferredLanguages()
(
    "de-US",
    "en-US",
    "sv-US",
    "fr-US",
    "es-US"
)
```

This might do the wrong thing and show me Swedish or French or Spanish even though the non-localized strings are in English, so maybe your proposed dictionary needs a key/value pair to hint what language the non-localized strings are in (we can't assume English).

gregneagle:
I forgot to give an example of NSBundle.preferredLocalizationsFromArray_forPreferences_. Here's an example where my language preferences are as above ("de-US", "en-US", "sv-US", etc), and the available localized strings are 'nl', 'de', 'en':

```
>>> NSBundle.preferredLocalizationsFromArray_forPreferences_(['nl', 'de', 'en'], None)
(
    de
)
```
This seems easier to me to use than trying to work with the ("de-US", "en-US", "sv-US", etc)-style strings.

bochoven:
Good points, made the changes. I also added info on the `fallback_locale` to the commit message.

gregneagle:
Why do this only for optional_items? I'd think we'd do it for all items, so that managed installs would also show localized names and descriptions.

bochoven:
Last commit adds `localized_strings` support for managed installs.

gregneagle:
I'll probably merge this later today after another review, but another thing to think about before we'd release this: we'll want `munkiimport` to copy the localized strings from the previous version to a new import.

bochoven:
Would adding the key to this list suffice? https://github.com/munki/munki/blob/master/code/client/munkiimport#L952-L962

gregneagle:
I think so, yes.

bochoven:
Tested:

```sh
Use existing item as a template? [y/n] y
Copying unattended_install: False
Copying unattended_uninstall: False
Copying category: ClamXAV
Copying developer: ClamXAV
Copying localized_strings: {
    de =     {
        description = "Duitse beschrijving";
        "display_name" = "Duitse ADF";
    };
    "fallback_locale" = en;
    nl =     {
        category = "Wetenschappelijke Software";
        description = "Nederlandse beschrijving <b>ClamXAV</b>";
        "display_name" = "Nederlandse ClamXav";
        "preinstall_alert" =         {
            "alert_detail" = "Er kan wat gebeuren";
            "alert_title" = "Let op";
            "cancel_label" = Stop;
            "ok_labels" = "Ga door";
        };
    };
}
```

gregneagle:
Let's get your notes above into the wiki.

gregneagle:
Another idea should anyone want to work on it: make Apple Updates behave the same way. There's some localization support now, but it could be even better.

arubdesu:
(Pardon that I just wanted to chime in to say this is sooo cool and I really appreciate your work @bochoven !)

bochoven:
Notes are in the wiki. I sent the PR against the munki3dev branch, but I can't find the instructions that told me to do that. What is the preferred branch @gregneagle?

gregneagle:
Munki3dev is the right place for new features. Small bug fixes go right to master.

gregneagle:
"Notes are in the wiki." Link, please?

bochoven:
https://github.com/munki/munki/wiki/Supported-Pkginfo-Keys
and
https://github.com/munki/munki/wiki/Pkginfo-Files#localized_strings

PR #810, opened 2017-10-13
Removing references to Managed Software Update in MunkiStatus

fuzzylogiq:
Noticed this while poking around for rebranding purposes. A couple of the MainMenu.strings for MunkiStatus still have the window title "Managed Software Update" even though their localised translation is now "Managed Software Center".

N.B. I believe this is the only bit of MunkiStatus that is actually user-facing. There are other problems in the localised MainMenu.strings files which all refer to Managed Software Center or its translations rather than MunkiStatus, but not sure if it's worth correcting if user will never see them?


gregneagle:
Did you check the localizations in Managed Software Center for da and fi? The intent would be for the MunkiStatus window to have the same title as (localized) Managed Software Center.

fuzzylogiq:
Yup, in fact that’s _why_ I noticed. Because when I was doing a search replace for Managed Software Center (which is untranslated from US English in the da and fi translations), I noticed these two weren’t changing the window title of MunkiStatus but were changing all other references in MSC.app

PR #820, opened 2017-12-01
Add a custom header with the serial number

np5:
I could add a configuration option too. IncludeSerialNumberHeader boolean ?


gregneagle:
I don't understand the purpose of this proposed change: what problem it solves; what application it has.

A Munki preflight script would be another place to construct arbitrary http headers and would require no code changes to Munki.

np5:
This could be used to detect when machines are cloned for example. But you are right, we could do it with a preflight script. I didn't think of that.

groob:
@np5 you also have osquery at your disposal to audit the munki configs accross your fleet ;)

PR #821, opened 2017-12-07
Added variable quoting

poundbangbash:
Added variable quoting to allow spaces and special characters (like parens) in the path name.


gregneagle:
How odd that no-one has run any of this from a path containing spaces in the past decade and hit this.

poundbangbash:
I blame "Dropbox (Personal)"

PR #824, opened 2017-12-26
Remove extraneous quote

ChefAustin:
Unnecessary trailing double-quote. Kill it!


homebysix:
For the sake of completion, you might also consider making the same change to **make_munki_mpkg.sh** and **make_munki_mpkg_from_git.sh**.

ChefAustin:
@homebysix  Good call; thanks for pointing that out!

I've updated the branch to reflect removing all 3x instances of that extraneous `"`.

gregneagle:
Closing this no-longer relevant PR.

PR #837, opened 2018-02-05
Implemented optional_installs for localonly manifest

tcinbis:
Hi everyone,
current configuration tools (chef, puppet or ansible) can be used to manage `manged_installs` and `managed_uninstalls` by manipulating the local only manifest file. 
This PR would add support for the optional install key.

For example: 
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>managed_installs</key>
	<array>
		<string>MunkiAdmin</string>
		<string>Remote Desktop</string>
		<string>OracleJava9JRE</string>
		<string>Mathematica</string>
		<string>XCode_commandline_tools</string>
		<string>Recipe Robot</string>
		<string>OracleJava8JRE</string>
		<string>uniprintd</string>
		<string>HipChat</string>
	</array>
	<key>managed_uninstalls</key>
	<array/>
	<key>optional_installs</key>
	<array>
		<string>Adobe_Premiere_Pro_CC_EN_2018</string>
		<string>MAXQDA12</string>
		<string>GoogleEarth</string>
		<string>munkimanifest</string>
	</array>
</dict>
</plist>
```
So far I tested if the logic is still working as expected for the following situations: 
* item is in localonly optional_installs **and** localonly managed installs
* item is in localonly optional_installs and managed installs (manifest on the server)
* item is in localonly optional_installs **and** optional installs (manifest on the server)

Could you please check and test it in your development environments?

### DIFF
--- a/code/client/munkilib/updatecheck/core.py
+++ b/code/client/munkilib/updatecheck/core.py
@@ -173,10 +173,14 @@ def check(client_id='', localmanifestpath=None):
                 localonlyuninstalls = manifestutils.get_manifest_value_for_key(
                     localonlymanifest, 'managed_uninstalls') or []
                 for item in localonlyuninstalls:
-                    analyze.process_removal(item, cataloglist, installinfo)
+                    dummy_result = analyze.process_removal(
+                        item,
+                        cataloglist,
+                        installinfo
+                    )
 
-                localonlyoptionals = manifestutils.get_manifest_value_for_key(localonlymanifest,
-                                                                              'optional_installs') or []
+                localonlyoptionals = manifestutils.get_manifest_value_for_key(
+                    localonlymanifest, 'optional_installs') or []
                 for item in localonlyoptionals:
                     dummy_result = analyze.process_optional_install(
                         item,
--- a/code/client/munkilib/updatecheck/core.py
+++ b/code/client/munkilib/updatecheck/core.py
@@ -174,6 +174,16 @@ def check(client_id='', localmanifestpath=None):
                     localonlymanifest, 'managed_uninstalls') or []
                 for item in localonlyuninstalls:
                     analyze.process_removal(item, cataloglist, installinfo)
+
+                localonlyoptionals = manifestutils.get_manifest_value_for_key(localonlymanifest,
+                                                                              'optional_installs') or []
+                for item in localonlyoptionals:
+                    dummy_result = analyze.process_optional_install(
+                        item,
+                        cataloglist,
+                        installinfo
+                    )
+
             else:
                 display.display_debug1(
                     "LocalOnlyManifest %s is set but is not present. "


erikng:
This seems like a good addition and I'm curious why it wasn't put in the original PR.

Perhaps @nmcspadden @natewalck or @mikedodge04 could add some context?

clburlison:
The only mention I can see regarding why optional was not included is:

> The feature is basic right now and only handles managed_install and managed_uninstalls. As this is not self-service, it behaves more like the main manifest than selfservemanifest.
> Nate https://github.com/munki/munki/pull/520#issue-107971022

With that said I think this adds complexity but I don't have to use this feature so 🤷‍♂️  

References:
1. Mailing list discussion [LocalOnlyManifest Idea](https://groups.google.com/forum/#!searchin/munki-dev/localonly/munki-dev/lgnBUFyMfAo/eSeTynSaBQAJ)
1. Original implementation #520

aschwanb:
Thanks to @clburlison for providing some context. I'm pinging @nmcspadden @natewalck and @mikedodge04 again to see if they have anything to add.

We make heavy use of the localonly manifest. I think most organisations that use any form of configuration management do. This would be a great addition for us.

Can I help with anything to get this merged?

gregneagle:
I don't use this feature. It's inconvenient for me to test. I need the organizations who are using this feature to test it. Lack of participation here makes it less likely I will ever merge it as I can't support it. If LocalOnlyManifests are important to you: test. Offer feedback.
@nmcspadden @erikng

aschwanb:
Thanks for the feedback.

I tested with @tcinbis and can confirm that his matrix is correct. I can test other scenarios but I can't think of any at the moment.

gregneagle:
Pinging @natewalck @nmcspadden @pudquick. If you don't test this and I merge it and it breaks you, 🤷‍♀️

macmule:
Seems fine for me from a jamJAR POV (not that that should matter).

I'll literally ignore this for jamJAR as no need in that context

natewalck:
This looks perfectly reasonable to me. Looking at `the process_optional_install` code, it handles dupes, so there are no worries about having them specified in both places, etc.

pudquick:
I'll test this change out for our environment today, but I'm not expecting this merge to be an issue.

code/client/munkilib/updatecheck/core.py
```diff
@@ -174,6 +174,16 @@ def check(client_id='', localmanifestpath=None):
                     localonlymanifest, 'managed_uninstalls') or []
                 for item in localonlyuninstalls:
                     analyze.process_removal(item, cataloglist, installinfo)
+
+                localonlyoptionals = manifestutils.get_manifest_value_for_key(localonlymanifest,
+                                                                              'optional_installs') or []
```
clburlison: This line is too long and doesn't match the style of this file. Please move `localonlymanifest, 'optional_installs') or []` to the second line.

code/client/munkilib/updatecheck/core.py
```diff
@@ -174,6 +174,16 @@ def check(client_id='', localmanifestpath=None):
                     localonlymanifest, 'managed_uninstalls') or []
                 for item in localonlyuninstalls:
                     analyze.process_removal(item, cataloglist, installinfo)
```
clburlison: Can you fix this line while you're in here? This line doesn't match the style above it or the code you're adding.
```python
                    dummy_result = analyze.process_removal(
                        item,
                        cataloglist,
                        installinfo
                    )
```

clburlison:
I've tested this over here and while I don't have a use case maybe I will in the future. 👍 from me.

tcinbis:
@clburlison thanks for your feedback! I added your style suggestions in my last commit.

pudquick:
@gregneagle Tested here with our configuration. Doesn't cause any issues for our usage of local only manifests (which are all centered around managed installs and managed uninstalls).

Admittedly - while I understand the value of this change, most everything in our catalog that could potentially be an optional install already is - we generally don't restrict application availability to specific groups of users.

As such, I couldn't test that specific desired functionality of this change easily.

But the concern for us was more in regards to breaking our existing usage, which it doesn't seem to.

Looks good on this end 👍

gregneagle:
Merged!

gregneagle:
And now I regret pulling the trigger so quickly: this should have been merged to Munki3dev. I'll try to fix that up now.

gregneagle:
Merged to Munki3dev now.

tcinbis:
Thanks @gregneagle !

PR #838, opened 2018-02-07
Skip middleware import if path isn't a directory.

ccaviness:
If munkilib is embedded in certain self-contained python binary formats
(e.g. https://github.com/google/subpar), the path returned by
os.path.dirname(__file__) may not actually be an existing path.

To work around this, check if munki_dir is actually a directory before
importing middleware.

### DIFF
--- a/code/client/munkilib/fetch.py
+++ b/code/client/munkilib/fetch.py
@@ -76,6 +76,10 @@ def import_middleware():
     the module is loaded under the 'middleware' name'''
     required_function_name = 'process_request_options'
     munki_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
+    if not os.path.isdir(munki_dir):
+        display.display_detail(
+            'Ignoring %s because it is not a directory' % munki_dir)
+        return
     for filename in os.listdir(munki_dir):
         if (filename.startswith('middleware')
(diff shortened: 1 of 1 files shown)


ccaviness:
That sounds like work and not a hacky workaround. I'll see what I can do.

By repoplugins you mean https://github.com/munki/munki/blob/e05b5798b675581d386e6f7ee587ba69f637d4da/code/client/munkilib/munkirepo/__init__.py#L40 ? If so, then yeah, that might break the same way.

gregneagle:
Yep.
https://github.com/munki/munki/wiki/Repo-Plugins

https://github.com/munki/munki/wiki/Middleware

gregneagle:
Seems to me we should be looking at supporting an external path -- somewhere like /Library/Managed\ Installs/plugins -- that can be searched for these files. Munki should always look for these external paths, and they should be the preferred location for the plugins. We'll need to support the current locations for backwards compatibility for a while, though.  For your case, if you can detect when the code is packaged in this format, you'd skip the check for the path inside the munki directory, but still check in the external directory.

This all sounds like it should work, but I'm probably not understanding something important...

gregneagle:
Closing due to lack of interest/movement/need from the originator, and no new interest from anyone else...

PR #849, opened 2018-05-04
Allow for precaching of optional installs

aysiu:
As @nmcspadden suggested, leave the delivery of the precache up to an outside mechanism and not a Munki run, but this would allow the designated optional installs (with the `precache` = `True` flag in the pkginfo) to remain in the `/Library/Managed Installs/Cache` folder even if the optional install isn't selected.

### DIFF
--- a/code/client/munkilib/updatecheck/analyze.py
+++ b/code/client/munkilib/updatecheck/analyze.py
@@ -243,10 +243,12 @@ def process_optional_install(manifestitem, cataloglist, installinfo):
     iteminfo['uninstallable'] = (
         item_pl.get('uninstallable', False)
         and (item_pl.get('uninstall_method', '') != ''))
-    # If the item is a precache item, get that precache flag and also the basename of the installer item
+    # If the item is a precache item, get that precache flag
+    # and also the basename of the installer item
     if item_pl.get('installer_item_location') and item_pl.get('precache'):
         iteminfo['precache'] = True
-        iteminfo['installer_item'] = download.get_url_basename(item_pl['installer_item_location'])
+        iteminfo['installer_item'] = \
+            download.get_url_basename(item_pl['installer_item_location'])
     iteminfo['installer_item_size'] = \
         item_pl.get('installer_item_size', 0)
     iteminfo['installed_size'] = item_pl.get(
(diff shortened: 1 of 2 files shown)


code/client/munkilib/updatecheck/analyze.py
```diff
@@ -243,6 +243,10 @@ def process_optional_install(manifestitem, cataloglist, installinfo):
     iteminfo['uninstallable'] = (
         item_pl.get('uninstallable', False)
         and (item_pl.get('uninstall_method', '') != ''))
+    # If the item is a precache item, get that precache flag and also the basename of the installer item
+    if item_pl.get('installer_item_location') and item_pl.get('precache'):
+        iteminfo['precache'] = True
+        iteminfo['installer_item'] = download.get_url_basename(item_pl['installer_item_location'])
```
gregneagle: What is this used for?

gregneagle: Oh, I see -- used in building the list of items to not remove from the cache.

gregneagle: Make sure you don't break the coding style -- lines need to be under 80 characters unless it is impossible to fold them.

aysiu: Thanks. I've wrapped the text so it should be under 80 characters per line.

aysiu:
@gregneagle Signed, scanned, and emailed this past Friday.

aysiu:
I was looking for a good place to put the download of the precache if it's going to be part of a Munki run, and I was thinking as an `else` to this `if`:
https://github.com/munki/munki/blob/master/code/client/munkilib/updatecheck/analyze.py#L186

At that point, we know it's an optional install and that it's not already installed, so that might not be a bad place to check to see if it's marked as `precache` and then try to download the installer item.

gregneagle:
I disagree that there would be a good place to download precache items.

I have an unpublished branch with more precaching code which I can publish we I get confirmation of your CLA.

aysiu:
Sounds good, @gregneagle

gregneagle:
I've merged this into a new preaching branch here: https://github.com/munki/munki/commit/ceb3e347bbb52df2936d9d03cb66ddb5ac06de2c

and implemented the code to actually download to the cache in later commits to that branch.

PR #851, opened 2018-05-20
Add "postware"

jessepeterson:
"Postware" is like middleware but gets called when file downloads are completed.

### DIFF
--- a/code/client/munkilib/fetch.py
+++ b/code/client/munkilib/fetch.py
@@ -70,14 +70,13 @@ DEFAULT_USER_AGENT = "managedsoftwareupdate/%s Darwin/%s" % (
     munki_version, darwin_version)
 
 
-def import_middleware():
-    '''Check munki folder for a python file that starts with 'middleware'.
+def import_middleware(file_prefix='middleware', globals_prefix='middleware', required_function_name='process_request_options'):
+    '''Check munki folder for a python file that starts with a prefix.
     If the file exists and has a callable 'process_request_options' attribute,
-    the module is loaded under the 'middleware' name'''
-    required_function_name = 'process_request_options'
+    the module is loaded under the 'middleware' name by default'''
     munki_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
     for filename in os.listdir(munki_dir):
-        if (filename.startswith('middleware')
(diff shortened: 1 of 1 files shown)


jessepeterson:
Created largely in support of https://github.com/jessepeterson/signware

code/client/munkilib/fetch.py
```diff
@@ -70,14 +70,13 @@
     munki_version, darwin_version)
 
 
-def import_middleware():
-    '''Check munki folder for a python file that starts with 'middleware'.
+def import_middleware(file_prefix='middleware', globals_prefix='middleware', required_function_name='process_request_options'):
```
gregneagle: Run your changes through a PEP-8 checker; this line is definitely over 80 characters...

code/client/munkilib/fetch.py
```diff
@@ -305,10 +306,18 @@ def get_url(url, destinationpath,
...
+            # postware module must have process_post_fetch function
+            postware.process_post_fetch(url, destinationpath)
         return connection.headers
     elif connection.status == 304:
         # unchanged on server
         display.display_debug1('Item is unchanged on the server.')
+        if postware:
+            display.display_debug2('Processing options through postware')
+            # postware module must have process_post_fetch function
+            postware.process_post_fetch(url, destinationpath)
```
gregneagle: If the item is unchanged, the signature was checked during the initial download (unless you've just switched to checking signatures). Is there a good reason to check the signature again?

jessepeterson: I think the idea was that if the signature changed but the download itself hadn't, that we should re-run the postware (in the case of the signature-checking). Due to the fact we place it in a temp file the signature is always re-downloaded, however. Perhaps another postware function should be called (or a different argument/option supplied to the function to indicate the separate calls)? Or perhaps we should save the signaures along-side the cats/manifests? This way we could not download them over and over again.

I don't want the "postware" to be exclusively for signature checking - hopefully it's generic enough for other use cases, too.

gregneagle:
Closing this as there was really no progress forward, and now that the command-line tools are rewritten in Swift (and all plugins therefore must be dylibs that implement a protocol), this can never be merged without a total re-think.

PR #853, opened 2018-05-29
Fix optimistic timeout value 

alexander-riss:
if the timeout is hit it will not cause the **very** expensive call result and instead discard it. the higher value should ensure, the hang case is not as likely to be confused with a system with many installed apps. (the system_profiler call takes around 50 seconds on my 2015 macbook when nothing else is active)


gregneagle:
Years of experience has shown the choice in the current code works well enough for the issue it was designed to address. Bumping it to 10 minutes would be a different choice, with a different set of trade-offs, and not the choice I would make.

Issue #855, opened 2018-06-04
10.14 Munki Error 


f3xy:
I haven't seen one yet on this issue, but a fresh Install of 10.14 and Munki generates this issue when running managedsoftwareupdate. Currently investigating possible fixes, but latest munki tools didn't seem to work (3.3.1.3525)
```
  File "/usr/local/munki/managedsoftwareupdate", line 58, in <module>
    from munkilib import appleupdates
  File "/usr/local/munki/munkilib/appleupdates/__init__.py", line 1, in <module>
    from .core import *
  File "/usr/local/munki/munkilib/appleupdates/core.py", line 22, in <module>
    from . import au
  File "/usr/local/munki/munkilib/appleupdates/au.py", line 39, in <module>
    from . import sync
  File "/usr/local/munki/munkilib/appleupdates/sync.py", line 41, in <module>
    from .. import fetch
  File "/usr/local/munki/munkilib/fetch.py", line 45, in <module>
    from . import keychain
  File "/usr/local/munki/munkilib/keychain.py", line 30, in <module>
    from OpenSSL.crypto import load_certificate, FILETYPE_PEM
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/OpenSSL/__init__.py", line 41, in <module>
    from OpenSSL import rand, SSL
ImportError: dlopen(/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/OpenSSL/SSL.so, 2): Symbol not found: _SSLv3_method
  Referenced from: /System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/OpenSSL/SSL.so
  Expected in: flat namespace
 in /System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/OpenSSL/SSL.so
```

clburlison:
I think this issue should be re-opened. Python on Mojave, including https requests, do work. The issue is Apple is now shipping a newer version of LibreSSL (2.6.4) in 10.14 while 10.13 was linked against (2.2.7). During this upgrade, the maintainers of LibereSSL have removed support for SSL v3 which causes the above import error.

As far as I can see we only import from OpenSSL once in the `keychain.py` file and use the `load_certificate` function

https://github.com/munki/munki/blob/e05b5798b675581d386e6f7ee587ba69f637d4da/code/client/munkilib/keychain.py#L157-L159

[...]

PR #856, opened 2018-06-06
fix: Remove OpenSSL import in keychain.py

clburlison:
In Mojave 10.14 Apple has updated the openssl version to a newer release
of LibreSSL (2.6.4) that no longer contains the proper Symbols for SSLv3.
This causes import errors in the OpenSSL module that ships with 10.14.

To resolve the import errors we now shell out to openssl to get the
client cert common name. This has been hand tested with the following
macOS/openssl versions:

- 10.12 - 0.9.8zh 14 Jan 2016
- 10.13 - LibreSSL 2.2.7
- 10.14b1 - LibreSSL 2.6.4

Fix: #855

For testing, I verified with two certs with the following formating:

```
subject= /UID=RP82Y2QL76/CN=Developer ID Installer: Clayton Burlison (RP82Y2QL76)/OU=RP82Y2QL76/O=Clayton Burlison/C=US

and 

subject= /CN=DESKTOP-FB2JSN1
```

both properly report the correct common name. I haven't tested
this in the context of a munki run to valid client certs work on 10.14
but the function is simply replacing the OpenSSL library so it
should work.

[test_certs.zip](https://github.com/munki/munki/files/2074633/test_certs.zip)

### DIFF
--- a/code/client/munkilib/keychain.py
+++ b/code/client/munkilib/keychain.py
@@ -27,7 +27,6 @@ import base64
 import hashlib
 import os
 import subprocess
-from OpenSSL.crypto import load_certificate, FILETYPE_PEM
 
 from . import display
 from . import osutils
@@ -151,11 +150,16 @@ def get_client_cert_common_name():
     cert_info = get_munki_client_cert_info()
     client_cert_path = cert_info['client_cert_path']
     if client_cert_path and os.path.exists(client_cert_path):
-        fileobj = open(client_cert_path)
-        data = fileobj.read()
-        fileobj.close()
-        x509 = load_certificate(FILETYPE_PEM, data)
-        common_name = x509.get_subject().commonName
+        cmd = ['/usr/bin/openssl', 'x509', '-noout', '-subject', '-in',
+               client_cert_path]
+        proc = subprocess.Popen(cmd,
+                                bufsize=-1, stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE, stdin=subprocess.PIPE)
+        (out, err) = proc.communicate()
+        if out:
+            for i in out.split('/'):
+                if i.startswith('CN='):
+                    common_name = i[3:].rstrip()
     return common_name
 
 


code/client/munkilib/keychain.py
```diff
@@ -151,11 +150,16 @@ def get_client_cert_common_name():
...
+        cmd = ['/usr/bin/openssl', 'x509', '-noout', '-subject', '-in',
+               client_cert_path]
+        proc = subprocess.Popen(cmd,
+                                bufsize=1, stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE, stdin=subprocess.PIPE)
+        (out, err) = proc.communicate()
+        if out:
+            for i in out.split('/'):
+                if i.startswith('CN='):
+                    common_name = i.strip('CN=').strip()
```
gregneagle: Let's make this `i[3:]`. Otherwise you might be stripping the characters "=", "N" and/or "C" from the beginning and end of the CN:

```
foo = "CN=CartooN"
print foo.strip("CN=")
'artoo'
print foo[3:]
'CartooN'
```

code/client/munkilib/keychain.py
```diff
@@ -151,11 +150,16 @@ def get_client_cert_common_name():
...
     if client_cert_path and os.path.exists(client_cert_path):
-        fileobj = open(client_cert_path)
-        data = fileobj.read()
-        fileobj.close()
-        x509 = load_certificate(FILETYPE_PEM, data)
-        common_name = x509.get_subject().commonName
+        cmd = ['/usr/bin/openssl', 'x509', '-noout', '-subject', '-in',
+               client_cert_path]
+        proc = subprocess.Popen(cmd,
+                                bufsize=1, stdout=subprocess.PIPE,
```
gregneagle: Small speed-up if `bufsize=-1`. I don't always remember this...

code/client/munkilib/keychain.py
gregneagle: And maybe even limit the strip to rstrip:
`common_name = i[3:].rstrip()`

IOW, is it legal to have leading and/or trailing spaces in a CN?

clburlison: I'm not sure if it is legal to have leading/trailing spaces in a CN. I did a few quick searched but couldn't come up with anything on that topic.

clburlison:
Updated with your code comments.

gregneagle:
LGTM

PR #867, opened 2018-10-03
Remove .lower() from  check_mode() in pkginfolib

### LINKED ISSUE #866: makepkginfo --mode converts mode to lowercase
While trying to fix the permissions on an Application bundle (Opera) by adding the `--mode go+rX` option to the `mkpkginfo` command line tool, I noticed that the permissions are lowercase in the resulting .plist file.

Looking in the sources I noticed that there is a `lowercase()` statement in the `check_mode()` function

https://github.com/munki/munki/blob/master/code/client/munkilib/admin/pkginfolib.py#L713

This prevents the uppercase `X` being recorded correctly. I would send a PR to remedy this, but I can't oversee the consequences of removing the `lowercase()` invocation.

bochoven:
Fix #866

### DIFF
--- a/code/client/munkilib/admin/pkginfolib.py
+++ b/code/client/munkilib/admin/pkginfolib.py
@@ -710,7 +710,7 @@ def makepkginfo(installeritem, options):
 
 def check_mode(option, opt, value, parser):
     '''Callback to check --mode options'''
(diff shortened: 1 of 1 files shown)


gregneagle:
Looks good to me...

PR #871, opened 2018-10-18
Fixing makedirs umask problem

jeffbair:
os.makedirs honors the umask of the current process so it may write the directory with incorrect permissions which causes issues down the line when the directory is checked for permission of 700.  This fix is to make sure the umask is set to 0 and then reset the umask after the directory is created.

https://stackoverflow.com/questions/5231901/permission-problems-when-creating-a-dir-with-os-makedirs-in-python

### DIFF
--- a/code/client/munkilib/installer/dmg.py
+++ b/code/client/munkilib/installer/dmg.py
@@ -174,7 +174,9 @@ def validate_source_and_destination(mountpoint, item):
         # write access to the parent directory from writing their own payloads
         # to this directory
         try:
+            original_umask = os.umask(0)
             os.makedirs(full_destpath, 0o0700)
+            os.umask(original_umask)
         except OSError, err:
             display.display_error(
                 "Error creating %s: %s", full_destpath, err)


gregneagle:
Can you provide an example of the problem you are addressing? I'd think a umask that affected user permissions would cause so many other issues.  I can cause odd behavior if I set the umask to something crazy like 0o700 but so many other things are broken that fixing this one thing while in this state seems a bit silly.
Perhaps I'm missing something, though.

jeffbair:
Sure, once Managed Software Center is installed we make sure to run:
/usr/local/munki/managedsoftwareupdate", "--checkonly
/usr/local/munki/managedsoftwareupdate", "--installonly

in sequence in order to download the apps that are set up for auto-install on munki.  For some reason the directories (specifically /Applications/Skype.app and /Applications/VLC.app as examples) were being created with 600 permissions during the os.makedirs process.  I made sure of this by setting a debug line before and after to check that the file was not there before that call and checking the file exists and the permissions on that location after the call.  I checked the umask as well that was returned from os.umask(0) and it was 127 so I think that's why the execution bit for root was not set and thus created with 600.  Note that if you open Managed Software Center and hit update it was created properly but we're trying to do this in an automated fashion without the end user having to open the app for auto-install apps to get pushed down.

gregneagle:
So root's umask is set to remove the traversal bit from directories it creates? I'd think this would break lots of stuff.

jeffbair:
I see what you mean now, my process that is calling /usr/local/munki/managedsoftwareupdate has the umask set to 177 but we have our own file creation libraries that take this into account and we didn't see this issue until the insecure check from https://github.com/munki/munki/commit/9e0d4e0426d241f4068dcef45e8834996a85d045#diff-418ce77299f8ae2ae32b6badda49fdf3 .  I think the best thing to do is for us to set the umask properly before calling the command line interfaces so going to close this pull request.

PR #879, opened 2018-11-19
Copy optional key precache

aysiu:
Needed to not delete precache OnDemand items after installation

### DIFF
--- a/code/client/munkilib/updatecheck/analyze.py
+++ b/code/client/munkilib/updatecheck/analyze.py
@@ -487,7 +487,8 @@ def process_install(manifestitem, cataloglist, installinfo,
                              'icon_name',
                              'PayloadIdentifier',
                              'icon_hash',
-                             'OnDemand']
+                             'OnDemand',
+                             'precache']
 
             if (is_optional_install and
                     not installationstate.some_version_installed(item_pl)):


gregneagle:
We don't want to _always_ copy the precache key; this change seems to to just that.

aysiu:
Is there another place to get that key into InstallInfo.plist? That seemed to be the place, and it seems to copy only if the key exists?
https://github.com/munki/munki/blob/Munki3dev/code/client/munkilib/updatecheck/analyze.py#L498-L500

Am I reading that wrong? Is there a better place to get the `precache` key in?

gregneagle:
You are basically throwing away this logic:

https://github.com/munki/munki/blob/Munki3dev/code/client/munkilib/updatecheck/analyze.py#L246-L257

aysiu:
I mean, you obviously know the code better than I do, but when I didn't have this PR in, this condition evaluated to false when it should be true:
https://github.com/munki/munki/pull/878/commits/9b3e9ebb3a751fd8b2d910fe0602c85027e9f1c5

Isn't there a difference between
```
process_optional_install(manifestitem, cataloglist, installinfo):
    """Process an optional install item to see if it should be added to
    the list of optional installs.
    """
```
and
```
process_install(manifestitem, cataloglist, installinfo,
                    is_managed_update=False,
                    is_optional_install=False):
    """Processes a manifest item for install. Determines if it needs to be
    installed, and if so, if any items it is dependent on need to
    be installed first.  Installation detail is added to
    installinfo['managed_installs']
    Calls itself recursively as it processes dependencies.
    Returns a boolean; when processing dependencies, a false return
    will stop the installation of a dependent item
    """
```
?

gregneagle:
Yes, I think you are right. I wish your two PRs were a single PR...

aysiu:
Yeah, that would have been cleaner. Sorry about that!

PR #898, opened 2019-01-10
adds line to remove the app status package receipt

tbridge:
This PR also removes the app_status package receipt during uninstall.

PR #899, opened 2019-01-10
Munki uninstall script—forget receipt for app_usage

aysiu:
Adds a line to forget the receipt for the app_usage pkg


gregneagle:
See #898.

PR #907, opened 2019-02-07
Various typo fixes and spelling corrections

homebysix:
Found and fixed various spelling issues across the code base.

Separate commits used for the larger changes, but feel free to "squash" to a single commit if you prefer.


gregneagle:
Thank you. It may take me a while to review 84 files with changes. If the changes were just limited to typos in _comments_ I'd move faster but fixing typos in localized files can actually break localization if not done properly...

code/apps/munki-notifier/munki-notifier/da.lproj/InfoPlist.strings
```diff
@@ -1,3 +1,3 @@
-/* (No Commment) */
```
gregneagle: This is interesting because something (one of the tools that generates localized files) is actually generating that typo. We can fix it, but it will most likely return.

homebysix:
I took extra care with the .strings files, and double checked to make sure all referenced strings were changed together. Hopefully your review will confirm that. 🤞

code/apps/munki-notifier/munki-notifier/da.lproj/InfoPlist.strings
homebysix: Yes, I see `(No Commment)` elsewhere on GitHub too. I'll try to find the typo upstream.

code/apps/Managed Software Center/Managed Software Center/Controllers/MSCAlertController.swift
```diff
@@ -161,7 +161,7 @@ class MSCAlertController: NSObject {
         } else if updatesRequireLogout() || installRequiresLogout() {
             let alert = NSAlert()
             alert.messageText = NSLocalizedString(
-                "Logout Required", comment: "Logput Required title")
```
gregneagle: This should then roll down to the Localizable strings. I see it in Base.lproj/Localizable.strings and en.lproj/Localizable.strings

homebysix: I don't see any results for `grep -R "Logput" .` in the repo on this branch. Am I missing something?

gregneagle: https://github.com/munki/munki/blob/Munki3dev/code/apps/Managed%20Software%20Center/Managed%20Software%20Center/Base.lproj/Localizable.strings#L107
https://github.com/munki/munki/blob/Munki3dev/code/apps/Managed%20Software%20Center/Managed%20Software%20Center/en.lproj/Localizable.strings#L107

gregneagle: ...and now I think I understand why grep doesn't find those instances: Xcode strings files are UTF-16 encoded text files; grep expects ASCII. Many other files are UTF-8 encoded, which is much much closer to ASCII (much code is encoded exactly the same as in an ASCII file).

gregneagle: Horrible hack:
```# grep -R "L.o.g.p.u.t." .
Binary file ./code/apps/Managed Software Center/Managed Software Center/en.lproj/Localizable.strings matches
Binary file ./code/apps/Managed Software Center/Managed Software Center/Base.lproj/Localizable.strings matches```

This "works" because the ASCII character set is encoded as two bytes; the second byte is 00; but we just use '.' to match on any character. One _could_ be more specific with

```# grep -R "L\x00o\x00g\x00p\x00u\x00t\x00" .
Binary file ./code/apps/Managed Software Center/Managed Software Center/en.lproj/Localizable.strings matches
Binary file ./code/apps/Managed Software Center/Managed Software Center/Base.lproj/Localizable.strings matches```

but that's a lot harder and more tedious to type.

homebysix: Great catch. I fixed those two instances in 4256580, and I'll go back and look for other changes in the UTF-16 files now.

homebysix: Using Sublime (which I verified successfully searches binary/utf-16), I went back and searched for more instances of every correction I made. I caught and fixed a few more typos in *.py files, but no more issues were found in the strings. 👍

PR #910, opened 2019-02-12
fix shebangs

mparamonau:
Do the same as for other scripts.

### DIFF
--- a/code/client/makecatalogs
+++ b/code/client/makecatalogs
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python
 # encoding: utf-8
 #
 # Copyright 2009-2018 Greg Neagle.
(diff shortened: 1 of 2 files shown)


arubdesu:
Why is this being submitted? Is there a conversation somewhere else we should be aware of? Shebangs are very deliberate.

mparamonau:
I do apologize if I've broken the repo changes workflow. The only thing that I wanted is to help someone to avoid potential issues I faced with. If these 2 trivial changes demand to spend a lot of time for hard discussing, I would close the PR for that. I know that - first discuss and go further, if so, please close it for discussion...

gregneagle:
These are intentionally the way they are. Rather than attempting to change them without discussion, it would be better to start with a discussion on why you think they should be changed.

PR #913, opened 2019-02-27
Handle empty icon_name strings

jamesez:
The Python version would default to the package's name if icon_name was missing or empty. This restores the empty string behavior.


gregneagle:
Thanks - I've already addressed this in yet unpushed code.

jamesez:
boo / yay

gregneagle:
https://github.com/munki/munki/commit/7ddf98055d5977765cc08456d827cdcd574ee7b4

PR #933, opened 2019-05-03
Align except clauses with PEP 3110

homebysix:
This should work for Python 2.6+, including 3.
https://www.python.org/dev/peps/pep-3110/#compatibility-issues

### DIFF
--- a/code/client/iconimporter
+++ b/code/client/iconimporter
@@ -47,7 +47,7 @@ def copy_icon_to_repo(repo, name, path):
     try:
         repo.put_from_local_file(icon_ref, path)
         print_utf8(u'\tWrote: %s' % icon_ref)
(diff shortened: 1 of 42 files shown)


gregneagle:
While I appreciate the thought and effort here, I wish there'd been a heads-up... It's _possible_ I have an unpublished branch where I've been working on Python 3 compatibility...

homebysix:
Understood — these changes were easy for me, so I'd be happy to close and revisit once the Python3 branch is published.

gregneagle:
No, it's fine. I might even reopen and merge this as the Python 3 stuff is just too large. It probably is better to slowly move the code toward Python 3 style and idioms, while still keeping it Python2-compatible for now. It would be nice to have a plan such that people could help do this enormous task.

PR #935, opened 2019-06-03
add colorized outputs for makecatalogs

mindovermiles262:
Adds colorized terminal output in `makecatalogs`

Screen_Shot_2019-06-03_at_11_54_34

Screen Shot 2019-06-03 at 12 02 15

Screen Shot 2019-06-03 at 12 02 25

Screen_Shot_2019-06-03_at_12_04_11

Screen_Shot_2019-06-03_at_12_12_49

PR #937, opened 2019-06-04
Munki3dev

mindovermiles262:
See https://github.com/munki/munki/pull/935

Adds colorized outputs to `makecatalogs` terminal output

### DIFF
--- a/code/client/munkilib/admin/makecatalogslib.py
+++ b/code/client/munkilib/admin/makecatalogslib.py
@@ -30,6 +30,7 @@ from .common import list_items_of_kind, AttributeDict
 from ..colors import colorize
 from .. import munkirepo
 from ..wrappers import readPlistFromString, writePlistToString
+from ..colorize import colorize
 
 
 class MakeCatalogsError(Exception):
@@ -37,10 +38,12 @@ class MakeCatalogsError(Exception):
     pass
 
 
-def hash_icons(repo, output_fn=None):
+def hash_icons(repo, options, output_fn=None):
     '''Builds a dictionary containing hashes for all our repo icons'''
     errors = []
     icons = {}
+    color = options.color
+    ENDC = color['end']
     if output_fn:
         output_fn("Getting list of icons...")
     icon_list = repo.itemlist('icons')
@@ -50,7 +53,7 @@ def hash_icons(repo, output_fn=None):
     for icon_ref in icon_list:
         if output_fn:
             output_fn(
-                colorize.OKGREEN + "Hashing " + colorize.ENDC +
+                color['ok_green'] + "Hashing " + ENDC +
                 "%s..." % (icon_ref))
         # Try to read the icon file
         try:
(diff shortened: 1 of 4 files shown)


code/client/munkilib/admin/makecatalogslib.py
```diff
@@ -50,7 +49,9 @@ def hash_icons(repo, output_fn=None):
         icon_list.remove('_icon_hashes.plist')
     for icon_ref in icon_list:
         if output_fn:
-            output_fn("Hashing %s..." % (icon_ref))
+            output_fn(
+                colorize.OKGREEN + "Hashing " + colorize.ENDC +
+                "%s..." % (icon_ref))
```
gregneagle: If output_fn isn't printing to stdout -- like if it's appending to a log, or sending info to Managed Software Center, what would be the effect of adding these colorization codes?

gregneagle:
See also https://github.com/munki/munki/wiki/Contributing-to-Munki#contributor-license-agreement

mindovermiles262:
Currently the color codes would be included as text to the output/log. 

Would only including colorized output if output_fn=None be the way to solve this?

gregneagle:
Maybe, but that adds a lot of if/then complexity to simple output lines.

gregneagle:
Maybe you should consider _not_ being so aggressive with your colors and simply wrap the print fn with one that looks for lines starting with "WARNING" and "ERROR" (or whatever) and colorize those.

I get what you are trying to do here, but not yet convinced this won't trigger other issues.

mindovermiles262:
Alternatively I can make colorize a function that checks if output_fn is STDOUT and adds the color codes if such

gregneagle:
What if STDOUT is a pipe to another command and is not going to the terminal?
I'm pretty ignorant about all this.

mindovermiles262:
STDOUT piped will include the color codes regardless of how it’s implemented. 

If you want to avoid this I think the best way would be to add a `—no-color` flag and change the colorize class to a function that checks for this flag

gregneagle:
I think a concern I have is "What if people have scripts that are parsing the output of makecatalogs?" (to post things to Slack or generate emails, or other things) MWA2 parses the output. I worry that the additional of colorization codes could break these. So -- maybe my concerns are alleviated if colorized output is opt-in: a `--colorized-output` flag, perhaps, or an environment variable, or even a preference read from somewhere.

gregneagle:
And I see we've hit upon the same basic idea, though I'd want it to behave the same as now by default.

mindovermiles262:
Agree and understand. 

Let me take a deeper look to find the best place to add an opt-in variable and I’ll update this PR

gregneagle:
I think some UNIX-y tools attempt to detect if STDOUT is connected to a tty-like object and only colorize the output if so, but I have zero idea how to do that.

gregneagle:
This looks promising:
https://stackoverflow.com/questions/1077113/how-do-i-detect-whether-sys-stdout-is-attached-to-terminal-or-not

which suggests
```python
if sys.stdout.isatty():
    # You're running in a real terminal
else:
    # You're being piped or redirected
```

gregneagle:
In any case, if the colorization is not-opt-in, we'll need to carefully communicate the change so people can test to make sure this doesn't break any tooling they've wrapped around `makecatalogs` output. I mostly worry about `if output.startswith("XXX"):`-style parsing.

mindovermiles262:
I think setting an option/setting is the best way to go about this. That way nothing breaks but the feature is still available for those that want it

PR #945, opened 2019-07-31
Fixed repoclean sorting by LooseVersion in reverse mode

LcTrKiD:
Tested with python 3.7.3 and system python in 10.14.5

### DIFF
--- a/code/client/repoclean
+++ b/code/client/repoclean
@@ -120,13 +120,15 @@ class RepoCleaner(object):
             manifests_list = self.repo.itemlist('manifests')
         except munkirepo.RepoError as err:
             self.errors.append(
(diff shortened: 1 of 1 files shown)


gregneagle:
Did you also test in Python 2.7?

gregneagle:
There are a lot of unrelated code-formatting changes in the PR, which makes review take longer.

LcTrKiD:
Yes, I used pipenv to generate an env with python 3.7.3 to test Python3 and /usr/bin/python to test Python2. Let me remove those changes and I'll make another PR

gregneagle:
The formatting changes are fine, but probably should be a separate commit.

gregneagle:
Thanks! I've made this error a few other times when converting from Python 2 to Python 3 and replacing the sorting custom functions...

PR #955, opened 2019-09-28
CI on GitHub Actions!

timsutton:
This PR contains everything that should be needed to do automated package builds and uploads to munkibuilds.org, using GitHub Actions.

The only change added to `make_munki_mpkg.sh` is the addition of a `-v` flag that will show all the output from `xcodebuild`.

I might need to add one more follow-up to make it still run successfully on a PR from a fork, as this PR will be the first time this is run this way.


natewalck:
<3

weswhet:
😍😍😍

PR #967, opened 2019-12-26
Added capabilities to run script on Catalina: Defaults Read removed

jonbrown21:


gregneagle:
Could you please describe the issue you encountered and your solution?

gregneagle:
Since there has been no description of the problem that is being addressed here, and since you _can_ use `defaults` to get the version in Catalina:

```
bash-3.2# sw_vers
ProductName:	Mac OS X
ProductVersion:	10.15.2
BuildVersion:	19C57
bash-3.2# defaults read /Users/Shared/munki-master/code/client/munkilib/version CFBundleShortVersionString
4.0.1
```

...I intend to close this PR in the next day or so unless persuaded otherwise.

gregneagle:
(I rather suspect the issue here is around Catalina's new restrictions on reading things in a user's Desktop, Documents and Downloads folders. The fix is to move your munki repo outside of those folders, or to grant Terminal rights to read those folders.)

PR #983, opened 2020-03-06
When checking for com.apple.quarantine xattr when copying from dmg, do not follow symlink

chrisgrande:
When checking for com.apple.quarantine if the xattr.list() hits a symlink it tries to follow it - raising a _Error removing com.apple.quarantine..._ warning.

Setting XATTR_NOFOLLOW on list() to stop behavior.


gregneagle:
Can you re-submit this against Munki4dev instead of master?

chrisgrande:
Should be all set against Munki4Dev

PR #984, opened 2020-03-06
Exclude "Managed Installs" from backups on install

### LINKED ISSUE #118: Excluding updates from backups
_From [hjuutila...@mac.com](https://code.google.com/u/104810762682981732871/) on October 03, 2011 00:25:16_

This is just a minor enhancement request...

The contents of /Library/Managed Installs/ should probably be excluded from time machine backups. Apple is already excluding /Library/Updates/ by default (in /System/Library/CoreServices/backupd.bundle/Contents/Resources/StdExclusions.plist) and I think it would be nice for munki to automatically mark files as excluded.

I'm not sure how to implement this in python but in cocoa the way to exclude a path is defined in BackupCore.h and found in CoreServices framework. For example:

NSURL *someURL = [NSURL fileURLWithPath:@"/path/to/file"];
Boolean exclude = TRUE; // exclude item from backups
Boolean excludeByPath = FALSE; // this item is excluded regardless of its location
CSBackupSetItemExcluded((CFURLRef)someURL, exclude, excludeByPath);

Is there something in /Library/Managed Installs/ that needs a backup or should we just exclude the whole directory?

BTW. CrashPlan respects this exclusion too.
## 

Hannes

_Original issue: http://code.google.com/p/munki/issues/detail?id=118_

Artoria2e5:
This commit closes #118 by sticking the xattr on the directory and telling pkgbuild to keep it. This might not be the perfect way, as it has a slight chance of putting `com.apple.quarantine` in [where you don't want it](https://carlashley.com/2018/11/03/preserving-xattr-with-pkgbuild/), but it does seem like the easiest approach without anything needed at run-time.

### DIFF
--- a/code/tools/make_munki_mpkg.sh
+++ b/code/tools/make_munki_mpkg.sh
@@ -757,7 +757,7 @@ for pkg in core admin app launchd app_usage python no_python; do
             --info "$PKGTMP/info_$pkg" \
             --component-plist "${PKGTMP}/munki_${pkg}_component.plist" \
             --scripts "$SCRIPTS" \
-	    --preserve-xattr \
+            --preserve-xattr \
             "$PKGDEST/munkitools_$pkg-$ver.pkg"
     else
         sudo /usr/bin/pkgbuild \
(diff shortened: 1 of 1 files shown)


gregneagle:
While this seems reasonable, it _is_ an opinionated _choice_.

Might it not be better to just let admins do this themselves if they so choose?

Has anyone documented which backup products actually respect this extended attribute?

Artoria2e5:
According to [this feature request](https://forum.duplicacy.com/t/honoring-com-apple-backup-excludeitem-on-macos/979), CrashPlan and Arq do.

PR #986, opened 2020-04-07
Munki including Python Notarization and signing

lifeunexpected:
Adds the feature of 
- signing python framework and making it ready for notarization later
- Runs build munkitools.pkg script to create a pkg ready for signing and notarization
- Automatically staples and notarize the finished munkitools.pkg

### DIFF
--- a/code/tools/MunkiPythonNotarizeAndSign.sh
+++ b/code/tools/MunkiPythonNotarizeAndSign.sh
@@ -30,9 +30,9 @@ DevApp="Developer ID Application: Name/Company (ID)"
 DevInst="Developer ID Installer: Name/Company (ID)"
 # Change Bundle_ID if you are using a custom one, default is "com.googlecode.munki"
 BUNDLE_ID="com.googlecode.munki"
-# Apple Developer app-specific password
+# Apple Developer account e-mail
 AppleAcc="DeveloperAppleAcc@Apple.com"
-# https://support.apple.com/en-us/HT204397
+# Apple Developer app-specific password https://support.apple.com/en-us/HT204397
 AppleAccPwd="Apple Developer Account app-specific password"
 
 # Stop changing!
--- a/code/tools/MunkiPythonNotarizeAndSign.sh
+++ b/code/tools/MunkiPythonNotarizeAndSign.sh
@@ -1,0 +1,105 @@
+#!/bin/bash
+# encoding: utf-8
+
+# Scripts are based on code by:
+# https://groups.google.com/forum/#!topic/munki-dev/FADUXPWJeds - Michal Moravec
+# https://github.com/rednoah/notarize-app/blob/master/notarize-app - rednoah
+# https://github.com/munki/munki/tree/master/code/tools - Greg Neagle
+
+# 1: Copy script to Munki folder 
+# 2: In terminal "cd FolderWheremunki" git repo is located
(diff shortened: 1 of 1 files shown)


gregneagle:
This is very nice work.

I have not yet decided if I want to accept the responsibility for maintaining and supporting this set of tasks and this script. I don't need to sign and notarize these for my own deployment. It might be better to just point to this tool someplace _you_ are willing to support and maintain it.

MichalMMac:
Munki build scripts already already have support for code signing binaries and the package. Maybe it would be better to split this script into two parts. 1. Add Python.framework code signing to Munki build scripts 2. Create separate tool for notarizing the whole package.

lifeunexpected:
Yes the original munki build script do support code signing of the binaries and the package but not for the python framework that was added in munki v4 and it does not notarize and staple the finished pkg file you need to do that later yourself.

If you do try to notarize the pkg file generated with the original munki build script apple will send you an error e-mail back saying its not approved for notarization because python framework wasnt signed.

The script above in the commit adds signing of the python framework, then builds a pkg file using the original script and then uploads the finished pkg file to apple for notarization and stappling so its approved by Apple.

Who knows maybe in macOS 10.16.x or later they will demand that everything is signed and notarized for the pkg file to be installed so this solves that problem for now until apple or the python framework is changed again.

I do understand Greg that you might not want to undertake the work to support it.
I have it uploaded to my own github page so i will continue supporting in the future becuase i will need to have the software signed and notarize for my own usage at work.

natewalck:
If this is something you would support, I'm happy to provide the signing keys to do it. 

Another question is whether we can store the keys in a secure way and sign it using Github actions/builds or not. We'll be figuring this out for the Python packages, so once that is sorted, happy to share our lessons learned if this is something you don't mind supporting on this repo too.

gregneagle:
At the current time this is not something I care to support.

natewalck:
Fair enough 👍

arubdesu:
Either the reasons notarization fail have changed since the python version was bumped, or trying this process on Big Sur is showing different un-addressed issues, it coughs on ~20 dylibs and .so's with the error message `The signature of the binary is invalid`.
Just to drop some data points into the conversation, while python.org's 3.8.5 mac pkg is notarized, saltstack 3001.1 is not. I don't know if re-signing the files this script is meant to operate on to get them to pass notarization is actually a good idea, I'd somewhat prefer to leave the embedded python out which is not currently an option in make_munki_mpkg (I understand it's a component I can repackage to remove).

lifeunexpected:
I fixed the 20 dylibs and .so error messages just now and the update is on my github.

If re-signing and signing the python files is the best solution or not im not really sure about but at least it solves the problem for now with getting the software signed and notarized with Apple.

lifeunexpected:
Just added a new script that makes it easier to notarize and sign a specific build.
I wasnt sure if i should add something regarding this into the Wiki maybe or not so more people could use it.
https://github.com/lifeunexpected/Scripts

natewalck:
Is this something that could be landed as an example but not actually called as part of the build process? I think it would be useful if people *do* want to sign/notarize their Munki build.

lifeunexpected:
Added the information regarding signing and notarization to the wiki here
https://github.com/munki/munki/wiki/Signing-Munki

PR #987, opened 2020-04-08
changes to avoid crash while creating icon in macOS 10.15.4 and above

trishiva:
Added not empty check to avoid crash while creating icon due to CGImageSourceCopyPropertiesAtIndex faliure, in macOS 10.15.4 and above.

### DIFF
--- a/code/client/munkilib/iconutils.py
+++ b/code/client/munkilib/iconutils.py
@@ -70,6 +70,9 @@ def convertIconToPNG(icon_path, destination_path,
         try:
             properties = CGImageSourceCopyPropertiesAtIndex(
                 image_source, index, None)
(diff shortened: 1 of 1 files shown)


gregneagle:
Any ideas why it sometimes fails in 10.15.4 and above, and have you filed a bug with Apple?

trishiva:
> Any ideas why it sometimes fails in 10.15.4 and above, and have you filed a bug with Apple?

After some troubleshooting, I found that the Apple API is working as expected. The API extracts properties of an .icns file provided to it, it returns nil in case it isn't able to extract .
The issue is with the packages for which the API isn't able to extract some of the .icns file. I checked those  .icns of the pkgs for which API was returning nil, it seems the .icns aren't supported in 10.15.4 and onwards.

PR #997, opened 2020-06-20
munki: rename "/usr/local/munki/python" symlink to "munki-python"

### LINKED ISSUE #996: sug: install symlinked python as munki-python
With munki installed, it is possible to get this result because `/usr/local/munki` is listed in `/etc/paths.d/munki`:

```
$ which python
/usr/local/munki/python
```

Most user may not notice, but this would be possible to remedy by renaming the symlink to `munki-python` and adjusting the scripts accordingly.

A naive `git grep /usr/local/munki/python` in the repo finds 19 occurrences, so this seems like it would be straightforward to change.

Since this looks really quick to fix, I'll put together a PR for review.  Thanks for Munki!

davvid:
Avoid masking the system /usr/bin/python by calling our symlink
"munki-python" instead of "python".

Closes #996

### DIFF
--- a/code/client/app_usage_monitor
+++ b/code/client/app_usage_monitor
@@ -1,4 +1,4 @@
-#!/usr/local/munki/python
+#!/usr/local/munki/munki-python
 # encoding: utf-8
 #
 # Copyright 2017-2020 Greg Neagle.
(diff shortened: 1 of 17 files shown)


davvid:
Meta: this PR was generated using:

```
git grep -l /usr/local/munki/python  |
xargs perl -p -i -e 's{/usr/local/munki/python}{/usr/local/munki/munki-python}g'
```

So it may have missed something if it wasn't found by the grep.

PR #1000, opened 2020-06-26
Loop control variables to start with _

alexgorbachev:
Just minor code style update following up on Flake8 linting.
Added underscores to the beginning of the loop variables names not used inside the loops.


gregneagle:
Thanks for the PR.

The linter I'm using (a TextMate 2 plugin that calls PyCheckMate 1.2 / Pylint 1.9.5) doesn't flag these as needed changes.
If I accept linter-related changes from people using different linters/different linter configurations, then I have to test that those changes don't in turn trigger _my_ linter configuration to complain. And if so, then try to figure out how to make both linters happy or how to adjust my linter configuration, etc etc.

I'm just not interested in that type of work, sorry!

I do think it would be useful to define and document the linting standards used by this project, but that's not something I have a lot of motivation to do right now.

Again, thanks for the effort here.

alexgorbachev:
Thanks for consideration.

PR #1005, opened 2020-07-09
Automatic Munki 5 Dev packages

erikng:
This PR will automatically create a new pre-release package every time a new commit is pushed to the `Munki5dev` branch.

This standard could also be used to automatically generate official releases.

### DIFF
--- a/.github/workflows/build_munki5dev.yml
+++ b/.github/workflows/build_munki5dev.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: macos-latest
 
     steps:
-    - name: Checkout python repo
+    - name: Checkout munki repo
       uses: actions/checkout@v2
 
     - name: Set environment variables
@@ -21,7 +21,9 @@ jobs:
          echo ::set-env name=BUILD_DATE_RELEASE::$(/bin/date -u "+%B %d %Y %H:%M:%S")
 
     - name: Run build package script
-      run: /usr/bin/sudo ./code/tools/make_munki_mpkg.sh -o ./outputs -p
+      run: |
+          /bin/mkdir -p ./outputs
+          /usr/bin/sudo ./code/tools/make_munki_mpkg.sh -o ./outputs -p
 
     - name: Create Release
       id: create_release
--- a/.github/workflows/build_munki5dev.yml
+++ b/.github/workflows/build_munki5dev.yml
@@ -1,10 +1,9 @@
 name: Build and upload recommended package
 
 on:
-  pull_request:
   push:
     branches:
-    - munki5devactions
+    - Munki5dev
 
 jobs:
   build:
--- a/.github/workflows/build_munki5dev.yml
+++ b/.github/workflows/build_munki5dev.yml
@@ -1,0 +1,54 @@
+name: Build and upload recommended package
+
+on:
+  pull_request:
+  push:
+    branches:
+    - munki5devactions
+
+jobs:
+  build:
+    runs-on: macos-latest
+
+    steps:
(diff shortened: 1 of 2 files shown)


erikng:
You can see the successful job and relevant logs from my own branch when I was testing this: https://github.com/erikng/munki/actions/runs/163910107

gregneagle:
Something's not quite right, as the versioning is off on the generated pkg and tag. It should be something like munkitools-5.1.0.4061.pkg (or whatever you get from `./code/tools/make_munki_pkg_from_git.sh -b Munki5dev`)

erikng:
I noticed this as well, but thought it might be the branch.

From the runner logs. Any idea what could be causing that?
```
Done!
Customized, relocatable framework is at /Users/runner/work/munki/munki/Python.framework
Build variables

  Bundle ID: com.googlecode.munki
  Munki root: /Users/runner/work/munki/munki
  Output directory: ./outputs
  munki core tools version: 5.1.0.483
  LaunchAgents/LaunchDaemons version: 3.0.483
  Apps package version: 5.2.0.483
  Python version: 3.8.3

  metapackage version: 5.1.0.483
```

erikng:
Thinking through this, it seems like the `defaults read` might be returning a `1` in the runner, vs outright failing and then being combined with `MAGICNUMBER` which would result in `483`.

gregneagle:
I think it's more likely the checkout action checks out a shallow clone, so there's one revision. 1 plus MAGICNUMBER is 483.
This would result in the final number always being 483.

gregneagle:
I'll need help understanding how the secrets work.

erikng:
I have a job queued that I'm hoping will solve the issue: https://github.com/erikng/munki/actions/runs/163945264

As far as the secrets, these are just environmental variables given to you by the runners: https://docs.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token

This will automatically look up your private tokens and use that as an authentication method to post the release. There are custom secrets you can use, but not for those particular actions. Custom secrets could be used, for example, to sign the package.

erikng:
https://github.com/erikng/munki/runs/856244835

```
Build variables

  Bundle ID: com.googlecode.munki
  Munki root: /Users/runner/work/munki/munki
  Output directory: ./outputs
  munki core tools version: 5.1.0.4061
  LaunchAgents/LaunchDaemons version: 3.0.3265
  Apps package version: 5.2.0.4056
  Python version: 3.8.3

  metapackage version: 5.1.0.4061
```

Looking good now: https://github.com/erikng/munki/releases/tag/v.5.1.0.4061

One thing that I was thinking about, but perhaps the python package should also follow this format? What if you had deployed Python 3.8.3, and then needed another update to PyObjC? As currently designed, this would be an issue.

I don't think solving it in this diff makes sense, but just food for thought.

gregneagle:
I think you are probably right about Python component pkg versioning.

I am worried about the Releases page being buried in an avalanche of Munki5dev autobuilds, making it difficult for people to find the latest official release and greatly increasing the odds people will install an autobuild on their fleet thinking it's a supported release (something that happened far too often with munkibuilds.org). Any ideas on that?

erikng:
I noticed that on the macadmins python the releases are sorted alphabetically.

All of your releases start with "Munki VERSION". My initial thought would say "Munki Development VERSION" which should put it lower, but that would likely mean it would be at the very last page.

I could be wrong here though and it's only alphabetical after the date. That said there is a "prerelease" notice I put on all of these releases.

erikng:
:( offer still stands if you want this.

PR #1012, opened 2020-08-18
Support repo-based makecatalogs

tboyko:
This PR allows a Munki repo plugin to define its own makecatalogs behavior via a `makecatalogs()` method. If not defined, Munki will use the default behavior of `makecatalogslib.makecatalogs()`, as before.

### DIFF
--- a/code/client/makecatalogs
+++ b/code/client/makecatalogs
@@ -96,7 +96,7 @@ def main():
 
     # Make the catalogs
     try:
-        errors = repo.makecatalogs()
+        errors = repo.makecatalogs(options, output_fn=print)
     except AttributeError:
         errors = makecatalogslib.makecatalogs(repo, options, output_fn=print)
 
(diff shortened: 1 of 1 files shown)


code/client/makecatalogs
```diff
@@ -95,7 +95,10 @@ def main():
         exit(-1)
 
     # Make the catalogs
-    errors = makecatalogslib.makecatalogs(repo, options, output_fn=print)
+    try:
+        errors = repo.makecatalogs()
```
gregneagle, line 99: Do we want to pass along the same parameters to `repo.makecatalogs()`? The implementation would be free to ignore them, but perhaps they might be useful.

tboyko: Good point. For my uses, they aren't needed, but make sense for some future proofing. Updating...

tboyko:
Thanks Greg 👍

gregneagle:
I reverted this commit in branch main, then cherry-picked it into branch Munki5dev. It will be in the next Munki 5.1 beta release.

PR #1025, opened 2020-09-16
Add cleanup scripts

grahamgilbert:
This PR will:

- Run any script in `/usr/local/munki/cleanup` at the start of a `managedsoftwareupdate` run.
- If the script is successful, `managedsoftwareupdate` will remove the script.
- Currently it will not do anything if the script fails, just leaves it there for the next run.
    - Should it halt the run? My gut says no.
- Includes a cleanup script for Python 3.7.


natewalck:
I like it. This allows for more flexibility in the future in case we need additional cleanup we hadn't considered.

code/client/managedsoftwareupdate
```diff
@@ -218,6 +218,24 @@ def runScript(script, display_name, runtype):
...
+    """
+    If there are executables inside the cleanup directory, run them and remove them if successful
+    """
+    cleanupdir = os.path.join(os.path.realpath(os.path.dirname(sys.argv[0])), 'cleanup')
+    if not os.path.isdir(cleanupdir):
+        # path doesn't exist or isn't a directory
+        return
+
+    # get all of the files in the cleanup directory
+    (_, _, filenames) = next(os.walk(cleanupdir))
```
gregneagle, line 231: what's the advantage of this over the simpler `filenames = os.listdir(cleanupdir)` ?

code/tools/make_munki_mpkg.sh
```diff
@@ -570,6 +570,10 @@ chmod -R 755 "$PYTHONROOT/usr"
 cp -R "$MUNKIROOT/Python.framework" "$PYTHONROOT/usr/local/munki/"
 # Create symlink
 ln -s Python.framework/Versions/Current/bin/python3 "$PYTHONROOT/usr/local/munki/munki-python"
+
+# copy in cleanup scripts
+rsync -a --exclude '*.pyc' --exclude '.DS_Store' "$MUNKIROOT/code/tools/pkgresources/cleanup_scripts/" "$PYTHONROOT/usr/local/munki/cleanup/"
+
```
gregneagle, line 576: Why `rsync` instead of a simple `cp`?
Also - are we putting the installation of the cleanup scripts in the python package because these particular cleanups have to do with python? if so, how do we generalize for possible future cleanup scripts for other components?

code/tools/pkgresources/cleanup_scripts/cleanup.sh
```diff
@@ -0,0 +1,11 @@
...
+
+set -e
+
+if [ -f "/usr/local/munki/python" ]; then
+    /bin/rm /usr/local/munki/python
+fi
+
+if [ -d "/usr/local/munki/Python.framework/Versions/3.7" ]; then
+    /bin/rm -r /usr/local/munki/Python.framework/Versions/3.7
+fi
```
gregneagle, line 11: Probably should add the trailing newline.

code/client/managedsoftwareupdate
grahamgilbert: None, it's just what I'm used to using. Not married to either if you'd like it to change.

code/tools/make_munki_mpkg.sh
grahamgilbert: I was just going on what was used elsewhere in the build script. A cp -R would work fine I guess. 

I originally put it in the core package, but moved it out. We could add yet another package that contains the cleanup scripts, move it to the core package or leave it where it is.

code/client/managedsoftwareupdate
gregneagle: Yeah, since I'm usually the one on the hook for maintaining the code after it gets merged, I'd rather it be the simpler version so I'm not confused when I look at it in a year. :-)

code/tools/make_munki_mpkg.sh
gregneagle: I guess I'm thinking more along the lines of -- should we be doing something like:

```shell
if [ -d "$MUNKIROOT/code/tools/pkgresources/$COMPONENTNAME_cleanup_scripts/" ] ; then 
    rsync -a --exclude '*.pyc' --exclude '.DS_Store' "$MUNKIROOT/code/tools/pkgresources/$COMPONENTNAME_cleanup_scripts/" "$PYTHONROOT/usr/local/munki/cleanup/"
fi
```

for each component package? Of course that could be a future enhancement.

grahamgilbert: I can do that

code/client/managedsoftwareupdate
grahamgilbert: Oh, the difference is `(_, _, filenames) = next(os.walk(cleanupdir))` only produces a list of files, `filenames = os.listdir(cleanupdir)` is files and directories.

gregneagle: OK. Keep it as is, but add a comment for future me:
`# get only filenames (and not directory names) from cleanupdir`

gregneagle:
I don't see your name on the approved contributors list...
https://github.com/munki/munki/wiki/Contributing-to-Munki#contributor-license-agreement

grahamgilbert:
Has it really been that long since I've submitted a PR? CLA sent via email.

grahamgilbert:
Got confirmation from WDAS, should be good to go on the CLA front.

PR #1027, opened 2020-09-21
Adding managed_updates support for local manifest

NoahRJ:


gregneagle:
Could you please search munki-discuss and munki-dev for discussions around when this feature (local-only manifests) was added to see if there was any thinking around what would be supported and what wouldn't, and why?

I don't use local-only manifests myself and am unlikely to any time soon.

Therefore: any changes to the functionality ideally will be vetted by people who do use the functionality, and changes must not open security vulnerabilities (which is why some things are not supported in the local-only manifest).

In any case, this should be targeted at Munki5dev instead of main, and will almost certainly not be merged before the 5.1 release.

gregneagle:
See also https://github.com/munki/munki/wiki/Contributing-to-Munki

erikng:
It seems safe to me. @natewalck @nmcspadden

NoahRJ:
Hey @gregneagle - thanks for pointing me to that Wiki page. I've signed the contributor agreement and sent that over.

I've been through all the discussions I could find relating to local-only manifests and their supported features. No mention of managed_updates in the [original PR](https://github.com/munki/munki/pull/520), in the [munki-dev conversation](https://groups.google.com/g/munki-dev/c/lgnBUFyMfAo) that originated the idea, nor in the [most recent local-only PR](https://github.com/munki/munki/pull/837) to add optional_installs.

Happy to have some other eyes on it, though, especially for those folks that heavily lean on local manifests in their workflows. I didn't run into any issues in my test scenarios, and I'm happy to walk through those if desired. The function `analyze.process_managed_update` looks to have similar duplicate validation as `analyze.process_optional_install` that was merged a couple years back, and I didn't see any problems testing with `managed_updates` specified in multiple manifests.

nmcspadden:
I don't have any thematic objection to this. I think it really came down to a few things:
* At the time, we only _needed_ the specific functionality of adding managed installs/removals.
* We didn't want to recreate the entire manifest logic locally, because at the time the intent was not to eliminate server-side manifests (and it still isn't), so more work didn't confer any benefit
* Managed installs and removals were generally the simplest thing, because there's typically nothing fancy about them. Users generally can't interact with them, and they're typically invisible unless you happen to notice it in the MSC Updates tab.

If @gregneagle is willing to accommodate, I'm certainly happy to see more functionality added to local manifests (i.e. they could be a full equivalent to "server" manifests in terms of specifying the behavior of software installs), although it's unlikely we'll use this particular feature.

gregneagle:
Haven't forgotten this. I'll merge this at some point after the 5.2 release.

PR #1037, opened 2020-10-09
Add support for building an autorun pkg to initiate an --auto run

LcTrKiD:


code/tools/make_munki_mpkg.sh
```diff
@@ -490,7 +496,7 @@ chmod -R go-w "$APPROOT/Applications/Managed Software Center.app"
 # sign MSC app
 if [ "$APPSIGNINGCERT" != "" ]; then
     echo "Signing Managed Software Center.app..."
-    /usr/bin/codesign -f -s "$APPSIGNINGCERT" --options runtime --verbose \
```
gregneagle, line 493: Why would we need to --force a signature on a newly-built app? In any case this would be unrelated to adding a new option to do an auto run. Please revert this.

code/tools/pkgresources/autorun_launchd_script/autorun.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+/bin/rm /Users/Shared/.com.googlecode.munki.autorunmanagedsoftwarecenter
```
gregneagle, line 3: What the heck is this referring to?

launchd/autorun_LaunchDaemon/com.googlecode.munki.autorun.plist
```diff
@@ -0,0 +1,18 @@
...
+		<key>PathState</key>
+		<dict>
+			<key>/Users/Shared/.com.googlecode.munki.autorun</key>
+			<true/>
+		</dict>
+	</dict>
+	<key>Label</key>
+	<string>com.googlecode.munki.autorun</string>
+	<key>Program</key>
+	<string>/tmp/autorun.sh</string>
```
gregneagle, line 16: This is really odd. You are installing a permanent launchdaemon that calls a script that will no longer be present after the next reboot. So after a reboot, we have a _broken_ or useless LaunchDaemon.

1) Why the need for a script? Why not just call the tool(s) directly?
2) If launchd is the right choice here, perhaps instead of installing a permanent launchd job for a one-time operation, maybe make use of `launchctl submit`?

gregneagle (changes requested):
Is creating a new, permanent launchdaemon job the right way to do this?
I think you should open a discussion on munki-dev and get some feedback on preferred approaches. This implementation is problematic as-is.

code/tools/pkgresources/autorun_launchd_script/autorun.sh
LcTrKiD: Ups sorry, I left things I used to test this in my local env

launchd/autorun_LaunchDaemon/com.googlecode.munki.autorun.plist
LcTrKiD: Ok, I will take a look at that option

LcTrKiD: I’m using a LaunchDaemons because I’m running managedsoftwareupdate and deleting the flag file

PR #1042, opened 2020-10-20
Update for html and css colors

lifeunexpected:
- Updated colors to make light mode and dark mode more like app store
- made colors and background into one fluent design

New design:

pictures of the updated design going to update some comparising pictures soon
Screenshot 2020-10-20 at 13 25 13
Screenshot 2020-10-20 at 13 24 58
Screenshot 2020-10-20 at 13 24 34
Screenshot 2020-10-20 at 13 24 12

Old design:
Screenshot 2020-10-20 at 13 53 45
Screenshot 2020-10-20 at 13 53 28
Screenshot 2020-10-20 at 13 52 56
Screenshot 2020-10-20 at 13 52 35

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center/Resources/WebResources/base.css
+++ b/code/apps/Managed Software Center/Managed Software Center/Resources/WebResources/base.css
@@ -13,48 +13,49 @@ html {
     --text-color-subdued: hsl(240, 2.7%, 55.7%);
     --text-color-warning: hsl(0, 100%, 40%);
     --description-link-color: #3E6397;
-    --button-text-color-enabled: hsl(0, 0%, 100%);
-    --button-text-color-active: hsl(0, 0%, 100%);
+    --button-text-color-enabled: hsl(211, 100%, 50%);
+    --button-text-color-active: hsl(105, 0%, 47%);
     --button-text-color-disabled: hsl(0, 0%, 74%);
-    --small-button-text-color-enabled: hsl(212, 100%, 51.4%);
-    --small-button-text-color-active: hsl(212, 95.4%, 42.7%);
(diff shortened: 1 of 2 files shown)


gregneagle:
While I appreciate the attempt, I don't agree with all the choices. I think perhaps this should turn into a discussion somewhere more people will see it -- the munki-dev list seems a good choice for that.

gregneagle:
Some more specific reactions follow.

You title this PR "Update for html and css colors to match macOS Big Sur" yet some of the changes don't look like the Big Sur app store at all. (100% matching the App Store appearance is not a real goal here. But once we decide to go a different direction, it becomes all a matter of opinion.)

For example, you've removed the different background color behind the "More by" and "More in" sections. The Big Sur (and Catalina) app store has a different background, as does MSC.app in 5.2 Betas 1 and 2.

image

gregneagle:
In your screenshots, I like the visual changes to the former "sidebar" area that is now a horizontal strip below the showcase images. But -- how does it look with larger lists, or lists containing pop-up menus?

lifeunexpected:
Here is a picture where you have a pop-up menu and a larger list. 

with scrollable menu and multiple text boxes

gregneagle:
With the more complex "sidebar" links, your redesign is less attractive than it is with a small number of links. I think the current design holds up better in this case.

Without a obvious "oh, this is better" reaction on my part (which I don't currently have) I can't see why I'd make the changes you propose. The changes are just... different. Another take. They're fine, but I think so is the current appearance.

gregneagle:
I'm going to close this. If you keep working on it and have a better take/design, feel free to open a new PR. But I think it's even better to start by posting screenshots in munki-dev and starting a discussion about any proposed changes (and why you think they are an improvement).

Issue #1022, opened 2020-09-04
new-msc-design: layout issues on Big Sur


gregneagle:
![Image 9-4-20 at 8 07 AM](https://user-images.githubusercontent.com/700560/92254486-d0d67c80-ee85-11ea-81d8-176473c2fe18.jpg)

![Image 9-4-20 at 8 09 AM](https://user-images.githubusercontent.com/700560/92254554-ecda1e00-ee85-11ea-9bf4-ac29e86e16a3.jpg)

as seen in the screenshots, on Big Sur (Beta 5) the content view has a large border on the top and left.

gregneagle:
New clue: it doesn't always do this. Sometimes when launched it lays out the view "normally". When it does this, the titlebar area is no longer draggable. I had it do the "wrong" layout a second time -- there were notification center notifications displayed, and the web view was pushed over enough to not be under those notifications. Once I dismissed the notifications, I could quit and relaunch MSC.app and the layout was "normal" again. 

I don't know how notification center notifications can interfere with an app's window layout, but it is looking like that is what is happening.

gregneagle:
I can only reproduce it when waking from sleep or otherwise clearing a screen lock, if notification center notifications are displaying. I've now seen the behavior on Beta 5 and Beta 6.

gregneagle:
I've now seen the behavior on beta 9. Once again, it happened when launching MSC.app shortly after waking the machine from sleep with notification center notifications displaying. The web view appears to "avoid" displaying under the displayed notifications.

flammable:
I saw similar behavior on beta 9 - I ended up rebuilding my client customizations with the current templates (I hadn't touched them for a few years), and that improved things. Might have been a coincidence; I'll need to test more.

However, I'm also seeing our custom banner cut off at the right edge. Should it be shrinking, instead? The banner is the recommended 1158x200. I'm testing on a MacBook Pro (Retina, 13-inch, Early 2015), which explains why I'm only able to make MSC display two columns of icons instead of three.

Here's a screenshot:

<img width="1280" alt="Screen Shot 2020-10-12 at 1 48 49 PM" src="https://user-images.githubusercontent.com/8499030/95786657-056bfe00-0ca6-11eb-96f8-ec46c1f7d335.png">

gregneagle:
> However, I'm also seeing our custom banner cut off at the right edge. Should it be shrinking, instead? The banner is the recommended 1158x200.

[...]

PR #1047, opened 2020-11-07
Directly create layout constraints

jamesez:
Fixes issue #1022 where the autoresizing masks occasionally end up pinning the webView to its smallest size.

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
+++ b/code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
@@ -494,12 +494,42 @@ class MainWindowController: NSWindowController, NSWindowDelegate, WKNavigationDe
             }
             // init our webview
             let replacementWebView = MSCWebView(frame: webViewPlaceholder.frame, configuration: webConfiguration)
-            replacementWebView.autoresizingMask = webViewPlaceholder.autoresizingMask
+            replacementWebView.translatesAutoresizingMaskIntoConstraints = false  // we'll add them later, by hand
             replacementWebView.allowsBackForwardNavigationGestures = false
             replacementWebView.setValue(false, forKey: "drawsBackground")
             // replace the placeholder in the window view with the real webview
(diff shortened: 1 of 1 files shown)


gregneagle:
I'll freely admit I do not understand layout constraints at all, but this looks reasonable...

gregneagle:
One new issue with this change is that using the Web Inspector to debug the HTML/CSS/Javascript parts of MSC is now partially broken. To see this, `defaults write com.googlecode.munki.ManagedSoftwareCenter developerExtrasEnabled -bool YES`
Then control-click anywhere in the webview and choose "Inspect Element".

jamesez:
Okay, I'll look at that tomorrow; out of brain after cranking on this for the last 5+ hours.

gregneagle:
No worries. Your help so far is greatly appreciated.

jamesez:
Yuck! I honestly never used the inspector there before today, so I didn't check.

I tinkered around with a few dead ends this morning, but the thing that seems to be working is to have a view controller manage the webview; that way, when the inspector tries to jam itself into the view hierarchy everything lines up correctly. I want to make sure I didn't introduce a race condition against a nil view before I'm going to push the patch.

Back at it after breakfast.

gregneagle:
I wonder if everything would be simpler if the top-level view was an NSSplitView, with the sidebar and webview as children of the NSSplitView.

gregneagle:
After some quick experiments, I think using NSSplitView seems promising. I can even go back to using the autoresizing mask and not have to manually create layout restraints.

gregneagle:
Even easier/better, since we don't need to resize the two views, seems to be to just have a superview contain the web view. Currently, the webview's superview is the window view; instead it should be a custom NSView with the correct position and autoresize mask...

gregneagle:
This branch: https://github.com/munki/munki/tree/msc-layout-test resolves the issues with the web inspector, and I have not yet seen the layout issue on Big Sur. Please give it a whirl.

jamesez:
I've updated this branch to build out a View Controller, instead, if you want to try it.

PR #1048, opened 2020-11-07
Moves the WebView to a View Controller - fixes Big Sur display bug and handles inspector correctly

jamesez:
This PR moves the WebKit view to a View Controller


jamesez:
For your consideration, instead of using a SplitView

gregneagle:
Since I decided to go with the fix that required fewer code changes i'm going to close this one without merging.

PR #1053, opened 2020-11-12
Implemented repoclean flag to preserve packages without a pkginfo

jc0b:
This change implements the `--delete-items-in-no-manifests` flag in repoclean. 
If the flag is set, orphaned packages are labelled with the `[to be DELETED]` label, and then deleted upon confirmation.
If the flag is not set, the orphaned packages are still displayed, but no action is taken on them.


gregneagle:
I'm a little confused here. `self.orphaned_pkgs` is a list of packages that do not have a corresponding pkginfo item.
It is not a list of items that are not in any manifest. So you are deleting some packages, but they aren't what you are saying they are...

jc0b:
Sorry, I misconstrued manifests and pkginfo. I'd still like to submit this as a feature request for repoclean (as we keep things on the munki server that don't have pkgsinfo, and would like to not delete them with repocleans), but I'll change this to a draft for now. Feel free to delete this PR if you'd rather I submit it some other way.

PR #1056, opened 2020-11-18
Add rosetta2 option to build script

poundbangbash:
Added an option (-R) to include a payloadless package that runs `software update --install-rosetta --agree-to-license` on ARM detected hardware prior to installing any Munki tooling.

While I don't have an M1 machine to test on, but Steve Yuroff was kind enough to test the MDM delivered package during enrollment to confirm bootstrapping worked as expected.

More testing is encouraged.


code/tools/make_munki_mpkg.sh
```diff
@@ -759,9 +794,10 @@ cat > "$DISTFILE" <<EOF
             <os-version min="10.10"/>
         </allowed-os-versions>
     </volume-check>
-    <options customize="allow" allow-external-scripts="no"/>
+    <options hostArchitectures="x86_64,arm64" customize="allow" allow-external-scripts="no"/>
```
gregneagle, line 797: Hmm. I wonder if this should be x86_64 unless we have the -R option...

poundbangbash: I wasn't sure what to do there either, but considering it's more of an opt-in type declaration I thought it'd be ok.

gregneagle: My thinking is that if the package _doesn't_ have the Rosetta pkg and declares hostArchitectures="x86_64,arm64", installer will happily install it without Rosetta present. Now what happens on Apple Silicon when `managedsoftwareupdate` attempts to run via launchd?

poundbangbash: As an example, Apple's tool "MacEvaluationUtility" has a Distribution that includes all 3 architectures. 
Screen Shot 2020-11-18 at 3 37 28 PM

gregneagle: That's interesting but doesn't really address my concern here.

poundbangbash: The architecture in the Distribution is to remove the act of installing software being a trigger for a Rosetta warning. If Rosetta is needed, but was not included by the admin creating the package, it's on them to address. 
Adding arm64 as a supported architecture makes sure this pkg doesn't trigger a warning/failure.

gregneagle: Seems like it would be straightforward to just make the `hostArchitectures="x86_64,arm64"` bit a variable like other things in the Distribution "template".

gregneagle: ...ala ROSETTA2OUTLINE, ROSETTA2CHOICE, and ROSETTA2REF.

poundbangbash: Yeah, updated the PR with changes. 
I built a pkg with and without `-R` and the hostArchitecture value reflected appropriately.

gregneagle:
We'll need to revisit this once a Universal2 build of Python is available.

PR #1057, opened 2020-11-20
Keep content-length on network connection error so downloads can resume

### LINKED ISSUE #817: gurl cannot resume from CloudFront (maybe other CDNs too) downloads
I'd like someone like @clburlison or @weswhet to confirm this, but I originally saw this behavior when writing the Imagr RAMDisk functionality, but chalked it up to the gurl being a fork.

In my testing yesterday of the MSC pending/problem updates issue, I realized that gurl wasn't resuming from CloudFront URL's, even though CloudFront supports resuming.

kcrawford:
Gurl requires the expected content length in the xattr data in order to
consider resuming a partial download.

When there is a network connection interruption
URLSession_task_didCompleteWithError_ is called with an error.

Prior to this fix content length was always removed from the xattr data
when URLSession_task_didCompleteWithError_ was called so downloads would
never resume after a connection error.

With this fix, content length is only removed from the xattr data when
there is no error since the download would be complete and resuming
would not be necessary.

Potentially fixes #817

### DIFF
--- a/code/client/munkilib/gurl.py
+++ b/code/client/munkilib/gurl.py
@@ -383,9 +383,10 @@ class Gurl(NSObject):
         '''NSURLSessionTaskDelegate method.'''
         if self.destination and self.destination_path:
             self.destination.close()
-            self.removeExpectedSizeFromStoredHeaders()
         if error:
             self.recordError_(error)
+        else:
+            self.removeExpectedSizeFromStoredHeaders()
         self.done = True
 
     def connection_didFailWithError_(self, _connection, error):


erikng:
Wow! As I was reading this I was thinking to myself "I think this will fix that CloudFront issue" and then I see you link it.

I am very excited to play with this.

Thanks,
Erik Gomez

gregneagle:
I hope Erik and others do take a look at this. This looks like a simple, safe fix, with little chance of unwanted side-effects. But I don't think I should slip it into Munki 5.2, which has already had three release candidates (two more than I'd wanted).

I'll leave it here until after the Munki 5.2 release and then merge it (barring any negative feedback in the meantime).

gregneagle:
@erik have you tested?

erikng:
No sorry. Took the week off.

PR #1074, opened 2021-02-24
Change false to False in prefs.py

KWik:
line 310 had value = false which of course fails ;)

### DIFF
--- a/code/client/munkilib/prefs.py
+++ b/code/client/munkilib/prefs.py
@@ -307,7 +307,7 @@ def print_config():
             # skip it
             continue
         if pref_name in FORCE_FALSE_ON_APPLE_SILICON:
-            value = false
+            value = False
             where = "Forced as False on Apple Silicon"
         else:
             value = pref(pref_name)


gregneagle:
Thanks for catching this. I caught it and fixed it in the universal2 branch here: https://github.com/munki/munki/commit/f2f1504a6f15ed30cc48368b2e64e55b63eb1f0c

But failed to back-port the fix to Munki5dev and main:

```
bash-3.2$ git diff universal2 main -- code/client/munkilib/prefs.py
diff --git a/code/client/munkilib/prefs.py b/code/client/munkilib/prefs.py
index d13fae80..3b5ad9a2 100644
--- a/code/client/munkilib/prefs.py
+++ b/code/client/munkilib/prefs.py
@@ -307,7 +307,7 @@ def print_config():
             # skip it
             continue
         if pref_name in FORCE_FALSE_ON_APPLE_SILICON:
-            value = False
+            value = false
             where = "Forced as False on Apple Silicon"
         else:
             value = pref(pref_name)

bash-3.2$ git diff universal2 Munki5dev -- code/client/munkilib/prefs.py
diff --git a/code/client/munkilib/prefs.py b/code/client/munkilib/prefs.py
index d13fae80..3b5ad9a2 100644
--- a/code/client/munkilib/prefs.py
+++ b/code/client/munkilib/prefs.py
@@ -307,7 +307,7 @@ def print_config():
             # skip it
             continue
         if pref_name in FORCE_FALSE_ON_APPLE_SILICON:
-            value = False
+            value = false
             where = "Forced as False on Apple Silicon"
         else:
             value = pref(pref_name)
```

PR #1077, opened 2021-02-28
Use keychain certificates for mTLS

np5:
This adds the functionality of looking for a client certificate in the Keychain, and use it if required.

The idea is to copy the behavior of the [MOLAuthenticatingURLSession](https://github.com/google/macops-MOLAuthenticatingURLSession/blob/ff9214adfbb3a70d6d59f40b957c6143a09df365/Source/MOLAuthenticatingURLSession/MOLAuthenticatingURLSession.m#L203), used for example in Santa.

Only the lookup based on the server response has been implemented. But it should be easy to add some configuration keys, to look for a certificate based on its Common Name or its Issuer Common Name for example, if the server response doesn't contain this information.

This is only a proof-of-concept. It relies on an extra dependency [asn1crypto](https://github.com/wbond/asn1crypto).

What do you think?

### DIFF
--- a/code/client/munkilib/gurl.py
+++ b/code/client/munkilib/gurl.py
@@ -61,8 +61,40 @@ from Foundation import (NSBundle, NSRunLoop, NSData, NSDate,
                         NSPropertyListMutableContainersAndLeaves,
                         NSPropertyListXMLFormat_v1_0)
 
+from Security import (errSecSuccess,
+                      kCFBooleanTrue,
+                      kSecClass, kSecClassIdentity,
+                      kSecMatchLimit, kSecMatchLimitAll,
+                      kSecReturnRef,
+                      SecCertificateCopyData, SecIdentityCopyCertificate,
+                      SecIdentityGetTypeID,
+                      SecItemCopyMatching)
+
+from asn1crypto.x509 import Certificate, Name
+
+
+# patch the credentialWithIdentity:certificates:persistence: signature
+# see https://github.com/ronaldoussoren/pyobjc/issues/320#issuecomment-784278944
+import objc
+objc.registerCFSignature(
+    "SecIdentityRef",
+    b"^{__SecIdentity=}", SecIdentityGetTypeID()
+)
+objc.registerMetaDataForSelector(
+    b'NSURLCredential',
+    b'credentialWithIdentity:certificates:persistence:',
+    {'arguments': {
+       2: {'null_accepted': False, 'type': b'@'},
+       3: {'_template': True, 'type': b'@'},
+       4: {'_template': True, 'type': b'Q'},
+     },
+     'classmethod': True,
+     'hidden': False,
+     'retval': {'_template': True, 'type': b"@"}}
+)
+
 try:
-    from Foundation import NSURLSession, NSURLSessionConfiguration
+    from Foundation import NSURLSession, NSURLSessionConfiguration, NSURLCredentialPersistenceForSession
     from CFNetwork import (kCFNetworkProxiesHTTPSEnable,
                            kCFNetworkProxiesHTTPEnable)
     NSURLSESSION_AVAILABLE = True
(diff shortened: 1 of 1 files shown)


arubdesu:
This is great! Especially for folks currently leveraging client certs that would like to transition to MDM-delivered, thanks!
I'd think the biggest barrier to consideration is the extra module dependency, there is only precedent for 'pluggable' interfaces that admins are expected to adhoc distribute modules for, as documented on the wiki: [Customizing Python for Munki](https://github.com/munki/munki/wiki/Customizing-Python-for-Munki).
I'd think at a minimum you'd be asked to wrap the import with a try and gracefully fail with feedback that the module isn't present. 
As per the comment I understand there's an unresolved issue upstream in pyobjc ([320](https://github.com/ronaldoussoren/pyobjc/issues/320)) code for which is included in this PR at present as a workaround. 
Just to add more context/color to folks who may be interested in what this functionality can unlock, there's also [MacSesh](https://github.com/sheagcraig/MacSesh) as an example of another project leveraging certs delivered to the keychain.

np5:
asn1crypto is only really needed to parse the issuer name from the server response. It is used again to parse the certificates and compare their issuer with the expected one, but only because it was easier this way. With a configuration key, we could set the Common Name of the certificate issuer for example, and use SecCertificateCopyValues to find the certificate in the keychain, like MacSesh is doing. We would loose the automation though, but it could be also that the web server is configured to not send this info.

gregneagle:
What problem is this designed to solve? I get that it's another way to provide client certs, but what does this make possible that can't be done with Munki's current support for client certs?

I hesitate on this because I really have no way to test or support this -- if I were to merge it and someone later were to tell me it was broken in some way there would be precious little I could do about it.

Convince me it's worth the risk here --  that it solves a real problem in a way that justifies the additional complexity and the risk that it might be hard for me to support.

gregneagle:
"Not everyone can deploy certs on disk."

Does this mean not everyone is _allowed_ to by organizational policy?

np5:
There is also the possibility of managing certificates with the SCEP payload, making sure that the private key never leaves the device. This is probably required in some organizations.

gregneagle:
OK. I'm going to request that @np5 find a third-party that will test this functionality. I can't test it, and I cannot tell if it breaks existing client cert functionality. We need someone who is currently using Munki client certs as local files to try this code out.

np5:
Would it help if I put together a python test server, with some extra functionality to setup a CA and build some mobileconfigs to install the root CA cert and the client certs.?

gregneagle:
We just need a way for me or a third-party to verify it works and doesn't break any existing functionality.

np5:
Then I will build it, but I need some time.

apettinen:
We can help testing this out with @hjuutilainen, as it would be a very welcome feature for our purposes. Let's come up with a plan together @np5

hjuutilainen:
Yep, I'm definitely interested in seeing this possibility in Munki. Will help with testing.

andre4ik3:
This PR looks like it will solve my problem. Long story short, my Munki server is behind Cloudflare's API Shield, which doesn't send a request for a client certificate, but rejects any traffic without one. Thank you so much for making this.

Will build and test right now.

rrenstrom:
Has anyone tried this with Big Sur 11.3 (since that seems to break Munki's current client certificate logic, resulting in the client cert not being sent to the server)?

arubdesu:
> Has anyone tried this with Big Sur 11.3 (since that seems to break Munki's current client certificate logic, resulting in the client cert not being sent to the server)?

Please open a new issue if you're seeing a regression when using non-mTLS certs.

rrenstrom:
I've started a post on munki-dev to discuss solutions for the issue with macOS 11.3 and Munki client certs. I'm hoping this PR solves the problem.

dcoobs:
This fixed our macOS 11.3 client certificate authentication issues as well.

gregneagle:
It's great that circumstances have conspired to force people to test this new code, and that it appears to solve issues introduced by macOS 11.3.

One tester also reports that the changes work back to High Sierra: https://groups.google.com/g/munki-dev/c/NnTkQeoM7Pk/m/91DGLoTfEgAJ

Some remaining concerns before I were to merge this:

1. @np5 says "This is only a proof-of-concept." To me that implies he doesn't think the code as-is is ready for real-world deployment. What needs to happen to make it so?
2. Maintenance: I cannot promise to be able to maintain this code. If Apple changes cause it to break again, it may remain broken until someone from the community digs into to fix it...

rrenstrom:
Q1 @np5 can address his thoughts on why POC, but in reviewing the code it appears to be doing the right things, and it works!

Q2 seems to me has already been addressed to some extent by the community digging in and testing the PR. And, in reviewing the PR and relevant Apple Developer Documentation, I certainly understand more about NSURLSession than I did a few days ago. ;-)

Q3 from asn1crypto project description, "A fast, pure Python library for parsing and serializing ASN.1 structures."

np5:
Q1: because it only works if the server is sending the expected Issuer info (hence the need for asn1crypto to read it), and the plan was to add an option, if you all agree, to specify the Issuer info. It would help if the server is not configured to send the information. There is also the issue about the extra dependency. Would you be open to include it ?

gregneagle:
Since Munki includes its own copy of Python, adding the asn1crypto dependency is not a big deal, as long as it pip installs cleanly and reliably.

When Munki was using Apple's python, _any_ additional dependency was essentially a show-stopper.

rrenstrom:
@np5 I think an additional Munki pref to specify the cert issuer info would be nice, but I wouldn't want it to hold up this PR, since the current implementation solves the immediate issue with clients certs no longer working _at all_ with Munki under macOS 11.3.  

Perhaps this key can be added later, since in the testing so far, by a few of us admins, it's working as designed with the server providing the issuer info.

arubdesu:
I think addressing point 1, it could help to consider having flags added for configuration/pref keys to ease setup/use, along the same lines as client certs, but more specifically how you want the lookup in the keychain/how fields in the cert are handled. Integrating this code to just restore previous client cert functionality seems right for now and we can kick that conversation down the road.
For the OP's main mTLS implementation, I'd imagine deciding a popular-enough 'standard implementation' to target would be helpful to make the more forward-looking decisions for the mTLS-	specific keys to add/flesh out with the probable options for security (by which I mean threat model) and flexibility. Cloudflare's product/service seems most worthwhile to mock up, but I haven't researched/don't understand enough to guide that conversation at the moment. Duo, Google, AWS, and Hashicorp Vault products (alongside traditional LDAP-complementing PKI/Microsoft NDES/ADCS) all seem like they have offerings in this space worth tracking. 
For point 2, I do, for some reason, blindly assume mTLS is where all agent-based communication will migrate, so I have every reason to believe the community will track and maintain it once available. Santa was the gold standard in spearheading this, and after munki adopts this, either via a community extension/'wrapper' or upstream/core change to osquery should be a clean shot. Again, blindly, I believe it will pay down the complexity cost of implementing/maintaining mTLS infra if it's reusable across agents. (In a 'brighter future', as another related benefit, this eases the implementation of Beyond Corp-ish services like Duo tends to be associated with.)

I can look into point 3 this week, if nobody beats me to it, seems very actionable.

np5:
OK. What should I do to prepare it for a release?

For the next steps, in a follow-up PR:
- Remove the [code for setting the identity preference](https://github.com/munki/munki/blob/9ab1e5fb795eb0a50bf05388c66ec4f0fc642b4c/code/client/munkilib/keychain.py#L307-L362)? It is probably the thing that doesn't work anymore in 11.3, and this PR makes it redundant.
- What about `ClientCertificateIssuerCN` for the preference key, to help select the certificate if the server is not sending the info?

gregneagle:
I don't know what needs to be done to prepare for a release since I'm not using or testing this code and/or this functionality.

If it can be merged without breaking existing Munki users, while also fixing a subset that were broken by the release of 11.3, that would be enough for the immediate need. The functionality can then be improved upon to help Munki admins that have servers that don't send the expected Issuer info.

As for future functionality to address servers that don't send the expected Issuer info -- why not specify the exact identity to use? Why use something less specific/less direct like `ClientCertificateIssuerCN`?

np5:
Because variables can be used in the SCEP payload to set machine specific information in the certificate request, and/or multiple certificates could be issued for the same machine using the same Subject.

In order to cover all the cases, we should have 3 modes: 
- auto selection with server info (this PR)
- `ClientCertificateIssuerCN`
- `ClientCertificateCN`

arubdesu:
asn1crypto on pip is pure python, https://github.com/wbond/asn1crypto

Issue #991, opened 2020-05-14
Feature Request: Allow repoclean to skip pkg items


homebysix:
It would be very useful to be able to remove pkginfo items only, and skip pkg items. A flag like `--skip-pkgs` might work well for this.

Together with the recently added `--auto` feature, this would make `repoclean` much more usable within CI/CD automation.

gregneagle:
Help me understand the use case.

homebysix:
We have our pkgsinfo and manifests in a GitLab repository. Our pkgs are synced separately, outside of GitLab.

We'd like to be able to use `repoclean` to remove pkginfo files from the Git repo that are older than the last N versions. We've been doing that manually for a while, and we have verified that it's safe to automate for us.

However, because `repoclean` in automatic mode also attempts to remove pkgs, which will not be present within our GitLab CI/CD environment, it exits nonzero and the job fails. Allowing pkgs to be skipped would solve that for us.

gregneagle:
OK.

Not something I have time/motivation to work on, but I'd look at a PR.

PR #1084, opened 2021-03-31
adding --skip-pkgs flag

midsandhighs:
This PR adds a `--skip-pkgs` option to the **repoclean** command, as a solution to [Issue#991](https://github.com/munki/munki/issues/991).

It adds an `if` statement to the `find_cleanup_items` function of the `RepoCleaner` class to loop through and add the _pkgpath_ and _uninstallpkgpath_ keys in **pkginfodb** to the `self.pkgs_to_keep` set while leaving the associated .pkginfo items in the `self.items_to_delete` list. 

Any feedback is welcome and appreciated.

### DIFF
--- a/code/client/repoclean
+++ b/code/client/repoclean
@@ -331,6 +331,15 @@ class RepoCleaner(object):
                     for item in item_list:
                         self.items_to_delete.append(item)
                         line_info = "[to be DELETED]"
+                if self.options.skip_pkgs:
+                    for item in item_list:
+                        if item['pkg_path']:
+                            self.pkgs_to_keep.add(item['pkg_path'])
+                        if item['uninstallpkg_path']:
+                            self.pkgs_to_keep.add(item['uninstallpkg_path'])
+                    line_info = "[pkginfo to be DELETED]"
+                else: 
+                     pass
                 if len(item_list) > 1:
                     line_info = (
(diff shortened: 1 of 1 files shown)


gregneagle:
How would doing this interact with "orphaned packages"? I'd think it would _create_ them...

Also, wondering why you don't just skip the pkg deleting section in this function if  `--skip-pkgs` is given instead of creating a  potentially huge structure containing the paths to a lot of packages to not delete.

gregneagle:
Ping.

midsandhighs:
Apologies for the long delay in response on this. 

I'm going to migrate it back to a _Draft_ so that I can update my fork to the latest release. My organization no longer hosts their own `munki` repository so I need to recreate the environment to test the changes against and perform debugging on the contents of `self.items_to_delete` & `self.pkgs_to_keep` to see if there is a more efficient way to handle the scenario listed in #991 

My understanding is there would be some process or worker external to `repoclean` performing the removal of the actual .pkg files but "orphaned packages" is not something I had factored into my code change. 

Thank you for your feedback and your time! 😄

gregneagle:
Are you still working on this?

gregneagle:
Closing. Please create a new PR if you resume work on this.

PR #1091, opened 2021-05-14
Add support for client cert package

rrenstrom:
Adds -T flag to make_munki_pkg build scripts to include a component package containing a client certificate for server mTLS, installed at /Library/Managed Installs/certs/client.pem

### DIFF
--- a/code/tools/make_munki_mpkg.sh
+++ b/code/tools/make_munki_mpkg.sh
@@ -23,6 +23,7 @@ CONFPKG=NO
 MDMSTYLE=NO
 ORGNAME=macOS
 ROSETTA2=NO
+CLIENTCERTPKG=NO
 
 # try to automagically find Munki source root
 TOOLSDIR=$(dirname "$0")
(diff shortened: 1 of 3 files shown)


code/tools/make_munki_mpkg.sh
```diff
@@ -897,6 +957,10 @@ for pkg in $ALLPKGS ; do
         "config")
             ver="1.0"
             ;;
+        "clientcert")
+            ver="1.0"
+            SCRIPTS="${MUNKIROOT}/code/tools/pkgresources/Scripts_nothing"
```
gregneagle, line 962: Why are you including a postinstall script that does nothing? Why not have no script?

rrenstrom: IIRC, if the SCRIPTS variable wasn't specified, the build script just included whatever script was last set for this variable, so at the time I worked around the problem by stubbing out this do nothing script, similar to Scripts_python, under the assumption that a script had to be defined.

But in looking at this again, I see that was a bad assumption, and it's better to just set the SCRIPTS variable to null. 

I'll make this change and test again.

gregneagle: I find that this entire script is getting too big and unwieldy, and probably desperately needs a re-think.

rrenstrom: Okay, that change, `SCRIPTS=""`, worked fine (checked resulting package in Suspicious Package).

Note that your `"bootstrap") `and `"config")` cases don't define SCRIPTS either (which is template I used first time in my code, before stubbing out a script), so they're just inheriting whatever this variable is set to, in this case the Scripts_python do nothing postinstall script is included in these component packages. Do you want to me also define `SCRIPTS=""` for these as well?

gregneagle: Yes, please.

rrenstrom: Changes pushed.

PR #1093, opened 2021-06-04
Add QuitButtonTimeout preference

natewalck:
This adds a QuitButtonTimeout preference so that one can make the timeout for the QuitButton longer or shorter depending upon the current state of the machine.

This is useful for situations where people are ignoring the updates despite the 5 second timeout and will be meant to increase awareness around the necessity of the update.

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center/Controllers/MSCAlertController.swift
+++ b/code/apps/Managed Software Center/Managed Software Center/Controllers/MSCAlertController.swift
@@ -568,7 +568,7 @@ class MSCAlertController: NSObject {
         alert.informativeText = alertDetail
         var quitButton = NSApplication.ModalResponse.alertFirstButtonReturn
         var updateButton = NSApplication.ModalResponse.alertSecondButtonReturn
-        let quitButtonTimeout = pref("quitButtonTimeout") as? Float ?? 5.0
+        let quitButtonTimeout = pref("QuitButtonTimeout") as? Float ?? 5.0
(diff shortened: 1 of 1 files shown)


gregneagle:
Won't most people intent on ignoring updates just Force Quit at this point?

natewalck:
That assumes people know how to force quit 🤣  (This is a good point though).

In my mind, this is for the 98% of people who might click away reflexively rather than the people who are intent on avoiding all updates.

gregneagle:
I just resist Yet Another Preference -- they increase the logic complexity and sometimes interact in unexpected ways. Is what you really want is for the delay to increase over time? The longer they ignore, the longer the delay?

natewalck:
Yep, that is exactly it. Basically, if they are 1 day after seeing it, 5 seconds is fine. If they are 10 days after seeing it, then it would be better to have a longer timeout to make them actually read the screen. It would be kind of nudge + deprecation notifier all rolled into one.

gregneagle:
This functionality is related to but not the same as the behavior where MSC takes over the UI and mostly prevents force quit and command-tab switching to another app, etc.

That aside, rather than a preference that you'd have to write external logic for to increase the delay over time, shouldn't this logic be built into Munki?

gregneagle:
Ping

natewalck:
Sorry for the delay, busy couple of days. 

I agree it would be better to have this be part of the aggressive notification logic itself. This would avoid yet another preference™  and be in line with the spirit of the aggressive notifications I'll probably be able to revisit this next week, but I'll think about it a bit until then.

gregneagle:
It is next week yet?

natewalck:
I don't have the bandwidth for this, so feel free to close it.

PR #1094, opened 2021-06-05
tighten security of keychain

rrenstrom:
Munki's keychain directory should be readable only by root


code/client/munkilib/keychain.py
```diff
@@ -250,7 +250,7 @@ def make_client_keychain(cert_info=None):
     if os.path.exists(abs_keychain_path):
         os.unlink(abs_keychain_path)
     if not os.path.exists(os.path.dirname(abs_keychain_path)):
-        os.makedirs(os.path.dirname(abs_keychain_path))
+        os.makedirs(os.path.dirname(abs_keychain_path),0o700)
```
gregneagle, line 253: Nit-picky but standard Python style requires a space after the comma.

rrenstrom: Fixed style.

PR #1143, opened 2022-06-14
Add raised exception to manifest retrieval output

danhutchings:
When running `managedsoftwareupdate` without an internet connection, the below error will happen:

```
ERROR: Could not retrieve managed install primary manifest.
```

This is a rare, but possible situation. Other Munki libraries add the raised exception to the output (ex: download.py). This PR includes a change that will add the raised exception to the manifest retrieval in `core.py`. Below is an example of the output could look like with these changes:

```
ERROR: Could not retrieve managed install primary manifest.: (-1009, "The Internet connection appears to be offline.")
```

For Munki administrators, this will be helpful when trying to diagnose why a system was unable to retrieve the manifest via log files.


gregneagle:
Seems like a reasonable, useful, and safe change.

PR #1151, opened 2022-08-10
Add 2022 M2 Air and 13" Pro to machine_type

jamesez:
This commit adds the NewWorld names for the 2022 M2 MacBook Air and 13" Pro, so they are correctly identified as 'laptop'.

### DIFF
--- a/code/client/munkilib/info.py
+++ b/code/client/munkilib/info.py
@@ -885,9 +885,14 @@ def predicate_info_object():
     build = machine['os_build_number']
     info_object['os_build_last_component'] = pkgutils.MunkiLooseVersion(
         build).version[-1]
-    if 'Book' in machine.get('machine_model', ''):
+    machine_model = machine.get('machine_model', '')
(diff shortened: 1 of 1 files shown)


gregneagle:
I understand what you are doing here, and can see that it addresses the immediate issue. But it means that now we have a manually-maintained list of ModelIdentifiers that are laptops, and it means every time Apple releases new hardware, a new release of Munki will be required. And that means there will almost always be a lag for some people before Munki works "properly" on new hardware.  I'd prefer if we found/developed a way to identify laptops without having to have a manually-maintained list. Some possibilities include using `ioreg` to find alternate model info, or looking for a battery.

jamesez:
I've re-coded it to use product name (e.g., `Mac mini (M1, 2020)`), and fetch the product name from the IORegistry. This will probably work fine until Apple releases a Macintosh Portable, like that'd ever happen.

jamesez:
I give up, this won't work on Apple Silicon

PR #1153, opened 2022-08-10
Change to detecting laptop/desktop by identifying if the system has a battery

jamesez:
primary author: @gregneagle, via Slack

### DIFF
--- a/code/client/munkilib/info.py
+++ b/code/client/munkilib/info.py
@@ -675,6 +675,28 @@ def has_intel64support():
 
     return buf.value == 1
 
+def has_battery():
+    """Determine if this Mac has a power source of 'InternalBattery'"""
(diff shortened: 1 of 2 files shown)


gregneagle:
I'm going to be perhaps a bit arbitrary and ask that the function be moved to `powermgr.py` where it can take advantage of some existing NSBundle function definitions (the function is really just a simplified version of `getBatteryPercentage()`) and also perhaps be renamed `hasInternalBattery()` to reflect that it won't be confused by external (UPS) batteries.

gregneagle:
I also wonder if there's any logic in Managed Software Center that needs to be updated.

gregneagle:
(I did a quick search and I _think_ no changes are needed in Managed Software Center)

jamesez:
I also can't find any other laptop/desktop detection code in the entire repo, so I think MSC.app is clear.

PR #1155, opened 2022-08-26
Adjust Search field size and position to better match other Apple apps

chrisgrande:
This is just a minor UI tweak of the search field to better match how Apple displays them in sidebars.
(Tested on 10.15+)

Comparison is the App Store and Music


gregneagle:
Looks great, thanks!

PR #1162, opened 2022-10-25
Fix Client Certificate keychain search for Ventura  

rrenstrom:
This fixes a problem where Munki can't search for the Client Certificate in the munki.keychain in macOS Ventura.

In grul.py NSURLAuthenticationMethodClientCertificate handler, the SecItemCopyMatching function doesn't find the client cert, since the munki.keychain is added to the keychain search list for the User domain, and Ventura isn't allowing Munki launchd jobs to access this domain.

Apple's guidance is launch daemons don't have access to keychains in the User context, so somewhat surprising this worked on earlier macOS versions; apparently Ventura is now enforcing this.

https://developer.apple.com/documentation/security/secpreferencesdomain

This PR address this problem by adding munki.keychain to the Common domain keychain search list, rather than the User domain. The Common domain search list is for all users and the system. And if there's a problem searching the keychain or if a matching client cert can't be found, added code to fail more gracefully.

Tested in Ventura public release (22A380), Monterey, Big Sur, Catalina, Mojave

### DIFF
--- a/code/client/munkilib/gurl.py
+++ b/code/client/munkilib/gurl.py
@@ -641,6 +641,8 @@ class Gurl(NSObject):
                     None)
             else:
                 challenge.sender().cancelAuthenticationChallenge_(challenge)
+
+        # Handle HTTP Basic and Digest challenge
         if self.username and self.password and authenticationMethod in [
                 'NSURLAuthenticationMethodDefault',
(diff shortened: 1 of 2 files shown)


gregneagle:
I see "jazz-o-mat" in MacAdmins Slack tested successfully.

PR #1182, opened 2023-07-15
Set consoleuser and thisuser to lowercase

MScottBlake:
Since macOS usernames are case insensitive, this change normalizes to lowercase so that the comparison is also case insensitive.

### DIFF
--- a/code/client/launchapp
+++ b/code/client/launchapp
@@ -49,9 +49,9 @@ def getconsoleuser():
 def main():
     '''Pass arguments to /usr/bin/open only if we are the current
     console user or no user and we're at the login window'''
-    consoleuser = getconsoleuser()
+    consoleuser = getconsoleuser().lower()
     try:
-        thisuser = os.environ['USER']
+        thisuser = os.environ['USER'].lower()
     except KeyError:
         # when run via launchd at loginwindow context, os.environ['USER']
         # is undefined, so we'll return root (the effective user)


gregneagle:
Have you actually encountered an issue this change addresss? I'd be interested to hear more...

MScottBlake:
I have, actually, and I'm working on getting Jamf to change as well. The short story is that Jamf Connect normalizes the username to all lowercase. If your username is `Scott` and then you install Jamf Connect, `getconsoleuser()` will return `scott` while `os.environ['USER']` continues to return `Scott`. The result is that the comparison on line 60 fails, even though it's the same user.

I don't agree that they should normalize the username of existing accounts, but the fact that it works and it breaks Munki is also an issue.

gregneagle:
It's the "Since macOS usernames are case insensitive" assertion I'm not 100% sure about.
The fact that _Jamf_ converts all usernames to lowercase doesn't necessarily imply to me that _macOS_ won't allow a username of "Scott" and another of "scott". This is something to investigate.

gregneagle:
Using the Users & Groups GUI, I created a user named "Scott".  I then tried to create a user named "scott" and got a warning "Name is used by another user". I wonder, though, if this limitation carries through to external directory services like AD and LDAPv3.

gregneagle:
I also tried to use `dscl` to create a "Scott" and a "scott" user, and this failed, but in a way I wasn't expecting. The attempt to create the second user simply edited the first. And `dscl . read /Users/$NAME` is definitely case-insensitive: `dscl . read /Users/SCOTT`, `dscl . read /Users/scott`, and `dscl . read /Users/Scott` all return the same record data. I'd like to do a little testing against an LDAP directory we have here next.

MScottBlake:
Makes perfect sense. You've thought of a couple scenarios I hadn't. I also tried things such as su/sudo and it didn't care about case sensitivity either. If you have any other scenarios you want to test, I may be able to help you out.

gregneagle:
OK, reading user account data from an LDAPv3 directory (using dscl) also appears to be case-insensitive. So this change seems safe.

gregneagle:
I'm merging into Munki6dev.

gregneagle:
One final comment -- it's interesting to me that launchapp has been essentially unchanged for over a _decade_ and no-one has encountered this issue. Seems a pretty difficult situation to get into!

MScottBlake:
I agree. For us, the symptom was no Munki notifications and it certainly took some time to work out the underlying issue.

PR #1184, opened 2023-07-18
Fix #1183

### LINKED ISSUE #1183: app_usage_monitor fails trying to call superclass's init
Running `/usr/local/munki/app_usage_monitor`

results in
```
  File "/usr/local/munki/app_usage_monitor", line 235, in <module>
    main()
  File "/usr/local/munki/app_usage_monitor", line 224, in main
    notification_handler = NotificationHandler.alloc().init()
  File "/usr/local/munki/app_usage_monitor", line 117, in init
    self = super().init()
AttributeError: 'super' object has no attribute 'init'
sys:1: UninitializedDeallocWarning: leaking an uninitialized object of type NotificationHandler
```

Fortunately this is easy to solve-I have a PR incoming.

sheagcraig:
PyObjC now requires the use of `objc.super` instead of `builtin.super` when calling superclass methods in a Cocoasubclass, which `app_usage_monitor.NotificationHandler` is.
https://github.com/ronaldoussoren/pyobjc/blob/master/docs/changelog.rst#version-92

The original issue: https://github.com/ronaldoussoren/pyobjc/issues/549

This PR updates the `super` call to make it more idiomatic python3 (i.e. argument-less) and imports the `objc.super` to mask the builtin one for this class.

I checked the rest of the module, and no other `super` usage exists, so this should be safe.

### DIFF
--- a/code/client/app_usage_monitor
+++ b/code/client/app_usage_monitor
@@ -35,6 +35,11 @@ import select
 import socket
 import sys
 
+# builtin super doesn't work with Cocoa classes in recent PyObjC releases.
+# pylint: disable=redefined-builtin,no-name-in-module
+from objc import super
+# pylint: enable=redefined-builtin,no-name-in-module
+
 try:
     # Apple frameworks
     from AppKit import NSWorkspace
@@ -43,7 +48,6 @@ try:
     from Foundation import NSDistributedNotificationCenter
     from Foundation import NSObject
     from Foundation import NSRunLoop
-    from objc import super
 except ImportError:
     logging.critical("PyObjC wrappers for Apple frameworks are missing.")
     sys.exit(-1)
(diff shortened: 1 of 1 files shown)


gregneagle:
It would be helpful for me to understand what specific change broke this -- yes, it seems an update to PyObjC was responsible, but update to _Munki_ broke this?

gregneagle:
Reading through @ronaldoussoren's notes, it looks like PyObjC 9.1 caused this breakage, which would mean it broke in the Munki 3.6.2 release.

sheagcraig:
That's my best guess too-PyObjC 9.1 is likely the culprit. The issue that is referenced in the changelog says it was working fine in 9.0.1 and broke for them on 9.1.1. 9.2 just adds in warnings about importing the `objc.super`.

sheagcraig:
https://github.com/munki/munki/commit/5f64d7941a738a7e42d98d839679dfc80dd439c0#diff-788129623951922b23efe87d95380a748190b316fdf06536fc6a8b0e22317ad8R5

PR #1186, opened 2023-07-28
Add the flag "ignore_min_version_change_less_than" to repoclean

kbrewersq:
This allows me to automatically clean my repo a bit more by grouping together versions on an application that has changed its `minimum_os_version` by minor versions I do not need to support. It defaults to behave exactly as it does today.

```
repoclean --keep=2 ./repo --ignore_min_version_change_less_than 11
```

### DIFF
--- a/code/client/repoclean
+++ b/code/client/repoclean
@@ -249,6 +249,10 @@ class RepoCleaner(object):
                     if key == 'receipts':
                         value = ', '.join(sorted(
                             [item.get('packageid', '') for item in value]))
+                    if key == "minimum_os_version" and self.options.ignore_min_version_change_less_than > 0:
+                        major_version = value.split('.')[0]
+                        if int(major_version) < self.options.ignore_min_version_change_less_than:
+                            value = self.options.ignore_min_version_change_less_than - 1
                     metakey += u"%s: %s\n" % (key, value)
             metakey = metakey.rstrip('\n')
             if metakey not in self.pkginfodb:
@@ -352,7 +356,7 @@ class RepoCleaner(object):
                 print("\t%s" % pkg)
         if print_this:
             print()
-            
+
 
         print("Total pkginfo items:     %s" % self.pkginfo_count)
         print("Item variants:           %s" % len(list(self.pkginfodb.keys())))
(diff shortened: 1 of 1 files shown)


gregneagle:
Could you explain the desired/expected behavior here? If an item has a minimum_os_version of 10.6 it may still work in 14.0.

I'd think you might want to filter out by _maximum_os_version_; that is, perhaps you don't support anything before macOS 11.0 and so you want to remove versions that have a maximum_os_version of 10.15.x or lower...

kbrewersq:
I'm looking to group the apps that have their minimum_os_version change from 10.16 -> 10.17 and group those as "minimum_os_version" = 10 (or variable-1) so that they would be considered "similar apps" in the keep sense. today I can have 3 apps: 

```
name: App1
catalogs: Production
minimum_os_version: 10.16
versions:
     8.99.0.403 (pkgsinfo/apps/App1/App1-8.99.0.403.plist) 
     8.99.0.202 (pkgsinfo/apps/App1/App1-8.99.0.202.plist) 

name: App1
catalogs: Production
minimum_os_version: 10.13
versions:
     8.98.0.403 (pkgsinfo/apps/App1/App1-8.98.0.403.plist) 

```

Where I don't care that the minimum version has jumped from 10.13-> 10.16 because I don't support those versions (I completely understand the need, which is why I wanted to make this compatible still). By grouping these changes together these 3 apps look the same in the eyes of repoclean 

```
name: App1
catalogs: Production
minimum_os_version: 10
versions:
     8.99.0.403 (pkgsinfo/apps/App1/App1-8.99.0.403.plist) 
     8.99.0.202 (pkgsinfo/apps/App1/App1-8.99.0.202.plist) 
     8.98.0.403 (pkgsinfo/apps/App1/App1-8.98.0.403.plist) 
```

allowing the 3rd one to be culled easily.

This could be extended to other pkginfo keys, this just the one that I felt caused me to parse through the `--show-all` and look deeper at why things weren't getting thrown out.

gregneagle:
I still find this confusing and am worried that others will as well. I think we need to figure out a cleaner/clearer way to meet the need here, which is, as I understand it, make it easier to identify older items that can be deleted that are currently preserved because they support older OSes.

PR #1194, opened 2023-11-29
Add preference to ignore middleware

patgmac:
This PR adds the preference IgnoreMiddleware. This could be helpful when moving away from using middleware because of a chicken & egg scenario where your new SoftwareRepoURL no longer needs middleware.

### DIFF
--- a/code/client/munkilib/fetch.py
+++ b/code/client/munkilib/fetch.py
@@ -118,7 +118,11 @@ def import_middleware():
 
 
 middleware = None
-import_middleware()
+ignore_middleware = prefs.pref('IgnoreMiddleware')
+if ignore_middleware is not True:
+        # If we haven't explicitly said to ignore middleware,
+        # the preference decides
+        import_middleware()
 
 
 class Error(Exception):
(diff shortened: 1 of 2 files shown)


gregneagle:
If merged, this probably wouldn’t make it into a release any time soon. If one had an immediate need, the middleware itself could check for the/a preference and return having done nothing at all…

patgmac:
Understood. I do have other ideas on how to handle this, but they involved using other tools (WS1 or Tanium). I do like the idea of handling this in a new revision of the middleware itself to check for that preference. I may go that route if needed. 

I thought this might be a good idea because of projects such as https://github.com/jc0b/munki-gcs-redirector which is a GCP cloud run service that does not need middleware and some of us previously needed to use middleware to talk to GCP (or other cloud providers).

code/client/munkilib/fetch.py
```diff
@@ -118,7 +118,11 @@ def import_middleware():
 
 
 middleware = None
-import_middleware()
+ignore_middleware = prefs.pref('IgnoreMiddleware')
+if ignore_middleware is not True:
```
gregneagle, line 122: Arguably more Pythonic to say
```
if not ignore_middleware:
```

gregneagle:
Still trying to decide if the pref should be "ignoreMiddleware" with a default of False, or "useMiddleware" with a default of True...

gregneagle:
Crud. Forgot this was here. I'll merge it now for a future (post 6.4.1) release.

PR #1199, opened 2024-01-20
Provide pkginfo version as parameter 1 to scripts

macBerlin:
Read the productversion from plist to use this a parameter 1 in the scripts. Useful to compare to compare the local version with the new installer.

### DIFF
--- a/code/client/munkilib/scriptutils.py
+++ b/code/client/munkilib/scriptutils.py
@@ -53,6 +53,7 @@ def run_embedded_script(scriptname, pkginfo_item, suppress_error=False):
 
     # get the script text from the pkginfo
     script_text = pkginfo_item.get(scriptname)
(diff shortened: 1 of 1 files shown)


gregneagle:
Since the script is also part of the pkginfo, you can just add the version to the script. I'll need some convincing on this, and it will need a better description and documentation to be considered. (One could really argue for all sorts of data to be available from the pkginfo this way...)

gregneagle:
You also need to be careful and check that your proposed changes don't cause unintended issues elsewhere. A quick search for "run_script(" in the code shows us that your proposed changes (as-is) would break this: https://github.com/munki/munki/blob/main/code/client/munkilib/installer/core.py#L612-L613

gregneagle:
I'm going to close this due to lack of discussion/engagement from the originator. Feel free to reopen.

PR #1209, opened 2024-03-25
MunkStatus.app icon amend

macmule:
This change amends to the icon shown within MunkiStatus.app's window from `AppIcon` to `NSApplicationIcon`.

This means that if an icon is updated via [NSWorkspace's setIcon](https://developer.apple.com/documentation/appkit/nsworkspace/1529882-seticon) the icon updates within the MunkiStatus.app too.

99% of folks will not notice this change at all.

The below can be used to change the icon of the MunkiStatus.app to test this pre and post the change, just need to update `objectPath` as needed.

```
// Imports
import AppKit

// Gets the modification date and time of the file and return as epoch
func getImageDetails(brandingImagePath: String) -> NSImage? {
    // Var declaration
    var imageData: NSImage?
    // If the file exists
    if FileManager.default.fileExists(atPath: brandingImagePath) {
        // Try to get the attributes of the file at brandingImagePath date of the file
        let fileAttributes = try? FileManager.default.attributesOfItem(atPath: brandingImagePath)
        // Get modification date
        guard let modificationDate = fileAttributes?[FileAttributeKey.modificationDate] as? Date else {
            // Return 0 and nil, so we do not proceed
            return nil
        }
        // Create imageData from the file passed to brandingImage
        imageData = NSImage(contentsOfFile: brandingImagePath)
        // If imageData isValid is nil, then brandingImage is not a valid icon
        if (imageData?.isValid) == nil {
            // Return 0 and nil, so we do not proceed
            return nil
        }
        // Get the modification date, in an epoch
        let modificationEpoch = Int(modificationDate.timeIntervalSince1970)
        // Return the modification date as epoch
        return imageData
    } else {
        // Return 0 and nil, so we do not proceed
        return nil
    }
}

// The image to use
let brandingImagePath = "/System/Library/CoreServices/CoreTypes.bundle/Contents/Resources/ErasingIcon.icns"
// NSImage representation of brandingImagePath
let imageData = getImageDetails(brandingImagePath: brandingImagePath)


// Path to the MunkiStatus.app to rebrand
let objectPath = "/path/to/MunkiStatus.app"

// Set the icon, returns bool
let rebrandStatus = NSWorkspace.shared.setIcon(imageData, forFile: objectPath,
                                               options: NSWorkspace.IconCreationOptions([]))
```


macmule:
Bugger.. basing this PR on dev brought over the other changes.

It's a quick simple change though.. should I submit against main or another branch @gregneagle?

gregneagle:
There's something seriously amiss here: 
image

gregneagle:
If this is something you've done _after_ all the other work moving the helper apps around, you should either add this to that PR, or wait for me to merge that PR, then rebase your new PR on that.

gregneagle:
And most of the time I want PRs to Munki6dev, not to main, unless they are simple/trivial bug fixes.

macmule:
Yes, that's what I said prior.

I tried to base this off of the munki6dev branch, but it pulled the changes over... want me to resubmit against another branch?

macmule:
https://github.com/munki/munki/pull/1209#issuecomment-2021129053 this is separate to that stuff, yes.. I'll re-submit after that PR is dealt with.

gregneagle:
There is currently no other reasonable branch to base against. Either wait for my merge to macmule-installhelper, or add these changes to your other PR (I'd prefer the former).

PR #1210, opened 2024-03-28
Macmule installhelper

macmule:
@gregneagle I've made the changes to installhelper as advised, and also lumped in the small MunkiStatus.app change.

Have tested and seems good my side


macmule:
Ugh.. apologies @gregneagle, this was supposed to update the branch.

PR #1214, opened 2024-04-01
custom sidebar item

SteveKueng:
we use Munki in combination with Intune.
To make it easier for users, I would like to integrate the company portal web directly into Munki.
As a possible solution I have built a custom sidebar item.
The title, icon and link of the custom sidebar item can be configured via ManagedInstalls.plist

ManagedInstalls.plist:
{
    AppleSoftwareUpdatesOnly = 0;
    CustonSidebarItem =     {
        icon = desktopcomputer;
        link = "https://portal.manage.microsoft.com/devices";
        title = "My Devices";
    };
    FollowHTTPRedirects = none;
    IgnoreMiddleware = 0;
    IgnoreSystemProxies = 0;
    InstallAppleSoftwareUpdates = 0;
    LastCheckDate = "2024-04-01 14:56:58 +0000";
    LastCheckResult = 0;
    LogFile = "/Library/Managed Installs/Logs/ManagedSoftwareUpdate.log";
    LogToSyslog = 0;
    LoggingLevel = 1;
    ManagedInstallDir = "/Library/Managed Installs";
    OldestUpdateDays = 0;
    PendingUpdateCount = 0;
    UseClientCertificate = 0;
}

defaults write /Library/Preferences/ManagedInstalls.plist CustonSidebarItem '<dict><key>title</key><string>My Device</string><key>icon</key><string>desktopcomputer</string><key>link</key><string>https://portal.manage.microsoft.com/devices</string></dict></dict>' 

SCR-20240401-pxyq

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
+++ b/code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
@@ -1088,7 +1088,8 @@ class MainWindowController: NSWindowController, NSWindowDelegate, WKNavigationDe
                 decisionHandler(.cancel)
                 return
             }
-            if scheme == "mailto" || scheme == "http" || scheme == "https" {
+            
+            if scheme == "mailto" {
                 // open link in default mail client since WKWebView doesn't
                 // forward these links natively
                 NSWorkspace.shared.open(url)
--- a/code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
+++ b/code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
@@ -21,7 +21,7 @@ class MainWindowController: NSWindowController, NSWindowDelegate, WKNavigationDe
     var htmlDir = ""
     var wkContentController = WKUserContentController()
     
-    let sidebar_items = [
+    var sidebar_items = [
         ["title": "Software", "icon": "AllItemsTemplate"],
         ["title": "Categories", "icon": "toolbarCategoriesTemplate"],
         ["title": "My Items", "icon": "MyStuffTemplate"],
@@ -75,6 +75,8 @@ class MainWindowController: NSWindowController, NSWindowDelegate, WKNavigationDe
                         loadMyItemsPage(self)
                     case 3:
                         loadUpdatesPage(self)
+                    case 4:
+                        loadCustomPage(self)
                     default:
                         loadUpdatesPage(self)
             }
@@ -292,6 +294,15 @@ class MainWindowController: NSWindowController, NSWindowDelegate, WKNavigationDe
     
     func loadInitialView() {
         // Called by app delegate from applicationDidFinishLaunching:
+        
+        // add custom link if nessesary
+        if let CustonMenuItem = pref("CustonMenuItem") as? Dictionary<String, String> {
+            if let title = CustonMenuItem["title"], let icon = CustonMenuItem["icon"] {
+                sidebar_items.append( ["title": title, "icon": icon])
+                self.sidebar.reloadData()
+            }
+        }
+        
         if optionalInstallsExist() {
             loadAllSoftwarePage(self)
         } else {
@@ -940,10 +951,17 @@ class MainWindowController: NSWindowController, NSWindowDelegate, WKNavigationDe
         msc_debug_log("load_page request for \(url_fragment)")
         
         let html_file = NSString.path(withComponents: [htmlDir, url_fragment])
-        let request = URLRequest(url: URL(fileURLWithPath: html_file),
+        var request = URLRequest(url: URL(fileURLWithPath: html_file),
                                  cachePolicy: .reloadIgnoringLocalCacheData,
                                  timeoutInterval: TimeInterval(10.0))
+        if url_fragment.starts(with: "http") {
+            request = URLRequest(url: URL(string: url_fragment)!,
+                                     cachePolicy: .reloadIgnoringLocalCacheData,
+                                     timeoutInterval: TimeInterval(10.0))
+        }
+        
         webView.load(request)
+        
         if url_fragment == "updates.html" {
             if !_update_in_progress && NSApp.isActive {
                 // clear all earlier update notifications
@@ -1623,6 +1641,18 @@ class MainWindowController: NSWindowController, NSWindowDelegate, WKNavigationDe
         load_page("updates.html")
     }
     
+    @IBAction func loadCustomPage(_ sender: Any) {
+        // Called by Navigate menu item'''
+        clearSearchField()
+        var page = "updates.html"
+        if let CustonMenuItem = pref("CustonMenuItem") as? Dictionary<String, String> {
+            if let link = CustonMenuItem["link"] {
+                page = link
+            }
+        }
+        load_page(page)
+    }
+    
     @IBAction func searchFilterChanged(_ sender: Any) {
         // User changed the search field
         let filterString = searchField.stringValue.lowercased()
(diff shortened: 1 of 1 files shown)


code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
```diff
@@ -292,6 +294,15 @@ class MainWindowController: NSWindowController, NSWindowDelegate, WKNavigationDe
     
     func loadInitialView() {
         // Called by app delegate from applicationDidFinishLaunching:
+        
+        // add custom link if nessesary
+        if let CustonSidebarItem = pref("CustonSidebarItem") as? Dictionary<String, String> {
```
gregneagle, line 299: Need to change the name to "CustomSidebarItem" in all occurrences.

code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
```diff
@@ -1070,7 +1088,8 @@ class MainWindowController: NSWindowController, NSWindowDelegate, WKNavigationDe
                 decisionHandler(.cancel)
                 return
             }
-            if scheme == "mailto" || scheme == "http" || scheme == "https" {
+            
+            if scheme == "mailto" {
```
gregneagle, line 1092: Does this then change the behavior for http/https links from other contexts, like in item descriptions?

code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
```diff
@@ -1670,6 +1701,16 @@ extension MainWindowController: NSOutlineViewDelegate {
             }
             if let imageView = view?.imgView {
                 imageView.image = NSImage(named: NSImage.Name(icon))?.tint(color: .secondaryLabelColor)
+                if #available(macOS 11.0, *) {
```
gregneagle, line 1704: What does this mean if an admin has a custom sidebar item and the user is running < macOS 11?

gregneagle:
I like the general idea. I have a few questions and comments, some of which are above. As for the icon, what values are legal, and where is the icon sourced? This will need good documentation, and I expect people might want to provide a custom icon as well. I wonder if this (instead of being in preferences) might better "live" in the other customization resources for MSC.app...

gregneagle:
Once the door has been opened to custom sidebar items, I can see admins wanting _multiple_ such items. Could this be easily implemented in a way that either supports multiple items now, or could be easily extended in the future? (This again maybe pushes the config into the custom client resources instead of ManagedInstalls preferences)

code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
SteveKueng: I must have missed it :)

SteveKueng:
SF Symbols are currently used. These are integrated in the OS and can be specified in the configuration. https://developer.apple.com/sf-symbols/ (icon name is the SF symbol name)
I was thinking of using images specified by the admin, but that's challenging with light and dark mode...
The downside of SF symbol is that it only works from 11.0. Therefore the restriction. With older systems no symbol is displayed.

I have already thought about several links. They should be easy to implement.
You could simply specify an array and then display it.
I can only think of one section that is a bit more difficult to set dynamically.

code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
SteveKueng: SF symbol only works with macOS 11.0 and above. Therefore the restriction. With older systems no symbol is displayed.
That could definitely be solved a little better

code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
SteveKueng: I'm not quite sure yet.
I'm still testing when and how this part triggers.

SteveKueng:
> Once the door has been opened to custom sidebar items, I can see admins wanting _multiple_ such items. Could this be easily implemented in a way that either supports multiple items now, or could be easily extended in the future? (This again maybe pushes the config into the custom client resources instead of ManagedInstalls preferences)

How do you mean the configuration in a custom client resource? With a separate config?

gregneagle:
> How do you mean the configuration in a custom client resource? With a separate config?

Yes, in a new file in the client customization archive (https://github.com/munki/munki/wiki/Client-Customization)

code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
gregneagle: It's "CustonSidebarItem" pretty much everywhere, and should be "CustomSidebarItem".

code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
gregneagle: Or just declare no support for custom sidebar items on macOS < 11.

code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
SteveKueng: OMG... you're right... I've been working on it for too long...

code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
gregneagle: Custom links that don't have `target="_blank"` would go though this code path, I'd think (Links in client customization, like the footer, the banners, and the (former) sidebar-now-link-strip)

SteveKueng:
I have updated this sidebar customization.
On the one hand the design is now aligned with the AppStore, on the other hand the sidebar can now be set completely custom.
I have not changed the default behavior. However, as soon as CustomSidebarItems are defined, only these are displayed. The views such as software, updates, etc. can then be displayed by selecting the page accordingly.
This allows you to freely design the sidebar.
Here is another picture:
SCR-20250705-soox

The sidebar can be configured as follows:
```
defaults write /Library/Preferences/ManagedInstalls.plist CustomSidebarItems '
<array>
    <dict>
        <key>title</key>
        <string>Software</string>
        <key>icon</key>
        <string>star</string>
        <key>page</key>
        <string>loadAllSoftwarePage</string>
    </dict>
    <dict>
        <key>title</key>
        <string>My Device</string>
        <key>icon</key>
        <string>desktopcomputer</string>
        <key>page</key>
        <string>https://portal.manage.microsoft.com/devices/c319a32f-6f50-48c1-b1ca-953b42178648</string>
    </dict>
    <dict>
        <key>title</key>
        <string>Helpdesk</string>
        <key>icon</key>
        <string>questionmark.circle</string>
        <key>page</key>
        <string>https://portal.manage.microsoft.com/helpdesk</string>
    </dict>
    <dict>
        <key>title</key>
        <string>My Homepage</string>
        <key>icon</key>
        <string>book.pages</string>
        <key>page</key>
        <string>https://umb.ch</string>
    </dict>
    <dict>
        <key>title</key>
        <string>Updates</string>
        <key>icon</key>
        <string>square.and.arrow.down</string>
        <key>page</key>
        <string>loadUpdatesPage</string>
    </dict>
</array>'
```

the allowed values for “page” is an http link or the following:
- loadAllSoftwarePage
- loadCategoriesPage
- loadMyItemsPage
- loadUpdatesPage

currently only the munki standard symbols or “SF Symbols” (https://developer.apple.com/sf-symbols) can be used as symbols

gregneagle:
I changed the base branch to Munki7dev which reduced the number of commits to be merged from 250(!) to 10. I'll play with this a bit and see how it feels.

gregneagle:
I'm going to merge this into a new `custom-sidebar` branch, then open issues we can work on together.

PR #1221, opened 2024-06-30
Add amd64 to supported_architectures

kevinmcox:
This adds `amd64` as an equivalent option to `x86_64` in the `supported_architectures` array. It is intended as a quality of life improvement specifically related to AutoPkg.

Many software titles use `amd64` and `arm64` to differentiate architectures and AutoPkg recipes need to specify one or the other for the correct download.

However this variable cannot be reused in building Munki's `supported_architectures` array requiring the use of a second variable to specify `x86_64`.

**EXAMPLES:**
https://github.com/autopkg/gregneagle-recipes/blob/master/Podman/Podman.munki.recipe#L7-L8
https://github.com/autopkg/timsutton-recipes/blob/master/Go/Go.munki.recipe#L18-L22

By having Munki support both `x86_64` and `amd64` as equivalents, it allows the simplification of dual-architecture AutoPkg recipes.

**LOGS:**
Before change:
```
    * Processing manifest item test.amd64 for optional install
    Looking for detail for: test.amd64, version latest...
    Considering 1 items with name test.amd64 from catalog development
    Considering item test.amd64, version 0.2.8 with minimum os version required 14.0
    Our OS version is 14.5
    Considering item test.amd64, version 0.2.8 with supported architectures: (
    amd64
)
    Our architecture is x86_64
    Not found
    Rejected item test.amd64, version 0.2.8 with supported architectures: (
    amd64
). Our architecture is x86_64.

    * Processing manifest item test.arm64 for optional install
    Looking for detail for: test.arm64, version latest...
    Considering 1 items with name test.arm64 from catalog development
    Considering item test.arm64, version 0.2.8 with minimum os version required 14.0
    Our OS version is 14.5
    Considering item test.arm64, version 0.2.8 with supported architectures: (
    arm64
)
    Our architecture is x86_64
    Not found
    Rejected item test.arm64, version 0.2.8 with supported architectures: (
    arm64
). Our architecture is x86_64.

    * Processing manifest item test.x86 for optional install
    Looking for detail for: test.x86, version latest...
    Considering 1 items with name test.x86 from catalog development
    Considering item test.x86, version 0.2.8 with minimum os version required 14.0
    Our OS version is 14.5
    Considering item test.x86, version 0.2.8 with supported architectures: (
    "x86_64"
)
    Our architecture is x86_64
    Found test.x86, version 0.2.8 in catalog development
    Checking existence of /usr/local/bin/test...
    Adding test.x86 to the optional install list
```
After change:
```
    * Processing manifest item test.amd64 for optional install
    Looking for detail for: test.amd64, version latest...
    Considering 1 items with name test.amd64 from catalog development
    Considering item test.amd64, version 0.2.8 with minimum os version required 14.0
    Our OS version is 14.5
    Considering item test.amd64, version 0.2.8 with supported architectures: (
    amd64
)
    Our architecture is x86_64
    Found test.amd64, version 0.2.8 in catalog development
    Checking existence of /usr/local/bin/test...
    Adding test.amd64 to the optional install list

    * Processing manifest item test.arm64 for optional install
    Looking for detail for: test.arm64, version latest...
    Considering 1 items with name test.arm64 from catalog development
    Considering item test.arm64, version 0.2.8 with minimum os version required 14.0
    Our OS version is 14.5
    Considering item test.arm64, version 0.2.8 with supported architectures: (
    arm64
)
    Our architecture is x86_64
    Not found
    Rejected item test.arm64, version 0.2.8 with supported architectures: (
    arm64
). Our architecture is x86_64.

    * Processing manifest item test.x86 for optional install
    Looking for detail for: test.x86, version latest...
    Considering 1 items with name test.x86 from catalog development
    Considering item test.x86, version 0.2.8 with minimum os version required 14.0
    Our OS version is 14.5
    Considering item test.x86, version 0.2.8 with supported architectures: (
    "x86_64"
)
    Our architecture is x86_64
    Found test.x86, version 0.2.8 in catalog development
    Checking existence of /usr/local/bin/test...
    Adding test.x86 to the optional install list
```
makepkginfo before:
```
 % makepkginfo -f /Applications/Managed\ Software\ Center.app --arch amd64
Usage: makepkginfo [options] [/path/to/installeritem]
       makepkginfo --help for more information.

makepkginfo: error: option --arch: invalid choice: 'amd64' (choose from 'i386', 'x86_64', 'arm64')
```
makepkginfo after:
```
% makepkginfo -f /Applications/Managed\ Software\ Center.app --arch amd64
....
	</array>
	<key>supported_architectures</key>
	<array>
		<string>amd64</string>
	</array>
	<key>version</key>
... 
```

### DIFF
--- a/code/client/munkilib/admin/pkginfolib.py
+++ b/code/client/munkilib/admin/pkginfolib.py
@@ -1073,7 +1073,7 @@ def add_option_groups(parser):
     additional_options.add_option(
         '--arch', '--supported_architecture', '--supported-architecture',
         action="append",
-        choices=['i386', 'x86_64', 'arm64'],
+        choices=['i386', 'x86_64', 'amd64', 'arm64'],
         metavar='ARCH',
         help=('Declares a supported architecture for the item. '
               'Can be specified multiple times to declare multiple '
--- a/code/client/munkilib/updatecheck/catalogs.py
+++ b/code/client/munkilib/updatecheck/catalogs.py
@@ -520,9 +520,10 @@ def get_item_detail(name, cataloglist, vers='',
                 item['name'], item['version'], item['supported_architectures'])
             display.display_debug1(
                 'Our architecture is %s', machine['arch'])
-            if machine['arch'] in item['supported_architectures']:
+            if (machine['arch'] in item['supported_architectures'] or
+                    (machine['arch'] == 'x86_64' and 'amd64' in item['supported_architectures'])):
                 return True
-            if ('x86_64' in item['supported_architectures'] and
+            if ({'x86_64', 'amd64'}.intersection(set(item['supported_architectures'])) and
                     machine['arch'] == 'i386' and
                     machine['x86_64_capable'] is True):
                 return True


gregneagle:
I don't understand the purpose of this. No Mac from Apple will report its own architecture as "amd64", so how would it be useful to have an item that declares it supports "amd64"?

kevinmcox:
The benefit would be in relation to AutoPkg and simplification of recipes.

For example this recipe of yours:
https://github.com/autopkg/gregneagle-recipes/blob/master/Podman/Podman.munki.recipe#L7-L8
```
GITHUB_ARCH: one of "amd64" (for Intel) or "arm64" (for Apple silicon)
MUNKI_ARCH: one of "x86_64" (for Intel) or "arm64" (for Apple silicon)
```
Could be simplified to:
```
ARCH: "amd64" (for Intel) or "arm64" (for Apple Silicon)
```
It would allow a single variable to work for both the download and the `supported_architectures` array, instead of AutoPkg users needing to configure multiple variables with different values to signify the same thing.

gregneagle:
Either I am missing something, or you are. Again: no Mac is going to report its own architecture as "amd64" and therefore it's never going to match an item that declares it supports "amd64". Macs released in the past ten years report their architecture as either "x86_64" or "arm64". You can put whatever values you like in the "supported_architectures" array, but only "x86_64" or "arm64" will be actually _useful_ with Macs released in the past ten years.

kevinmcox:
Right I understand that, I'm obviously doing a bad job of explaining, apologies for that.

The Intel computer itself will always report `x86_64` and never `amd64`.

The goal is to allow admins to set either `x86_64` or `amd64` in the `supported_architectures` array of a PkgInfo file and have Munki evaluate them as equivalents with the same end result.
```
            if (machine['arch'] in item['supported_architectures'] or
                    (machine['arch'] == 'x86_64' and 'amd64' in item['supported_architectures'])):
                return True
```
The reason is because of the large number of software developers that use `amd64` instead of `x86_64` in their releases. 

Right now this requires AutoPkg recipes to use two different variables for Intel releases. One variable is needed to download the correct package (`amd64`) and the second variable is needed to configure Munki's `supported_architectures` array with the correct value (`x86_64`).

If Munki understands that `amd64` is really just another way to specify `x86_64` then those AutoPkg recipes can be simplified to only require a single `amd64` variable for both the download and the `supported_architectures` array.

I think this recipe is a great example of where two variables could be reduced to one with this change:
https://github.com/autopkg/gregneagle-recipes/blob/master/Podman/Podman.munki.recipe

Instead of needing to configure both `GITHUB_ARCH` as `amd64` for the download _AND_ `MUNKI_ARCH` as `x86_64` for the array; a single variable of `ARCH` as `amd64` could be configured for both the download and array.

gregneagle:
I'm going to pass on this. I've seen vendors use "Intel" and "Mac" and "Mac_Intel" and other strings to identify their Intel-compatible builds, and "M1" and "AppleSilicon"  and "aarch64" and other strings to identify their Apple silicon-compatible builds. If we start modifying _Munki_ to accept all these alternative strings, it will just confuse Munki admins and make for really messy code. It's possible to address this problem without changing Munki in autopkg recipes, and existing recipes do this. Since vendors can always find new and unique ways to make our lives challenging, I'd rather keep the complexity in autopkg, where an admin can react to vendor oddities without needing a code change in Munki.

kevinmcox:
Gotcha, thanks for the discussion. I'll look for a solution in AutoPkg.

PR #1222, opened 2024-06-30
Run swiftlint to see what changed

natewalck:


gregneagle:
Thanks for this, but I've continued to make unpushed changes and it would be too difficult to merge everything at this point. I need to wrap my head around this stuff soon, but not today.
(And munkitester should not be linted or formatted: it's just a convenient target for me to test code snippets)

gregneagle:
I looked and decided I could merge it...

PR #1223, opened 2024-07-18
Update make_munki_mpkg.sh to set no restart for launchd component package 

rrenstrom:
The munkitools_launchd.pkg component package in Munki 6.6 beta 1 is still built with a postinstall-action="restart"

This fix modifies make_munki_mpkg.sh to build it with norestart to reflect changes in Munki 6.6 to not require a restart when updating the launchd items.

### DIFF
--- a/code/tools/make_munki_mpkg.sh
+++ b/code/tools/make_munki_mpkg.sh
@@ -102,7 +102,7 @@ do
         "A")
             AUTORUNPKG=YES
             ;;
-        "R") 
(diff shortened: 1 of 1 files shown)


gregneagle:
Good catch. Assuming the new launchd helper sticks around, there will be really no need to split out the sub packages any longer --  the primary reason for doing this in the past was to avoid the need for a logout or restart when updating the Munki tools.

PR #1224, opened 2024-08-14
Add --show-config-plist option

grahamgilbert:
This adds an option to `managedsoftwareupdate` that will allow the operator to display the current Munki configuration as a plist.


gregneagle:
Not sure what I think about this output:
```xml
	<dict>
		<key>preference</key>
		<string>ClientCertificatePath</string>
		<key>source</key>
		<string>not set</string>
		<key>value</key>
		<string>None</string>
	</dict>
```

grahamgilbert:
That was my thinking, and I was trying to mirror the existing non machine readable format as much as possible

PR #1236, opened 2025-02-04
Add Chain Support in Client Certificate Authentication

### LINKED ISSUE #1235: Client Certificate Validation does not walk chain to find a compatible certificate
### Current Behaviour: 

The current client certificate code within `gurl.py` in `munkilib` uses this code, to validate whether a Certificate can be used to authenticate with the server.

https://github.com/munki/munki/blob/b720ed5e592cf7e4bb64d4dc25bff0925e4b6354/code/client/munkilib/gurl.py#L714-L717

This code currently doesn't walk up the Issuing Chain, so if a server only returns the accepted root certificates, Munki refuses to use a certificate to authenticate with the server, despite it being valid for authentication. 

### Expected Behaviour: 

It's expected that when validating whether a certificate can be used for authentication, that the chain is walked to check for accepted certificates somewhere within the chain, and then use the certificate for authentication.

liamn:
Fixes #1235 

When using a Client Certificate, Munki didn't walk the Certificate Chain to check whether a certificate was supported. This should allow Munki to be used in organisations with larger PKI Infrastructures and should allow Munki to work with mTLS Authentication without Identity Preferences being configured, such as when certificates are distributed through SCEP.

For example, imagine the following chain: 

```
Root
   Issuing
      ClientCert
```

If the server presented only the Root Cert as trusted, then Munki would not return the certificate - as it's expecting the issuing cert to be returned in the list of trusted CAs. 

Additionally, it then didn't pass the chain into `NSURLCredential.credentialWithIdentity_certificates_persistence_` - which meant although it was supported, it wasn't actually used - leading to unknown CA errors. 

I've attempted to keep this as change free to cases where there isn't a chain on the machine, and it should retain backward compatibility.

### DIFF
--- a/code/client/munkilib/gurl.py
+++ b/code/client/munkilib/gurl.py
@@ -20,6 +20,7 @@ Created by Greg Neagle on 2013-11-21.
 Modified in Feb 2016 to add support for NSURLSession.
 Updated June 2019 for compatibility with Python 3 and PyObjC 5.1.2+
 Updated May 2022 for compatibilty with PyObjC 8.5 on macOS Mojave
+Updated Feb 2025 to add support for Certificate Chains when using Client Certificates
 
 curl replacement using NSURLConnection and friends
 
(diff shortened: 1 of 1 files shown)


gregneagle:
Thank you for this. I have no way of testing this since I don't use client cert authentication. I'll post on some of the public discussion forums to see if anyone can test. We need to test two things:
1) Your changes don't break any existing users of client cert authentication, and
2) Your changes solve the problem you set out to solve. 
It's reasonable to think that your changes solve the problem for _you_ (and that might be as good as we get), but definitely need feedback on #1.

gregneagle:
https://groups.google.com/d/msgid/munki-discuss/09493687-239D-42E8-BDAA-36F572E34FC1%40mac.com

gregneagle:
I'm going to merge this and continue to ask for testing.

gregneagle (approved):
Seems reasonable, and we've had one tester confirm these changes don't break his existing client cert config.

PR #1239, opened 2025-04-03
Ignore expired, or not-yet valid certificates

ufven:
When using mutual TLS (i.e. client certificates), Munki doesn't validate the date range of a candidate certificate. As such, it may choose one that has expired and subsequently be denied access by the server. This PR adds a fast fail approach by inspecting the `not_valid_before` and `not_valid_after` properties of the Certificate object.

### DIFF
--- a/code/client/munkilib/gurl.py
+++ b/code/client/munkilib/gurl.py
@@ -27,6 +27,7 @@ curl replacement using NSURLConnection and friends
 """
 from __future__ import absolute_import, print_function
 
+from datetime import datetime, timezone
 import os
 import xattr
 
@@ -734,6 +735,12 @@ class Gurl(NSObject):
                     continue
                 cert_data = SecCertificateCopyData(cert_ref)
                 cert = Certificate.load(cert_data.bytes().tobytes())
+
+                # do not consider certificates that are not yet valid, or have
+                # expired.
+                now = datetime.now(timezone.utc)
+                if now < cert.not_valid_before or now > cert.not_valid_after:
+                    continue
 
                 # includes the certificate issuer in the accepted subjects, 
                 # to retain pre-chain behaviour


nixtar:
My 2 cents:

I think this behaviour should be opt in, and in saying that I'm not sure it's worth exposing yet another setting?

There have been times in the past when a client cert was expired, and we had to configure the server to accept expired certs while scrambling to deploy updated certs.
The above example was for enterprise Wi-Fi auth where it introduced a race condition where the client needed network access to remediate but I could see the same be valid if you were relying on Munki to deploy/maintain the certs.

There is also the scenario where the local date time on the client is not configured correctly and a cert that would be accepted by the server gets invalidated early by the client. (However other aspects of the TLS connection may be impacted by this regardless)

IMO it's better to just rely on the server to be the authority when it comes to validating the cert.
Is this change in response to an actual issue/incident that was caused by Munki trying to use expired certs?

When I've looked this up in the past, I remember seeing that browsers exclude expired certs from the selection UI unless the expired cert is the only option. So while web browsers do have logic to try and exclude expired certs it is still possible for the user to make it use it.

As a quick and dirty test, I created a site with client cert auth and generated a self-signed cert that was intentionally expired, and the browser (Edge) still gives it as an option:
image

ufven:
> I think this behaviour should be opt in, and in saying that I'm not sure it's worth exposing yet another setting?

This is my first PR to this project so please forgive my ignorance regarding the project's development practices. I'm not sure if it follows semantic versioning or not but I would agree with you on exposing this as another setting. The rationale behind it is that this could be a breaking change for a lot of deployments if they are depending on this behaviour. As such, I'd be in favor of a setting where this is turned off by default for now but later changed to on by default when the next major version is released. In environments where (mutual) TLS is used, I would expect this to be the default behaviour, along with some other checks like Extended Key Usage (e.g., choosing a certificate with Server/Client Authentication as required.)

> There have been times in the past when a client cert was expired, and we had to configure the server to accept expired certs while scrambling to deploy updated certs. The above example was for enterprise Wi-Fi auth where it introduced a race condition where the client needed network access to remediate but I could see the same be valid if you were relying on Munki to deploy/maintain the certs.

We've had similar experiences while renewing certificate authorities, and we're using mutual TLS for a lot of things (client and server certificates), including Munki and 802.1x for network access.

> There is also the scenario where the local date time on the client is not configured correctly and a cert that would be accepted by the server gets invalidated early by the client. (However other aspects of the TLS connection may be impacted by this regardless)

This happens from time to time, and I would argue that this is out of scope for a lot of things and should be resolved properly. You'll see Kerberos and other parts of the system to fail as well.

> IMO it's better to just rely on the server to be the authority when it comes to validating the cert. Is this change in response to an actual issue/incident that was caused by Munki trying to use expired certs?

Yes, this is a consistent problem we have. All our devices renew their certificates yearly, and since we have several hundreds of clients we keep seeing this on a weekly basis. As of now, we're purging old certificates routinely but there are situations where the old certificate is still available and Munki chooses the first it can find which often is the expired one. As for the server, I would expect it to follow standard practices of denying invalid certificates. I'll be in the office in a few hours and can pull the errors we're getting from the Munki client (something about -1005 if my memory serves me right), but the server clearly denies the connection as the certificate is no longer valid.

griffer:
> There have been times in the past when a client cert was expired, and we had to configure the server to accept expired certs while scrambling to deploy updated certs. The above example was for enterprise Wi-Fi auth where it introduced a race condition where the client needed network access to remediate but I could see the same be valid if you were relying on Munki to deploy/maintain the certs.
> 

Just want to chime in real quick — I agree the setting should definitely be optional. I’d be concerned about a scenario where you lose connection to clients and have to manually fix the issue on each machine.
We’ve also run into a case where we had to allow invalid certificates on the server to get the clients back in line.

As for the suggested change, it worked as expected in my Munki test setup — no comments there.

ufven:
Thank you all for chiming in on this. 

> I’d be concerned about a scenario where you lose connection to clients and have to manually fix the issue on each machine.

I didn't consider this with my initial PR, but I fully agree that we shouldn't break production with a fix like this. I already have a new commit in the works that makes this optional with the default set to the current behaviour. The default could be changed with a major version bump, as that is where I would expect a breaking change to occour.

For completeness, here are the errors I'm observing. The client log show this:

```shell-session
$ managedsoftwareupdate -vvv
...
    Found matching identity
    Will attempt to authenticate
    Download error -1005: The network connection was lost.
    Headers: None
ERROR: Could not retrieve managed install primary manifest.: (-1005, 'The network connection was lost.')
...
```

And on the server (Apache HTTP Server), I see the following:

```
[Fri Apr 04 09:31:01.304646 2025] [ssl:error] [pid 31003] [client x.y.z.w:56461] AH02039: Certificate Verification: Error (10): certificate has expired
[Fri Apr 04 09:31:01.304692 2025] [ssl:error] [pid 31003] [client x.y.z.w:56461] AH02261: Re-negotiation handshake failed
[Fri Apr 04 09:31:01.304703 2025] [ssl:error] [pid 31003] SSL Library Error: error:14089086:SSL routines:ssl3_get_client_certificate:certificate verify failed
```

ufven:
The latest commit introduces a new preference called `AllowExpiredClientCertificate` with its default set to `True` to retain the current behaviour. I tested this on a few of my devices, and it seems to work as intended here. Would you perhaps be able to test this on your end as well? Just flip the setting with:

```shell
# Enable the new behaviour:
defaults write /Library/Preferences/ManagedInstalls AllowExpiredClientCertificate -bool FALSE
# Revert to the original behaviour:
defaults write /Library/Preferences/ManagedInstalls AllowExpiredClientCertificate -bool TRUE
```

ufven:
> Let's say the admin opts-out (or doesn't opt-in) but then gets into a situation where they need to temporarily allow expired certs.

Ideally, this would be handled through an MDM or similar to get the client compliant again, or if you could configure the server to (temporarily) allow expired client certificates during a remediation period.

> Might there be a different implementation: where instead of just using the first cert we find, we find all the possible certs, loop through them looking for one that is not expired (additional checks for other attributes could be added in the future), and if this fails, choose an expired certificate?

I suppose I could look into implementing something like that. I have done something similar in the past, where I iterate over all available certificates and sort them according to eligibility. Not sure what algorithm would be the best here, but perhaps just looking at the `not_valid_after` property and sort descending on it could work? In that case, an expired certificate could still be used if there are no other candidates. This would work for us since a valid certificate would be chosen in our environment, and an expired certificate could still be used in such a scenario. I'd like to keep it as simple as possible to avoid the code base from growing too complex, and this approach might be the least disruptive. Any thoughts on this?

rrenstrom:
> > Might there be a different implementation: where instead of just using the first cert we find, we find all the possible certs, loop through them looking for one that is not expired (additional checks for other attributes could be added in the future), and if this fails, choose an expired certificate?
> 
> I suppose I could look into implementing something like that. I have done something similar in the past, where I iterate over all available certificates and sort them according to eligibility. Not sure what algorithm would be the best here, but perhaps just looking at the `not_valid_after` property and sort descending on it could work? In that case, an expired certificate could still be used if there are no other candidates.

Seems reasonable.

gregneagle:
Hoping to see a new implementation (and then figuring out how best to add it to the Swift version: https://github.com/munki/munki/blob/swift-cli/code/cli/munki/shared/network/clientcerts.swift)

ufven:
Work required my attention elsewhere at the moment, but I will revisit this shortly. I haven't abandoned it :-)

gregneagle:
Ping

PR #1248, opened 2025-07-14
Custom sidebar localized_strings

SteveKueng:
I have played around with the localized titel. The configuration is now built in such a way that both title and page can be defined alternatively.
Also another link can be given -> here example with Google:

```
 <dict>
        <key>title</key>
        <string>My Homepage</string>
        <key>icon</key>
        <string>book.pages</string>
        <key>page</key>
        <string>https://google.com</string>
        <key>localized_strings</key>
        <dict>
            <key>de</key>
            <dict>
                <key>title</key>
                <string>Meine Webseite</string>
                <key>page</key>
                <string>https://google.de</string>
            </dict>
        </dict>
    </dict>
```

image


gregneagle:
I merged your changes, then made this addition: https://github.com/munki/munki/commit/7b815be05ea8d49f907031824e9bf8870bdab9ee

This is an attempt to more closely mimic Apple's behavior around finding localized strings. Users can specify a list of preferred languages:
image
and so we attempt to find the best match among the available localizations.

This does mean that the admin should explicitly include one for the language the "default" strings are in, like so:

PR #1257, opened 2025-08-12
Fix path traversal in load_page

HunterStanton:
This PR resolves a fairly minor/low risk path traversal issue in `load_page` by resolving both the base directory and requested path to their canonical form (`.standardizedFileURL`) and verifying via `pathComponents.starts(with:)` that the request remains within the base directory.

If this needs any changes or you think something should be different here  please let me know. I tried to keep this one short and sweet, and after fuzzing a bit to try to find some combo of obscure characters that could somehow get around this check I was unable, so this should fix this for good.


gregneagle:
This looks reasonable. I don't think the attempt to load an error page will actually work:
```
            let errorURL = baseURL.appendingPathComponent("error.html")
            webView.load(URLRequest(url: errorURL))
```
because I don't think error.html actually exists, but I'll figure out how/what to do when this problem occurs.

PR #1259, opened 2025-08-20
Attempt to download icon if icon_hash exists in pkginfo

dtheyer:
Munki minimizes unnecessary icon downloads by assuming if an icon_name does not exist in the `ICON_HASHES_PLIST_NAME` file (and the repo has that file), that the icon must not exist. 

Munki provides the ability to specify icons_name and icon_hash in the pkginfo itself for special cases where an icon is in a non-standard location and will even overwrite values in `ICON_HASHES_PLIST_NAME` if these exist. The problem is that if an icon_name is missing from the `ICON_HASHES_PLIST_NAME`, munki does not attempt to use the information from the pkginfo due to the previously mentioned optimization.

To fix this while still minimizing the number of invalid icon requests, we use the exists of an icon hash (either from the `ICON_HASHES_PLIST_NAME` file or the pkginfo) as the guard.

[CLA_MUNKI_DTHEYER.pdf](https://github.com/user-attachments/files/21874437/CLA_MUNKI_DTHEYER.pdf)

### DIFF
--- a/code/client/munkilib/updatecheck/download.py
+++ b/code/client/munkilib/updatecheck/download.py
@@ -269,10 +269,10 @@ def download_icons(item_list):
                 return
         if server_icon_hash != local_hash:
             # hashes don't match, so download the icon
-            if icon_hashes and icon_name not in icon_hashes:
-                # if we have a list of icon hashes, and the icon name is not
-                # in that list, then there's no point in attempting to
-                # download this icon
+            if not server_icon_hash:
+                # The icon hash must be in the ICON_HASHES_PLIST_NAME file 
+                # or the pkginfo itself. If it is not in either of these,
+                # then there's no point in attempting to download this icon.
                 continue
             item_name = item.get('display_name') or item['name']
             icon_url = icon_base_url + quote(icon_name.encode('UTF-8'))


gregneagle:
Any additional discussion here?

dtheyer:
Hey, sorry about the delay. I fully admit this is an edgecase we hit by doing things a strange way in our environment. Given the context you shared about there not being another Python release I am fine with abandoning this.

We are moving away from using a static munki catalog/repo to a more dynamic one which gives us more granularity in how we scope packages. As part of this we are leveraging the icon_hash and icon_name fields in the pkginfo files rather than relying on the ICON_HASHES_PLIST_NAME file.

During the migration process, we noticed that the data in the pkginfo files is ignored if there is no name entry in the ICON_HASHES_PLIST_NAME which was an unexpected behavior.

For example, it the hashes file contains the following:
```
<dict>
	<key>chrome.png</key>
	<string>612911da931e8acf18345763ab6cbc34c280c80098e412dbabff034ffc21fd7b</string>
</dict>
</plist>
```

Having a catalog entry for chrome which contains the same name but a different hash
```
...
<key>name</key> 
<string>googlechrome</string>
<key>icon_name</key> 
<string>chrome.png</string>
<key>icon_hash</key> 
<string>5b07e872f794577b1f557fb24a953af884e9314dfc0dc4697264e465675fc7eb</string>
...
```

Will allow MSC to display the chrome.png icon correctly.

On the other hand, if the name and the hash are different: 

```
...
<key>name</key> 
<string>googlechrome</string>
<key>icon_name</key> 
<string>chrome_new_icon.png</string>
<key>icon_hash</key> 
<string>5b07e872f794577b1f557fb24a953af884e9314dfc0dc4697264e465675fc7eb</string>
...
```

MSC will not display the icon since the name is not in ICON_HASHES_PLIST_NAME. This behavior struck me as a bit odd since the pkginfo is providing all of the info that MSC needs to fetch the icon. This seconds case does work if we do not serve the ICON_HASHES_PLIST_NAME file which further makes this behavior odd.

gregneagle:
I guess I'm not understanding why the icon name is not in the _icon_hashes.plist or why you would not be serving the _icon_hashes.plist file. makecatalogs puts the names of all known icons in the _icon_hashes.plist file, so if the name isn't in that file, it's because the filename doesn't exist. What am I missing here?

dtheyer:
The icon name is missing since we are migrating from an older static repo which is generated by makecatalogs to a newer dynamically generated one where the catalog is generated by an external system. Given that this is a pretty niche edgecase I can likely just add some post processing which adds the icons from the dynamic system to the _icon_hashes.plist file after makecatalogs runs.

PR #1261, opened 2025-08-26
Add YAML Support for Munki v7 Swift CLI Tools

rodchristiansen:
## What this PR proposes

This PR adds YAML support to Munki v7 Swift CLI tools as an enhancement to provide a modern, human-readable alternative to XML plists for **manifests** and **pkginfo** files. Perhaps this can be included in a future v7.1 release.

**This proposed implementation maintains complete backward compatibility while enabling admins to choose their preferred format on a per-file basis.**

### Proposed Key Feature: Mixed-Format Repository Support
This enhancement would allow administrators to:
- Keep existing XML plist files unchanged
- Create new files in YAML format using `--yaml` flag
- Set global YAML preference using macOS defaults, if ready to go all in
- Have both .plist and .yaml files in the same directory
- Migrate files gradually at their own pace
- Switch back and forth between formats as needed

### Before (XML plist):
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>name</key>
    <string>Firefox</string>
    <key>version</key>
    <string>126.0</string>
    <key>catalogs</key>
    <array>
        <string>testing</string>
        <string>production</string>
    </array>
    <key>conditional_items</key>
    <array>
        <dict>
            <key>condition</key>
            <string>os_version >= '10.15'</string>
            <key>managed_installs</key>
            <array>
                <string>ModernApp</string>
            </array>
        </dict>
    </array>
</dict>
</plist>
```

### After (YAML):
```yaml
name: Firefox
version: "126.0"
catalogs:
  - testing
  - production
conditional_items:
  - condition: os_version >= '10.15'
    managed_installs:
      - ModernApp
```

## Script Handling: Eliminates XML Escaping

One of the biggest advantages of YAML is **block scalar support** for scripts, eliminating the frustrating XML character escaping requirements.

### Before (XML plist with script):
```xml
<key>postinstall_script</key>
<string>#!/bin/bash
if [ &quot;$USER&quot; = &quot;admin&quot; ]; then
    echo &quot;Setting up &lt;special&gt; configuration...&quot;
    /usr/bin/defaults write com.example.app setting -bool true &amp;&amp; echo &quot;Success&quot;
fi
exit 0</string>
```

### After (YAML with block scalar):
```yaml
postinstall_script: |
  #!/bin/bash
  if [ "$USER" = "admin" ]; then
      echo "Setting up <special> configuration..."
      /usr/bin/defaults write com.example.app setting -bool true && echo "Success"
  fi
  exit 0
```

**Key Benefits:**
- **No character escaping needed** - `&`, `<`, `>`, quotes work naturally
- **Preserves formatting** - Indentation and line breaks maintained
- **Human readable** - Scripts can be edited directly in YAML files
- **Version control friendly** - Clean diffs show actual script changes

### Real-World Script Example

YAML transforms script handling from frustrating to natural:

**XML Plist (before):**
```xml
<key>postinstall_script</key>
<string>#!/bin/bash
# Configure application settings
if [ &quot;$3&quot; != &quot;/&quot; ]; then
    echo &quot;Installing to non-root volume: $3&quot;
fi

# Set permissions &amp; ownership
/usr/sbin/chown -R root:admin &quot;$3/Applications/MyApp.app&quot;
/bin/chmod 755 &quot;$3/Applications/MyApp.app&quot;

# Configure defaults
/usr/bin/defaults write /Library/Preferences/com.company.myapp EnableFeature -bool true
if [ $? -eq 0 ]; then
    echo &quot;Configuration complete&quot;
else
    echo &quot;Error: Configuration failed&quot; &gt;&amp;2
    exit 1
fi

exit 0</string>
```

**YAML (after):**
```yaml
postinstall_script: |
  #!/bin/bash
  # Configure application settings
  if [ "$3" != "/" ]; then
      echo "Installing to non-root volume: $3"
  fi

  # Set permissions & ownership
  /usr/sbin/chown -R root:admin "$3/Applications/MyApp.app"
  /bin/chmod 755 "$3/Applications/MyApp.app"

  # Configure defaults
  /usr/bin/defaults write /Library/Preferences/com.company.myapp EnableFeature -bool true
  if [ $? -eq 0 ]; then
      echo "Configuration complete"
  else
      echo "Error: Configuration failed" >&2
      exit 1
  fi

  exit 0
```

**Character escaping comparison:**
- No more `&quot;` just use `"`
- No more `&amp;` just use `&`
- No more `&gt;` just use `>`
- No more `&lt;` just use `<`

Clean, readable bash that you can copy/paste and test directly!

### New Files Added:
- `shared/yamlutils.swift` - Core YAML parsing and serialization functions
- `yaml_migrate/` - Complete CLI tool for migrating existing repositories
- Sample YAML files for testing and validation

### Files Enhanced:
- `Package.swift` - Added Yams v6.0.2 dependency (latest stable)
- `Package.resolved` - Updated dependency resolutions
- CLI tools (`makepkginfo`, `makecatalogs`, `munkiimport`, `manifestutil`) - Adds YAML support and global preferences

### Build Validation
```bash
$ cd code/cli/munki
$ swift run yaml_migrate --help

OVERVIEW: Migrate Munki repository files from plist to YAML format
USAGE: yaml_migrate <repo-path> [OPTIONS]

OPTIONS:
  --manifests-only        Process only manifest files
  --pkginfo-only         Process only pkginfo files
  --backup/--no-backup   Create backup copies (default: enabled)
  --dry-run              Preview changes without applying them
  --force                Overwrite existing YAML files
  -v, --verbose          Show detailed progress information
  -h, --help             Show help information
```

**Result:** Builds cleanly with comprehensive CLI interface

### Functional Testing
```bash
$ swift run yaml_migrate test_repo --dry-run -v

Starting YAML migration...
Repository: test_repo
Backup: enabled
Dry run: enabled
Processing directory: test_repo/manifests
YAML file already exists, skipping: test_repo/manifests/test_client.yaml
Processing directory: test_repo/pkgsinfo
YAML file already exists, skipping: test_repo/pkgsinfo/Firefox.yaml
YAML file already exists, skipping: test_repo/pkgsinfo/Photoshop.yaml

Migration Summary:
Files processed: 0
Files skipped: 3 (existing YAML files)
Errors: 0
```

**Result:** Correctly handles mixed-format repositories and existing YAML files

### Data Integrity Validation
Tested with complex Munki structures:
- Manifests with conditional_items, managed_installs arrays
- Pkginfo files with all standard Munki fields
- Complex nested dictionaries and arrays
- Special characters, version strings, large numbers
- Round-trip conversion preserves all data structures

Sample output maintains data fidelity and complex manifest structures:

```yaml
catalogs:
- testing
- production
conditional_items:
- condition: os_version >= '10.15'
  managed_installs:
  - ModernApp
managed_installs:
- Firefox
- Chrome
- Office
optional_installs:
- DevTools
```

## Global YAML Preference Support

The **global YAML preference** eliminates the need to specify `--yaml` for every command. When enabled, all CLI tools default to YAML output format.

### Setting the Preference

```bash
# Enable YAML as default format for all CLI tools
defaults write com.googlecode.munki yaml -bool true

# Disable YAML default (return to plist format)
defaults write com.googlecode.munki yaml -bool false

# Check current setting
defaults read com.googlecode.munki yaml
```

### Command Priority Order
1. **Explicit flag** - `--yaml` always takes precedence over global preference
2. **Global preference** - `yaml` defaults setting when no explicit flag provided
3. **Default behavior** - Falls back to plist format if no preference set

### Example Usage

**With global preference enabled:**
```bash
defaults write com.googlecode.munki yaml -bool true

makepkginfo Firefox.pkg > Firefox.yaml    # Uses YAML (global default)
munkiimport Chrome.dmg                     # Creates YAML pkginfo
manifestutil create-manifest lab_machines  # Creates YAML manifest
```

## Dependencies

- **Yams v6.0.2** - Latest stable Swift YAML library (industry standard)
- **swift-argument-parser v1.6.1** - Updated to latest stable version

Both dependencies are:
- Latest stable releases with active maintenance
- No known security vulnerabilities
- No conflicts with existing dependencies

## Backward Compatibility Guarantee

- **All existing plist workflows continue unchanged**
- **Mixed repositories fully supported** (XML and YAML in same directory)
- **No changes to existing CLI tool behavior for XML**
- **Migration is completely optional**
- **All operations are reversible** (YAML can convert back to plist)

### 1. Build and Verify Installation
```bash
cd code/cli/munki
swift run yaml_migrate --help
```

### 2. Test Migration on Sample Repository

First, perform a dry-run to preview changes:

```bash
swift run yaml_migrate /path/to/munki/repo --dry-run -v
```

Then perform the actual migration with backup enabled:

```bash
swift run yaml_migrate /path/to/munki/repo --backup -v
```

### 3. Verify Data Integrity
- Check converted YAML files are human-readable
- Confirm all data structures preserved correctly
- Test that tools can read both formats seamlessly

### 4. Test Mixed-Format Repositories
- Place both .yaml and .plist files in same directory
- Verify all Munki tools continue to work normally with both formats
- Test that manifest references work regardless of source file format
- Confirm catalog processing handles mixed pkginfo formats correctly

### File Format Detection
Automatic detection by file extension enables seamless mixed-format repositories. Files are processed by existing plist utilities or new YAML utilities based on their extension:

```
repo/
├── manifests/
│   ├── lab_machines.plist
│   ├── office_macs.yaml
│   └── mobile_devices.yaml
└── pkgsinfo/
    ├── Firefox.plist
    ├── Chrome.yaml
    └── Office365.yaml
```

The format mapping works as follows:
- `.plist` files are processed by existing plist utilities
- `.yaml` (or `.yml`) files are processed by new YAML utilities
- Mixed directories are supported where both formats can coexist in the same directory
- No configuration needed, works transparently

### Error Handling
The implementation provides comprehensive error handling:
- **Malformed files** are reported but don't stop batch processing
- **Permission issues** generate clear error messages with suggested fixes
- **Interrupted operations** support graceful recovery with detailed status
- **Invalid paths** provide user-friendly error reporting

### Memory Efficiency
The implementation is designed for optimal performance:
- **Streaming implementation** handles large files without memory issues
- **Minimal overhead** provides performance comparable to existing plist processing
- **Batch operations** enable efficient processing of entire repositories

### Immediate Benefits
- **Improved readability** making configuration files easier for humans to read and edit
- **Better diffs** where Git shows cleaner, more meaningful changes for YAML files
- **Reduced verbosity** as YAML is more concise than XML plists
- **Enhanced tooling** with better editor support, syntax highlighting, and validation

### Team Workflow Improvements
- **Easier onboarding** where new team members can understand YAML configurations faster
- **Reduced errors** through human-readable format that reduces configuration mistakes  
- **Better code reviews** making configuration change reviews more effective
- **Script editing liberation** eliminating XML character escaping for embedded scripts
- **Clean version control** where YAML block scalars show actual script content in diffs

## Test Coverage Matrix
| Component | Status | Coverage |
|-----------|--------|----------|
| YAML Parsing | Complete | 100% |
| YAML Serialization | Complete | 100% |
| File Detection | Complete | 100% |
| Migration Tool | Complete | 100% |
| CLI Interface | Complete | 100% |
| Error Handling | Complete | 100% |
| Global Preferences | Complete | 100% |

## Third-Party Tool Compatibility Impact
**Important note: Third-party tools will require updates to work with repositories containing YAML files.**

- **[MunkiAdmin](https://github.com/hjuutilainen/munkiadmin)** - Will require adding support for YAML read/write operations for manifest and pkginfo editing to work with mixed-format repositories
- **[MunkiWebAdmin2](https://github.com/munki/mwa2/tree/py3)** - Will need YAML parsing/serialization capabilities added to the Python backend to handle YAML files

I would be happy to work on adding support for YAML to 3rd party Munki tools, I like it that much :)


rodchristiansen:
When Yams encounters a Swift `Date` object, it automatically serializes it to ISO 8601 format `2025-08-26T07:07:03.651Z` and back. I believe its JSON that has no native date support and requires string parsing.

gregneagle:
Hey Rod:

You are right in guessing that I have no appetite for a PR of this size and complexity until sometime _after_ the Munki 7 release.

In the meantime, some concerns off the top of my head:

- Reliance on a third-party yaml library/package. So far, the Swift Munki has only used Swift packages from _Apple_. I'm vaguely uncomfortable with using third-party packages, though perhaps my discomfort is unwarranted. I worry about open source licensing and what happens if the package maintainer abandons the package or doesn't keep it otherwise maintained.
- Chicken-and-egg: 
  - admins using MunkiAdmin or MunkiWebAdmin2 (that would include _me_) can't really migrate to a YAML-based repo until these tools are updated to handle YAML-formated files.
  - You'd have to make sure _every single client_ you have was updated to a newer version of the Munki tools before you could transition your repo
- Mixed-format repos. You write "Automatic detection by file extension enables seamless mixed-format repositories. Files are processed by existing plist utilities or new YAML utilities based on their extension". But my manifest's filenames don't _have_ extensions, and renaming them to add file extensions would break every single existing Munki client. My pkginfo files also are largely missing ".plist" extensions, but renaming them would not be disruptive.

Another question. You write "When Yams encounters a Swift Date object, it automatically serializes it to ISO 8601 format 2025-08-26T07:07:03.651Z and back.". How does yaml distinguish between a date and a string that just happens to look like an ISO 8601 formatted date? (I don't think in practice that will be a big issue with Munki, but still curious)

I'm pretty impressed with the effort made here. I'm hoping we can figure out a way to incorporate your ideas in the future, but I cannot promise this will be merged.

rodchristiansen:
Hi @gregneagle, thanks for reviewing the PR and feedback.
I figured Munki v7.0 feature set is locked in.
Thought I'd at least see if I could get the ball rolling on this.

**- Chicken-and-Egg Problem**
If the implementation approach looks sound and reasonable from your perspective, I can work on PRs for YAML support for both MunkiAdmin and MunkiWebAdmin2 **first**, before Munki itself adds it.

**- Mixed-Format Repository Issue**
Yup... I completely didn't consider manifests and pkginfo files without extensions.

How about content-based detection instead of extension-based? Check first few lines for YAML indicators such as the `---` header, or a general YAML-style formatting and fallback to plist parsing if at all ambiguous?

**- Date vs String Distinction**
YAML behavior for date handling:

- `actual_date: 2024-06-20T12:00:00Z` → `Date` object (auto-converted)
- `quoted_date_string: "2024-06-20T12:00:00Z"` → `String` (explicit string)
- `plain_string: this-is-not-a-date` → `String` (not date-like)

YAML auto-converts ISO 8601 patterns to Date objects **UNLESS** explicitly quoted.

**- Third-Party Library Dependency (Yams)**
Could make YAML support optional via compile-time flags. The library is self-contained with no external dependencies.

Info on it:
- **Maintainer**: JP Simard (former Apple employee, maintains SwiftLint)
- **Maturity**: 1,100+ stars, 6+ years active development
- **License**: MIT (very permissive)
- **Usage**: Used by major Swift projects including SwiftLint, Vapor
- **Dependencies**: None beyond Swift stdlib
- **Size**: Small and focused (just YAML parsing)
- **Apple Alternative**: None - Apple doesn't provide YAML support

Let me know your further thoughts, thanks!

gregneagle:
I do think you'll need to do content-based detection. Possible strategies:
  - Just try to parse the file as one format; if it fails, try to parse as the other format (which one you try first could depend on the value of the "global YAML preference")
  - Read the first line of the file. If it starts with "<?xml", process as a plist. (Some testing might need to be done to see how permissive Apple's plist parsing code is; it's possible a plist can parsed without that header...)

rodchristiansen:
Bringing in info from the Slack Munki channel [thread](https://macadmins.slack.com/archives/C04QVPFGU/p1756229597312949) for context.

### Adoption Concerns
- **Kevin M. Cox**: Acknowledged the "neat idea and well thought out PR" but noted potential adoption challenges due to third-party tools that would need updates
- **Allister Banks**: Raised concerns about "widening the tent" potentially opening doors to other format requests
- **BK**: supported it as long as it remains strictly backward compatible, defaults to plist, and is reversible

### Maintainer Burden & Support Complexity
- **elios** highlighted key concerns:
  - Additional support burden and documentation duplication (similar issues seen in AutoPkg)
  - Mac admins cannot avoid plist syntax (it's everywhere in macOS for 25+ years)
  - Worry about JSON support requests following YAML request here

### YAML Technical Considerations
- **Indentation criticality**: Confirmed indentation is vital in YAML, similar to Python
- **Quoting flexibility**: Optional single/double quotes can be coded, with strong linting in makecatalogs
- **Error handling**: Enhanced makecatalogs catches YAML formatting errors effectively

### Proposed Path Forward
- **Rod's commitment**: jokingly offered to handle all future YAML-related questions and issues
- **Documentation responsibility**: Rod volunteered to help with documentation
- **Dogfooding approach**: Rod planned to use YAML implementation with his team for several months and report findings
- **Ecosystem development**: Proposed creating PRs for MunkiAdmin and MunkiWebAdmin2 YAML support first

### Key Insights
- **New admin on-ramp**: YAML could provide easier entry point for new Munki administrators
- **Ecosystem compatibility**: Need to consider broader tool ecosystem beyond core Munki tools
- **Backward compatibility**: Critical requirement for any implementation

rodchristiansen:
### Recent Commits Since PR Opening

**[85f2c55](https://github.com/rodchristiansen/munki/commit/85f2c55) - Refactor plist reading functions to use detectFileContent**
- Implements content-based file format detection
- Addresses mixed-format repository concerns

**[b71b090](https://github.com/rodchristiansen/munki/commit/b71b090) - Smarter pkginfo handling with file format detection for extensionless files**
- Enhanced handling of pkginfo files without extensions
- More robust format detection for existing repositories

**[f6c1f69](https://github.com/rodchristiansen/munki/commit/f6c1f69) - Adds parsing approach content detection instead for extensionless manifests**
- Should solve manifest file handling without requiring file extensions - tests indicate so
- Maintains compatibility with extensionless manifests and existing Munki clients

**[e9e7127](https://github.com/rodchristiansen/munki/commit/e9e7127) - Add YAML support to `makecatalogs`**
- Enhanced YAML serialization handling NSDate, NSData, URL objects
- `--yaml` command line flag for makecatalogs
- Should be fully backward compatible with existing plist/xml workflows

rodchristiansen:
Fixed. Branch is back to correct last branch commit from Sept 2nd: 
[e9e7127](https://github.com/rodchristiansen/munki/commit/e9e7127) - Add YAML support to makecatalogs

gregneagle (commented):
There's a lot to digest here in this review but I suppose that's only fair :-)

rodchristiansen:
Sweet thanks for all the notes and comments.

The MunkiShared library and the `public` code is due to me using Swift Package Manager with `Package.swift` instead what I should have done which is follow current codebase that has Xcode `munki.xcodeproj/project.pbxproj` that targets directly from /shared and does not need `public` statements.

Let me address them in follow up commits along with the other minor things that are not YAML related.. 

I also have a commit from last week in the queue that adds a `convert` subcommand to `makepkginfo` and `manifestutil` for conversion between YAML and plist formats

.gitignore
```diff
@@ -16,6 +16,7 @@ Python.framework
 # ignore paths related to building swift munki
 code/build
 code/cli/munki/build
+.build
```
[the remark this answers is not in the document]
rodchristiansen: Reverted in [78d52ff](https://github.com/munki/munki/pull/1261/commits/78d52ffc09ad62966819e775613b3ab33c1ee560) ✔︎

rodchristiansen (commented):
PR diff should be now at a state that is exclusively focused on the YAML support additions only to be reviewed, scrutinized, and tested without extra fluff  :-)

code/cli/munki/shared/admin/makecatalogslib.swift
```diff
@@ -20,28 +20,36 @@
...
     var pkgsinfoList: [String]
     var pkgsList: [String]
     var catalogs: [String: [PlistDict]]
-    var errors: [String]
-    var warnings: [String]
+    public var errors: [String]
+    public var warnings: [String]
 
-    init(repo: Repo,
+    public init(repo: Repo,
```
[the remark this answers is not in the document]
rodchristiansen: Resolved in [2e225be](https://github.com/munki/munki/pull/1261/commits/2e225bec502d8dc165a66484cc4ef7c7ef612657) :-)

The MunkiShared library and the `public` statements were due to me using Swift Package Manager with `Package.swift` instead what I should have done which is follow codebase using Xcode `munki.xcodeproj/project.pbxproj` that targets libraries directly from `/shared` and does not need public statements.

code/cli/munki/shared/admin/makecatalogslib.swift
```diff
@@ -274,7 +285,7 @@ struct CatalogsMaker {
     /// Assembles all pkginfo files into catalogs.
     /// User calling this needs to be able to write to the repo/catalogs directory.
     /// Returns a list of any errors it encountered
-    mutating func makecatalogs() async {
+    public mutating func makecatalogs() async {
```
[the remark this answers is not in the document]
rodchristiansen: Resolved in [2e225be](https://github.com/munki/munki/pull/1261/commits/2e225bec502d8dc165a66484cc4ef7c7ef612657) :-)

The MunkiShared library and the `public` statements were due to me using Swift Package Manager with `Package.swift` instead what I should have done which is follow codebase using Xcode `munki.xcodeproj/project.pbxproj` that targets libraries directly from `/shared` and does not need public statements.

test_yaml_repo/catalogs/testing.yaml
```diff
@@ -0,0 +1,27 @@
...
+  installer_item_hash: 3d2f1a7b8c9e0f4a5b6c7d8e9f0a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8
+  installer_item_location: apps/Firefox-91.0.2.dmg
+  installer_type: dmg
+  installs:
+  - CFBundleIdentifier: org.mozilla.firefox
+    CFBundleShortVersionString: 91.0.2
+    CFBundleVersion: '20210818153112'
+    path: /Applications/Firefox.app
+    type: application
+  minimum_os_version: '10.12'
```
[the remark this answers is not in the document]
rodchristiansen: Even better fix by adding type-flexible string extraction for YAML values in: https://github.com/munki/munki/pull/1261/changes/d8de254df10f58c074f59eabcf62daa6f510d205

rodchristiansen:
**Production usage update**

Wanted to share that we've been running a fully YAML-based Munki repo in production since early December, so for about two months now.

- 1,466 pkgsinfo 
- 1,514 manifests (we use a waterfall manifest structure with a lot of `included_manifests`)

One manifest kept intentionally as `.plist` so older clients can still find new `MunkiTools` that supports yaml and updated our `MunkiPreflight` pkg (both pkginfo still `.plist` as well) to update their `preflight` which is what gets clients to switch `ClientIdentifier` to new .yaml manifest.

The minor issues we had initially were fix in the branch commits here.

The conversion was done using the `makepkginfo convert --to-yaml` and `manifestutil convert --to-yaml` bulk commands. 

**Ecosystem PRs submitted:**

- **MunkiAdmin**: [hjuutilainen/munkiadmin#225](https://github.com/hjuutilainen/munkiadmin/pull/225) — YAML read/write for pkgsinfo and manifests, format auto-detection and preservation
- **MunkiReport** (3 modules): [munkireport#14](https://github.com/munkireport/munkireport/pull/14), [munkiinfo#7](https://github.com/munkireport/munkiinfo/pull/7), [managedinstalls#11](https://github.com/munkireport/managedinstalls/pull/11) — adds `DataParser` class for auto-detecting plist vs YAML in client report data

**Still to do:**

- MunkiWebAdmin2 YAML support
- AutoPkg YAML output support
- Wiki documentation for YAML usage (I have drafts locally)
- Other ecosystem tools as they come up I can try to contribute to.

Happy to answer questions or provide more detail. 
Admins, me included, really like the readability of YAML and the git integration has been smooth.

gregneagle:
Thanks, Rob.

I do want to get back to looking at this, hopefully after the Munki 7.1 release. My biggest concern currently is adding a dependency on the third-party Yams Swift package. There would need to be plan for what to do should that package get abandoned, pulled, or its development went in a direction we could no longer support.

Munki 7 does currently have some external library/package dependencies, but so far, they are all packages developed and supported by _Apple_.

gregneagle:
As Munki 7.1 gets closer to release I'm thinking about this PR more.

The additional Swift packages used by Munki currently -- swift-argument-parser, swift-asn1, swift-certificates,  and swift-crypto --  are not only from _Apple_, but all are licensed under the Apache 2 license -- the same as Munki itself. You noted that the Yams package is MIT-licensed. I have no idea how mixing licenses works for an open-source project. This is something we'd need to clarify before moving forward.

I'd spent some time yesterday looking at Apple's PKL project. pkl-swift is Apache 2 licensed. pkl can read and emit both plist and yaml. But it looked incredibly non-trivial to implement as something to simply read and write dictionaries (and lists of dictionaries) as plists or yaml. I haven't 100% given up on it, but it doesn't look terribly promising right now.

code/cli/munki/shared/utils/yamlutils.swift
```diff
@@ -0,0 +1,836 @@
...
+    case let nsDictionary as NSDictionary:
+        var result: [String: Any] = [:]
+        for (key, value) in nsDictionary {
+            // Skip NSNull values and null-like values - don't include them in output
+            if isNullValue(value) { continue }
+            let sanitizedValue = sanitizeForYaml(value)
+            // Also skip if sanitization returned the null sentinel
+            if let str = sanitizedValue as? String, str == "__YAML_NULL_SENTINEL__" { continue }
+            // Skip empty strings - they serialize ambiguously in YAML and get interpreted as null
+            if let str = sanitizedValue as? String, str.isEmpty { continue }
```
homebysix: In sanitizeForYaml (yamlutils.swift:616, 654, 666), every empty string is silently dropped before writing, regardless of key. This means `description: ""` in a YAML pkginfo is lost on write and indistinguishable from a missing field on read — the intent to explicitly blank the field is not honored.

code/cli/munki/shared/utils/yamlutils.swift
```diff
@@ -0,0 +1,836 @@
...
+    } else {
+        return object
+    }
+}
+
+
+/// Attempt to read YAML from data
+func readYaml(fromData data: Data) throws -> Any? {
+    do {
+        let yamlString = String(data: data, encoding: .utf8) ?? ""
```
homebysix: **Bug: binary plist silently reads as nil under `preferYaml: true`**

`String(data: data, encoding: .utf8) ?? ""` coerces a failed UTF-8 decode (e.g. a binary plist) into an empty string. `Yams.load(yaml: "")` then succeeds returning `nil` — no throw — so `readData`'s catch block never fires and the plist fallback is skipped. The caller gets `nil` with no error or warning.

Fix: replace the `?? ""` with a guard-throw:
```swift
guard let yamlString = String(data: data, encoding: .utf8) else {
    throw YamlError.readError(description: "Data is not valid UTF-8")
}
```
That makes `readData` catch the error and fall through to `PropertyListSerialization` as intended.

.github/ISSUE_TEMPLATE/bug_report.md
homebysix, line 1: Was this deletion intentional? Seems unrelated to yaml support.

.github/ISSUE_TEMPLATE/feature_request.md
homebysix, line 1: Was this deletion intentional? Seems unrelated to yaml support.

code/cli/munki/shared/utils/yamlutils.swift
```diff
@@ -0,0 +1,836 @@
...
+/// Without this normalization, the version check in catalogs.swift would fail:
+///   if let minimumOSVersion = item["minimum_os_version"] as? String
+/// The cast would return nil for a float, silently bypassing the OS version check.
+///
+/// This ensures version-related fields are always strings, matching munki's expectations.
+func normalizeYamlTypes(_ object: Any?) -> Any? {
+    guard let object = object else { return nil }
+    
+    // Known keys that should always be strings, even if they look like numbers
+    let stringKeys: Set<String> = [
```
homebysix, line 479: `stringKeys` is a hidden contract with no compiler enforcement, and a future contributor adding a new version-like key could easily break things. That said, it doesn't appear to break things yet.

code/cli/munki/shared/updatecheck/manifests.swift
```diff
@@ -95,44 +101,95 @@ func getManifest(_ name: String, suppressErrors: Bool = false) throws -> String
...
+            }
+            Manifests.shared.set(name, path: manifestLocalPath)
+            return manifestLocalPath
+            
+        } catch let FetchError.http(errorCode, description) {
+            // HTTP 404 - silently try next extension
+            if errorCode == 404 {
+                lastHttpError = (errorCode, description)
+                continue
+            }
```
homebysix: Any 404 — transient or permanent — unconditionally continues through all three extensions (`""`, `.yaml`, `.plist`) at `manifests.swift:153–155`. There's no distinction between "file doesn't exist" and "server hiccup".

Result: a single transient 404 on `site_default` produces three requests. With 5 included manifests under server load, that's up to 15 requests where there were 5. Since included manifests use `suppressErrors: true`, all intermediate attempts are completely silent — the admin sees one "Getting manifest" line per manifest and then the error, with no indication retries occurred.

code/cli/munki/manifestutil/MUmanifestFileOperations.swift
```diff
@@ -34,8 +35,10 @@ func saveManifest(repo: Repo,
         }
     }
     do {
-        let data = try plistToData(manifest)
-        try await repo.put("manifests/\(name)", content: data)
+        let fileExtension = yamlOutput ? ".yaml" : ""
```
homebysix: The duplicate-check compares the caller-supplied `name` against raw repo identifiers, which include the `.yaml` suffix when YAML manifests exist. If `foo.yaml` is in the repo and an admin runs `new-manifest foo` (without `--yaml`), `contains("foo")` is false and a second file `manifests/foo` is created silently alongside `manifests/foo.yaml`. The client then has two candidates for the same logical name.

Fix: normalize both sides of the comparison by stripping known extensions before checking, or check whether any existing name's extension-stripped form matches `name`.

code/cli/munki/shared/updatecheck/manifests.swift
gregneagle: I'm going to argue that we should not be trying any extensions at all: that the manifest name _is_ the manifest name, and if Munki tries to ask for manifest "foo" and gets a 404 it should not then try "foo.plist" and/or "foo.yaml".

code/cli/munki/manifestutil/MUmanifestFileOperations.swift
gregneagle: Again I'll argue the manifest name is the manifest name, and that we should _not_ be trying to automagically insert/detect .yaml or .plist extensions. It should be entirely legal for an admin to name a manifest that is actually an XML plist "l.love.yaml", and a yaml-formatted plist "i.hate.plist"

code/cli/munki/makepkginfo/MPIconvert.swift
```diff
@@ -0,0 +1,361 @@
...
+            
+            if verbose {
+                print("Converting all pkginfo files in \(directoryPath) to \(targetFormat) format...")
+                print("Backup: \(backup ? "enabled" : "disabled")")
+                print("Dry run: \(dryRun ? "enabled" : "disabled")")
+            }
+            
+            var stats = ConversionStats()
+            
+            // Get all files in directory (non-recursive for now)
```
homebysix: The comment says "non-recursive for now" but `FileManager.enumerator` with no `skipsSubdirectoryDescendants` option is fully recursive. The recursive behavior is correct for real pkgsinfo repos (which are typically organized in subdirectories), so the comment should be updated.

code/cli/munki/shared/updatecheck/manifests.swift
gregneagle: I point to this documentation: https://github.com/munki/munki/wiki/Default-Manifest-Resolution

Some excepts with commentary:

> If ClientIdentifier is defined and non-empty, Munki asks for a manifest named after the ClientIdentifier:
> http://webserver/repo/manifests/SomeArbitraryIdentifier
> If that fails, no further requests are made and the managedsoftwareupdate run is ended with an error.

Nothing here about retrying with additional file extensions.

> If ClientIdentifier is undefined or empty, Munki asks for a manifest named after the fully-qualified hostname of the machine:
> http://webserver/repo/manifests/hostname.mycompany.com
> If that fails, it tries the short hostname:
> http://webserver/repo/manifests/hostname
> If that fails, it tries the machine serial number. If the serial number is "AABBCCDDEEFF"; the request looks like:
> http://webserver/repo/manifests/AABBCCDDEEFF
> If that fails, it tries to retrieve a manifest named, literally, "site_default":
> http://webserver/repo/manifests/site_default

Again: nothing about retrying with additional file extensions.

rodchristiansen:
In regards to the mixed licensing @gregneagle some examples I found:

- SwiftNIO (Apple project) has a [NOTICE.txt](https://github.com/apple/swift-nio/blob/main/NOTICE.txt) documenting mixed uses of MIT, BSD-3, and Apache 2.0 components

- Kubernetes is [Apache 2.0](https://github.com/kubernetes/kubernetes/blob/master/LICENSES/LICENSE) with mixed of 3rd party MIT dependencies such as [gimme](https://github.com/kubernetes/kubernetes/blob/master/LICENSES/third_party/gimme/LICENSE) and [vishhstress](https://github.com/kubernetes/kubernetes/blob/master/LICENSES/third_party/forked/vishhstress/LICENSE)

Apache Software Foundation's legal [policy](https://www.apache.org/legal/resolved.html#category-a) has MIT/X11 in Category A ("What Can We Include In An ASF Project?"):

> For inclusion in an Apache Software Foundation product, we consider the following licenses to be similar in terms to the Apache License 2.0

**No copyleft in either license.** Unlike GPL, neither MIT nor Apache 2.0 requires the combined/derivative work to adopt the dependency's license. Munki remains Apache 2.0.

**MIT is less restrictive than Apache 2.0.** MIT's only requirement is preserving the copyright notice and license text. Apache 2.0 adds requirements (NOTICE file, patent grant, state changes), but these don't conflict with MIT.

(I'm learning about all this as I go here)

gregneagle:
@rodchristiansen Thanks for looking into the licensing issues.

.github/ISSUE_TEMPLATE/bug_report.md
rodchristiansen: Accidental! Restored.

.github/ISSUE_TEMPLATE/feature_request.md
rodchristiansen: Accidental! Restored.

code/cli/munki/shared/updatecheck/manifests.swift
rodchristiansen: Of course, the extension fallback was wrong on many levels. The manifest name is the manifest name. Removed/reverted!

code/cli/munki/manifestutil/MUmanifestFileOperations.swift
rodchristiansen: Removed the automatic .yaml extension append in saveManifest — the `--yaml` flag only controls the serialization format, not the filename. The duplicate-check issueraised goes away as a consequence since the name is now stored exactly as supplied. Resolved in [de0eec8](https://github.com/munki/munki/pull/1261/commits/de0eec8f7a4ab8237da863def93e2bbf693407fa).

code/cli/munki/makepkginfo/MPIconvert.swift
rodchristiansen: Yea the comment was wrong. The enumerator is indeed fully recursive, which is the correct behavior for pkgsinfo repos with subdirectories. Thanks~

code/cli/munki/shared/utils/yamlutils.swift
rodchristiansen: Added comments to `stringKeys` that any new version-like key added to Munki's pkginfo schema needs to be added there

code/cli/munki/shared/utils/yamlutils.swift
rodchristiansen: Thank-you.  Applied your suggested fix exactly: replaced ?? "" with a guard-throw

code/cli/munki/shared/utils/yamlutils.swift
rodchristiansen: Cool, good catch, removed the three empty-string skip lines in sanitizeForYaml and added explicit single-quoted serialization in toNode so empty strings now round-trip as description: '' instead of disappearing.

rodchristiansen:
Added a commit to fix a trailing-zero version bug in the YAML read path.

Trailing-zero versions like `minimum_os_version: 10.10` were being parsed by Yams as Double `10.1` before `normalizeYamlTypes` could see them — the trailing zero was lost at parse time and couldn't be recovered. "10.10" and "10.1" are distinct macOS versions, so this was a real semantic corruption.

Fix is a one-liner: `munkiYamlResolver = Resolver.default.removing(.float)`, passed to `Yams.load` in the three `readYaml` entry points. Float-looking scalars now arrive as Strings directly, preserving exact text. Integer/boolean resolvers untouched so `installer_item_size: 1024` and `optional: yes` still parse natively. `normalizeYamlTypes` stays as a safety net for the Int case.

Got the idea from @homebysix's loader-level approach in [autopkg#1023](https://github.com/autopkg/autopkg/pull/1023).

Added 6 test cases in `yamlutilsTests.swift`; wired Yams into the `munkiCLItesting` target. Results:

- Trailing-zero versions (10.10, 11.20, 1.0) survive as strings
- Float-like versions (10.13, 2.3, 26.4) parse as strings
- Integer versions on version-like keys normalized to strings
- Non-version integers and bools stay native-typed
- Explicitly quoted versions round-trip unchanged
- Nested versions in receipts/installs arrays preserved

These should address concerns about unquoted float points to strings @gregneagle? 
Admins can quote or not, and `yamlutil` takes care of it.

gregneagle:
This closure was accidental (due to me deleting the Munki7_3dev branch). I'll fix things and re-open this.

gregneagle:
And now, reopened and pointed at the Munki7_4dev branch; there's a fair number of conflicts to resolve :-(

rodchristiansen:
Conflicts resolved!

rodchristiansen:
Hi @gregneagle, was thinking about how this PR is so big and how to potentially get it moving

Perhaps instead of one PR make it a stack of seven smaller ones. Each layer builds on the one below it and can be reviewed and merged on its own

1. [Yams dependency + YAML reader](https://github.com/rodchristiansen/munki/compare/Munki7_4dev...yaml/1-yams-reader) (+707, mostly one new file, NOTICE, and Xcode project wiring; nothing calls it yet)
2. [YAML writer](https://github.com/rodchristiansen/munki/compare/yaml/1-yams-reader...yaml/2-writer) (+748, key ordering, block scalars for scripts, quoting of ambiguous scalars, tests)
3. [Client reads YAML manifests](https://github.com/rodchristiansen/munki/compare/yaml/2-writer...yaml/3-client-reads-yaml) (+28)
4. [Version values via `stringValue(forKey:)`](https://github.com/rodchristiansen/munki/compare/yaml/3-client-reads-yaml...yaml/4-version-string-values) (+46, defensive, droppable)
5. [Admin tools: `--yaml` and `use_yaml` pref](https://github.com/rodchristiansen/munki/compare/yaml/4-version-string-values...yaml/5-admin-tools) (+94)
6. [YAML catalogs](https://github.com/rodchristiansen/munki/compare/yaml/5-admin-tools...yaml/6-yaml-catalogs) (+22, droppable per the discussion above)
7. [`convert` subcommands](https://github.com/rodchristiansen/munki/compare/yaml/6-yaml-catalogs...yaml/7-convert-subcommands) (+703)

I open layer 1 as its own PR against Munki7_4dev (or a new Munki7_5dev?), and open each next layer once the one below it lands. This PR can stay open as the discussion thread; layer 1 would be a new PR linking back here.

I'm assuming YAML support is still of interest since this has been kept opened :-)

Would that work better?

gregneagle:
I think this approach might make sense. You're right that it should target a 7.5 release (at the earliest -- all I can be sure of is it won't be part of Munki 7.4).

It's concerning to me that you say "Conflicts resolved" here https://github.com/munki/munki/pull/1261#issuecomment-5482230086 yet I still see the message "This branch has conflicts that must be resolved"...

Issue #1279, opened 2025-10-20
Munki 7 - HTTP request headers do not include `If-Modified-Since`


sphen13:
Noticed that on Munki 7, my CDN stopped sending 304 response codes. 304 is used to tell the client that the requested file has not changed from what was previously pulled and saved bandwidth. This does work in Munki 6 and earlier.

To best illustrate and test I did a tcpdump capture to show the HTTP request headers for /icons/_icon_hashes.plist which very typically gets a 304 response.

Munki 6:
```
GET /icons/_icon_hashes.plist HTTP/1.1
Host: XXX
Connection: keep-alive
Accept: */*
If-Modified-Since: Mon, 20 Oct 2025 15:47:14 GMT
User-Agent: managedsoftwareupdate/6.6.3.4704 Darwin/25.0.0
Accept-Language: en-US,en;q=0.9
Accept-Encoding: gzip, deflate
```

Munki 7 (7.0.1.5340):
```
GET /icons/_icon_hashes.plist HTTP/1.1
Host: XXX
Accept: */*
Accept-Language: en-US,en;q=0.9
Connection: keep-alive
Accept-Encoding: gzip, deflate
User-Agent: managedsoftwareupdate/7.0.1.5340 Darwin/25.0.0
```

because Munki 7 is not including the `If-Modified-Since` header it is impossible to receive a 304.

with the change to swift i am really not sure where to look in the code to see where this needs to be fixed :/

Thank you!

PR #1281, opened 2025-10-20
Change ETag / Last-Modified fetch.swift logic

sphen13:
Potentially addresses #1279. I applied these changes and recompiled on my machine. in my extremely limited testing it appears to include the `If-Modified-Since` header properly when a file is present/cached on disk.


gregneagle:
My confusion about the changes here is that it's completely different logic from the Python version:

https://github.com/munki/munki/blob/Munki6/code/client/munkilib/fetch.py#L525-L532

sphen13:
actually now that i look at it again, i wonder if its a single character typo on line 352, should the test just be negative/opposie of what it is right now?

in python:
```
if etag:
    getonlyifnewer = False
```

in swift:
```
if eTag.isEmpty {
    getOnlyIfNewer = false
}
```

but it should just be?
```
if !eTag.isEmpty {
    getOnlyIfNewer = false
}
```

PR #1283, opened 2025-10-23
Tweak ETag / Last-Modified fetch.swift logic

sphen13:
address fetch w etag / last-modified

slightly updated logic which i think covers us better than the previous PR.


sphen13:
@gregneagle hate to raise this up again but i would hate to have any kind of error introduced by the previous PR. tested good on my end and if you can just test with your setup i think we should be good.

PR #1287, opened 2025-11-07
Use a clock unaffected by system sleep for script timeouts (Take 2)

sphen13:
Munki 7 added 60 second timeouts when running certain scripts, such as preflight/postflight and conditions. The implementation uses [Date](https://developer.apple.com/documentation/foundation/date), [TimeInterval](https://developer.apple.com/documentation/Foundation/TimeInterval) and friends, which don't handle system sleep gracefully causing spurious timeouts e.g. if a user closes the lid in the middle of a munki run.

This PR changes it to use [mach_absolute_time](https://developer.apple.com/documentation/kernel/1462446-mach_absolute_time) instead, which uses a monotonic clock that does not increment while the system is asleep.

respect to @MagerValp for really figuring this out! This is an alternative to #1286 which works fine in my testing as well - just keeping this here as i had also created this method and pushed.

### DIFF
--- a/code/cli/munki/shared/utils/cliutils.swift
+++ b/code/cli/munki/shared/utils/cliutils.swift
@@ -38,6 +38,28 @@ struct CLIResults {
     var failureDetail: String = "" // error text from this code
 }
 
+/// Tracks elapsed time via mach_absolute_time so sleep/wake doesn't skew deadlines.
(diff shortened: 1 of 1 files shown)


gregneagle:
Going to use @magervalp's PR for this issue.

PR #1307, opened 2026-01-24
Add retry on interrupted download

MagerValp:
When munki needs to fetch a resource from the repo any network interruption causes the current run to fail. This is normally not an issue as munki will try again an hour later, or the user can check again in MSC. It is however a missed opportunity and it causes a fair amount of noise in the logs. We're seeing a steady stream of errors in Grafana and MunkiReport — not a huge issue, but it raises the noise floor.

I've experimented a bit with implementing retries in network/middleware/fetch.swift:getHTTPfileIfChangedAtomically(). It adds two new pref keys:

DownloadRetries: int between 0 and 10. Defaults to 0 which is no retries and thus keeps the old behavior.
RetrySleepSeconds: int between 1 and 30. Defaults to 10 seconds.

### DIFF
--- a/code/cli/munki/shared/network/fetch.swift
+++ b/code/cli/munki/shared/network/fetch.swift
@@ -368,34 +368,69 @@ func getHTTPfileIfChangedAtomically(
             // fall through - no cached headers
         }
     }
-    var headers: [String: String]
-    do {
-        headers = try getURL(
-            url,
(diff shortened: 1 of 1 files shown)


gregneagle:
Merged into the new Munki7_1dev branch for (hopefully) eventual release in Munki 7.1. Will be doing a test build soon.

PR #1315, opened 2026-02-16
Better Danish localization for new assisted quit feature and other improvements

ttv-kglakademi-dk:


gregneagle:
Thanks for this. We're still discussing changes to the new UI around quitting blocking apps, so it's possible more changes will be needed before the official release of 7.1.

PR #1316, opened 2026-02-17
Add new quit-relaunch keys to --show-config

aysiu:
Adding new quit-relaunch–related keys specified in https://github.com/munki/munki/releases/tag/v7.1.0b1 to the `--show-config` list of preference keys.


gregneagle:
Waiting to be sure these names are final.

PR #1323, opened 2026-03-04
Update Dutch localization: translate remaining strings and reviewed a…

pnerum:
Melvin Peeters and I edited the new strings and reviewed and changed other strings as well.
We both are Dutch natives and Melvin has done most of the language edits as it's one of his many special skills ;-)

I've just build a new Managed Software Centre app and tested the new keys for the visual testing for example:
 
Scherm­afbeelding 2026-03-04 om 16 32 21


gregneagle:
Thank you!

PR #1324, opened 2026-03-04
Update Finnish localization

hjuutilainen:
Here's a quick update for the Finnish localization. Translated the missing and verified others.


gregneagle:
Thank you!

PR #1325, opened 2026-03-04
Mitigate Keychain errors when reopening apps

jordancalhoun:
### Summary
- Check login keychain accessibility before reopening apps that were closed for an update; if the keychain is in a degraded state, alert the user to reboot instead of reopening apps into a broken session
- Switch from `NSWorkspace.shared.openApplication()` to `/usr/bin/open -g` for reopening, which routes through LaunchServices for proper security session inheritance
- Stagger app launches with an initial 2-second delay and 1-second gaps between each app to give `securityd` time to settle after root-level package installations

### Background
In very rare cases, after `managedsoftwareupdate` runs as root to install packages, the security session's connection to the user's login keychain can become disrupted. When apps are immediately relaunched in this state, they fail to access keychain items, producing errors like:
image

These errors can only be resolved by rebooting. Previously, MSC would reopen all apps simultaneously via `NSWorkspace` with no keychain health check, which could put the user in an unrecoverable state until reboot.

### Changes
- **`MSCBlockingAppsController.swift`**
  - Added `import Security` for `SecKeychain` APIs
  - `reopenApps()` now checks keychain status before launching anything
  - Added `isKeychainAccessible()` — queries the default keychain via `SecKeychainCopyDefault` / `SecKeychainGetStatus` and verifies it is unlocked, readable, and writable
  - Added `showKeychainInaccessibleAlert()` — presents a warning sheet informing the user the update succeeded but a restart is needed before apps can be safely reopened. (This does not have Localization.  This should be removable once we have tested the other changes have fully resolved the issue)
  - App launches now use `/usr/bin/open -g` instead of `NSWorkspace.shared.openApplication()` and are staggered with delays

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center/Controllers/MSCBlockingAppsController.swift
+++ b/code/apps/Managed Software Center/Managed Software Center/Controllers/MSCBlockingAppsController.swift
@@ -7,7 +7,6 @@
 //
 
 import Cocoa
-import Security
 
 /// A flipped NSClipView that positions content from top to bottom.
 /// Used in scroll views to ensure content aligns to the top rather than the bottom.
@@ -899,30 +898,19 @@ class MSCBlockingAppsController: NSObject {
     /// Call this method after the update has completed.
     /// Clears the list of apps to reopen after attempting to open them.
     ///
-    /// Before reopening, checks that the user's login keychain is accessible.
-    /// Package installations running as root can occasionally disrupt the
-    /// security session's connection to the login keychain. If the keychain
-    /// is inaccessible, the user is advised to reboot instead.
-    ///
-    /// Apps are launched via `/usr/bin/open -g` (through LaunchServices) with
-    /// staggered delays to allow the security session to settle after installs.
+    /// Apps are launched with staggered delays to allow the system to settle
+    /// after package installations by `managedsoftwareupdate`.
     func reopenApps() {
         guard !appsToReopenAfterUpdate.isEmpty else { return }
-
-        // Check keychain accessibility before reopening apps
-        if !isKeychainAccessible() {
-            msc_debug_log("Keychain is not accessible after update - advising reboot")
-            showKeychainInaccessibleAlert()
-            appsToReopenAfterUpdate = []
-            return
-        }
 
         let apps = appsToReopenAfterUpdate
         appsToReopenAfterUpdate = []
 
-        // Stagger app launches to give the security session time to settle
-        // after package installations. Launch via /usr/bin/open which goes
-        // through LaunchServices for proper session inheritance.
+        let config = NSWorkspace.OpenConfiguration()
+        config.activates = false // Open apps in background without bringing to foreground
+
+        // Stagger app launches to give the system time to settle
+        // after package installations.
         let initialDelay: TimeInterval = 2.0
         let staggerDelay: TimeInterval = 1.0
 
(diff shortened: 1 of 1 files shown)


gregneagle:
I'm not understanding the underlying issue here. If a user quit and relaunched an app manually, could they encounter the same issue? It's not clear to me how/why the keychain gets into the state and how MSC is involved.

If the machine is in a bad state, perhaps we should warn the user before _anything_ happens, not after quitting apps and installing software, but before relaunching apps.

gregneagle:
Also: you say a _reboot_ is required to fix this "bad login keychain" state: a _logout and login_ is not sufficient?

jordancalhoun:
We have only had two (reported) instances of it happening so validating a logout and login hasn't been ruled out.

I believe the crux of the issue is some form of race condition with the keychain for these apps.  The keychain is in a perfectly fine state before running the updates; and the moment the apps relaunch we saw those errors.  I believe the actual fix is having apps open in a staggered pattern vs all at once.

As an aside we have seen more and more keychain issues in macOS 26 that we have in previous releases and could be the root cause.  However, we have not had any reports of this issue happening from users running MSC 7.0.  

My ideal outcome here is to have the delay re-opening implemented. I believe will prevent the issue from happening and then remove keychain logic altogether before general release.  I only added it now for better end user testing during the beta period.

I am torn on using `/usr/bin/open` vs `NSWorkspace` and would like others opinions on that change.  We have a script that helps re-open apps and it calls `open` and have not seen this issue before with it

gregneagle:
Could you share what apps you either know or suspect trigger/are affected by this issue?

jordancalhoun:
Instance 1: 1Password, Slack, VS Code

Instance 2: Spotify, Raycast 

Originally I suspected it was an electron issue but Raycast also had the issue

gregneagle:
I'd feel more comfortable merging code that did this:
1) Continued to use NSWorkspace methods to (re-)launch the apps
2) Added the initial and inter-app delays to the relaunching

I'd want to avoid the Keychain stuff until/unless we determine there are no alternatives.

I'd also point out that if as an admin you know an app is problematic, you can and should mark it with `blocking_applications_manual_quit_only` = true

jordancalhoun:
Sounds good i'll make the change.

> I'd also point out that if as an admin you know an app is problematic, you can and should mark it with blocking_applications_manual_quit_only = true

Agreed; however, we don't seem to have a common denominator on specific apps causing the issue.  So if it's a wider spread issue I'd like to address it in the logic

gregneagle:
@jordancalhoun Do you use osquery in your org? Might either of these be relevant?
https://github.com/osquery/osquery/pull/8815
https://github.com/fleetdm/fleet/issues/43313

jordancalhoun:
> @jordancalhoun Do you use osquery in your org? Might either of these be relevant? [osquery/osquery#8815](https://github.com/osquery/osquery/pull/8815) [fleetdm/fleet#43313](https://github.com/fleetdm/fleet/issues/43313)

😖 that wouldn’t be the first time they have had that issue

gregneagle:
I ask because it's possible _osquery_ is the real trigger for this issue and therefore nothing we do in Munki can address it.

jordancalhoun:
Agreed. I wonder if the install is triggering osquery to do its thing. Interestingly we haven’t ran into the issue since this change so we’re good either way

rrenstrom:
Apparently not completely solved yet. Got this after relaunch,  "A keychain cannot be found to store "Slack Key."

```
% managedsoftwareupdate --version
7.3.0.5808

% sw_vers
ProductName:		macOS
ProductVersion:		15.7.9
BuildVersion:		24G830
```

Screenshot 2026-08-31 at 4 29 05 PM


I'm adding` blocking_applications_manual_quit_only = true` as suggested work-around.

PR #1329, opened 2026-03-10
Update German localization for new 7.1 messages

a-vogel:
Apple's German translation for `Force Quit…` is `Sofort beenden …`. Translations have been updated to reflect the official wording.

The other translations are proper German except for `Skip and Update Others`. This has also been updated.


gregneagle:
Thank you for this. There will likely be some more things to translate in the next beta release.

PR #1330, opened 2026-03-17
Next batch of Danish translations

ttv-kglakademi-dk:
Removed all verified TODOs and added a couple of other better translations


gregneagle:
Thank you!

PR #1342, opened 2026-03-31
Approve out-of-date translation

a-vogel:
Translation is correct. I assume we don't distinguish between one day and several days, since this would differ in English as well.


gregneagle:
Message is not displayed until we are several days out-of-date, so we don't have to worry about 0 or 1 days.

PR #1347, opened 2026-04-09
Complete Italian translation for localizable strings

bonzini:
Also fixed "Due:" with the same translation as nl.


gregneagle:
Thank you!

PR #1360, opened 2026-05-27
Split load_page() into internal-page and arbitrary-URL variants

kevinmcox:
Alternative to #1359, implementing the "split `load_page()` into two different functions" suggestion. Splits `load_page()` into two functions with distinct intents:

- `load_page(name)` — loads a Munki-internal HTML page from htmlDir. Path traversal still blocked. Will not navigate to remote URLs.
- `load_url(url_string)` — loads an arbitrary http(s) URL. Used only by `loadSidebarItemPage` to honor admin-configured `CustomSidebarItems` with remote `page` values.

`loadSidebarItemPage` is the only caller that branches on input scheme — the other 13 call sites pass internal page names and continue calling `load_page` unchanged.

Also removes the input guard added in #1356. The function split makes that guard redundant — `handleMunkiURL` → `load_page(filename)` can no longer reach remote-loading code regardless of what `filename` contains — and keeping it alongside the split would create an asymmetry (explicit refusal log for http/https inputs, silent fall-through for other schemes that hit the same downstream code).

## Verified locally

Built the patched MSC and tested:

| URL | Result |
|---|---|
| `munki://category-all`, `munki://categories`, `munki://myitems`, `munki://updates` | Pages load correctly |
| `munki://detail-Slack` | Detail page loads |
| `munki://` wrapping encoded `http://127.0.0.1/poc.html` | Re-interpreted as nonexistent filename in htmlDir; no remote navigation |
| `munki://` wrapping encoded `file:///etc/passwd` | Same — no remote navigation |
| `munki://` wrapping encoded `smb://attacker/share` | Same — no remote navigation |
| `munki://` wrapping encoded `javascript:alert(1)` | Resolves to nonexistent filename; webView shows standard not-found page (same UX as `munki://detail-fake`) |

Attack URLs that smuggle remote URLs through `munki://` are neutralized by construction — `handleMunkiURL` → `load_page` no longer has the capability to navigate remote.

## Relationship to #1359

This PR is an alternative approach to #1359, not an addition. Both close the same vulnerability:

- **#1359** — input-layer fix: rejects any URL with embedded scheme at the entry of `handleMunkiURL`.
- **This PR** — architecture fix: removes the remote-navigation capability from the `handleMunkiURL` → `load_page` path entirely.

Please let me know which approach you prefer and I'll happily close the other.

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
+++ b/code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
@@ -251,7 +251,12 @@ class MainWindowController: NSWindowController {
             } else {
                 msc_debug_log("Could not parse sidebar item page URL \(page)")
             }
+        } else if let components = URLComponents(string: page),
+                  ["https", "http"].contains(components.scheme)
+        {
+            load_url(page)
         } else {
+            // Treat as a relative page name in htmlDir.
             load_page(page)
         }
     }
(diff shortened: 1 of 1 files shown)


gregneagle:
This to me seems a cleaner overall fix for the issue: making it clear what the intention is (display a MSC "page" or navigate to an arbitrary URL) and making functions that do only that. Less likely there will be other edge-case vulnerabilities around munki:// URLs.

PR #1369, opened 2026-07-02
Fix missing comma in ASU symlink list

homebysix:
I realize this is old/unsupported code, but wanted to offer a quick fix regardless.

### DIFF
--- a/code/client/munkilib/adobeutils/core.py
+++ b/code/client/munkilib/adobeutils/core.py
@@ -468,7 +468,7 @@ def run_adobe_cpp_pkg_script(dmgpath, payloads=None, operation='install'):
     tmpdir = tempfile.mkdtemp(prefix='munki-', dir='/tmp')
 
     # make our symlinks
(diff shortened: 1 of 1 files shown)


gregneagle:
Thank you. I'm of two minds here. First: a fix is always good. Second: any change at all means a new pkg version for `com.googlecode.munki.pythonlibs`. which then implies there's maintenance/support. But the intention is to remove that package (at least as a default install) in the not-too-distant future.

The fact that no-one found this bug is a pretty good indication this code was rarely if ever used back when it made some sort of sense to install Adobe products with this method.

Curious how/why you found this -- were you actually using the code, or did you point Claude/whatever at the codebase and ask it to look for typos/whatever?

homebysix:
The latter — I'm experimenting with using multiple agents and models to cross-check each other and reduce false-positives when bug hunting. This was one of the true positives flagged (despite my instruction that the agents ignore the python codebase).

gregneagle:
I'm going to close this since the Python libraries exist only as a transitional aid and will be removed in the future. Fixing a bug will only make it more likely that people continue using it.

PR #1370, opened 2026-07-12
Enable client certificate auth against servers that don't advertise acceptable CA names

nstrauss:
## Problem

Munki's client certificate support currently depends on the server advertising a list of acceptable client-certificate CAs in its TLS CertificateRequest. When a client certificate is requested, `getClientCertCredential()` in `clientcerts.swift` parses `protectionSpace.distinguishedNames` and matches keychain identities against the advertised issuer DNs. If the server sends an empty CA list, the function returns `nil` before examining any keychain identity, and the connection is cancelled:

```
Authentication challenge for Host: example-repo.example.net Realm:  AuthMethod: NSURLAuthenticationMethodClientCertificate
Client certificate required
The server didn't send the list of acceptable certificate-issuing authorities
Can't authenticate
Download error -999: cancelled
ERROR: Could not retrieve managed install primary manifest: There was a connection error: -999: cancelled
```

Per RFC 8446 (TLS 1.3), this `certificate_authorities` advertisement is optional. Some widely used servers never send it, notably Google Cloud external HTTPS load balancers (GFE), which omit the CA name list on both TLS 1.2 and TLS 1.3 with no server-side configuration option to change this. Other cloud load balancers behave similarly.

The result is that Munki cannot perform mutual TLS against a Munki repo fronted by a Google Cloud load balancer (or any server that omits CA names), even when a perfectly valid client identity is present in the keychain. There is currently no workaround short of modifying the server (impossible for GFE) or patching Munki. I believe I have a much better chance of fixing this via Munki than convincing Google to listen.

## Solution

Add a new preference, `ClientCertificateAcceptableCAs`, an array of strings. Each string is a distinguished name in the same format Munki already logs for server-advertised issuers (RFC 4514-style, as rendered by swift-certificates' `DistinguishedName.description`):

```xml
<key>ClientCertificateAcceptableCAs</key>
<array>
    <string>CN=Munki Client CA,O=SomeOrg</string>
</array>
```

During the client certificate authentication challenge, configured DNs are treated as a union with any server-advertised CA names when matching keychain identities. This lets an admin declare which issuing CA(s) to expect when the server doesn't say, the same information the server would have provided in its CertificateRequest.

Because the configured strings and the candidate identities' issuer DNs are both rendered by the same library (`DistinguishedName.description`), matching is exact and deterministic, with no custom DN string parsing. The correct value for the preference can be copied directly from Munki's own debug log output (`Accepted certificate-issuing authority: ...`) or derived from `openssl x509 -noout -subject` on the issuing CA certificate.

## No impact to existing deployments

This is a safe, purely opt-in change:

- Preference unset (the default): behavior is unchanged. If the server advertises CA names, identity matching proceeds exactly as before. If the server advertises none, Munki declines to authenticate exactly as before (`-999: cancelled`). Only the log message differs, now mentioning the new preference to aid discovery.
- Preference set: configured DNs are only *added* to the acceptable-issuer set. Matching against server-advertised names is unchanged. An identity is never presented unless its issuer chain matches an acceptable DN, so no identity that wouldn't have matched before can be silently offered.
- No changes to keychain handling, server trust evaluation, or any other network behavior.
- The preference is added to `CONFIG_KEY_NAMES` so it appears in `managedsoftwareupdate --show-config`, following the same display behavior as the existing array-valued preference `AdditionalHttpHeaders`.

### Live test against a Google Cloud external HTTPS load balancer

Tested against a private Munki repo served through a Google Cloud external Application Load Balancer configured for mTLS, with a client identity in the System keychain.

Before (stock behavior / preference unset): connection cancelled before any keychain identity is considered:

```
Client certificate required
The server didn't send the list of acceptable certificate-issuing authorities
No acceptable certificate-issuing authorities are available, either sent by the server or configured via the ClientCertificateAcceptableCAs preference
Can't authenticate
Download error -999: cancelled
ERROR: Could not retrieve managed install primary manifest: There was a connection error: -999: cancelled
```

After (preference set): the matching identity is found and presented, and the TLS handshake proceeds with the client certificate:

```
Client certificate required
The server didn't send the list of acceptable certificate-issuing authorities
Configured acceptable certificate-issuing authority: CN=Example Client CA,O=ExampleOrg
Found matching identity
Will attempt to authenticate
```

Backward-compatibility check: after deleting the preference, behavior returns to the exact stock outcome (`-999: cancelled` before any keychain access), confirming existing deployments are unaffected by upgrading.

## Documentation

If this is accepted, I'm happy to add `ClientCertificateAcceptableCAs` to the [Preferences wiki page](https://github.com/munki/munki/wiki/Preferences) alongside the other client certificate preferences, and an explanation of the proper DN format.

### DIFF
--- a/code/cli/munki/munki.xcodeproj/project.pbxproj
+++ b/code/cli/munki/munki.xcodeproj/project.pbxproj
@@ -428,6 +428,7 @@
 		C0AD96302F3E58D80029A293 /* appleupdatesTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C0AD96192F3E58D80029A293 /* appleupdatesTests.swift */; };
 		C0AD96312F3E58D80029A293 /* cliutilsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C0AD961A2F3E58D80029A293 /* cliutilsTests.swift */; };
 		C0AD96322F3E58D80029A293 /* compareTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C0AD961B2F3E58D80029A293 /* compareTests.swift */; };
+		C0AD96FE2F3E58D80029A293 /* clientcertsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C0AD96FD2F3E58D80029A293 /* clientcertsTests.swift */; };
 		C0AD96332F3E58D80029A293 /* downloadTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C0AD961C2F3E58D80029A293 /* downloadTests.swift */; };
 		C0AD96342F3E58D80029A293 /* factsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C0AD961D2F3E58D80029A293 /* factsTests.swift */; };
 		C0AD96352F3E58D80029A293 /* fetchTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C0AD961E2F3E58D80029A293 /* fetchTests.swift */; };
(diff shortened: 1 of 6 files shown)


gregneagle:
This seems reasonable. I'd really like other folks using client certs/mTLS to weigh-in: does this seem reasonable?

Are there any other scenarios around client certs/mTLS that might require additional preferences?

weswhet:
We use mTLS with CA advertising. Seems reasonable to me.

matt-cahill:
Seems reasonable to me. 
Would it matter if the configured DN was also received from a server? E.g. in the case that you configure clients to handle both a local server inside your network that sends acceptable CAs and a cloud one for outside that doesn't.

nstrauss:
> Would it matter if the configured DN was also received from a server? E.g. in the case that you configure clients to handle both a local server inside your network that sends acceptable CAs and a cloud one for outside that doesn't.

@matt-cahill This PR is written to use a union of values. CA DNs configured with `ClientCertificateAcceptableCAs` are added to those values advertised by the server. Both will be included so your use case would work.

SoxIn4:
Looks like a good and reasonable approach to fill a gap.

gregneagle:
Have created a new Munki7_3dev branch and will merge this there.

gregneagle:
Test build here: https://github.com/macadmins/munki-builds/releases/download/v7.3.0.5790/munkitools-7.3.0.5790.pkg

matt-cahill:
Test build works against our mTLS server

```bash
Managed Software Update Tool
Version 7.3.0.5790
Copyright 2010-2026 The Munki Project
https://github.com/munki/munki

Starting...
    No client cert info provided, so no client keychain will be created.
Checking for available updates...
    Getting manifest development...
    Fetching URL: https://redacted.tld/manifests/development
    FollowHTTPRedirects is: none
    Authentication challenge for Host: redacted.tld Realm:  AuthMethod: NSURLAuthenticationMethodServerTrust
    Allowing OS to handle authentication request
    Authentication challenge for Host: redacted.tld Realm:  AuthMethod: NSURLAuthenticationMethodClientCertificate
    Client certificate required
    Accepted certificate-issuing authority: CN=Redacted-Root-CA,OU=redacted,O=redacted
    Found matching identity
    Will attempt to authenticate
    Status: 304
    Item is unchanged on the server.
    /Library/Managed Installs/manifests/development already exists and is up-to-date.
    Retrieved manifest development
```

PR #1371, opened 2026-07-20
Defer downloads on low data connections

grahamgilbert:
## Overview

Munki currently downloads all managed items regardless of the network connection. This adds an opt-in ability to **defer downloads on low-data connections** — cellular, personal hotspot, or a Wi-Fi network in Low Data Mode — with an optional user override in Managed Software Center and per-item control in pkginfo.

Detection uses Apple's Network framework (`NWPathMonitor`): a connection is "low data" when `path.isExpensive` (cellular/hotspot) or `path.isConstrained` (Low Data Mode) is true — the same signals macOS uses for its own Low Data Mode restrictions, so no SSID/interface configuration is required.

## Admin preferences

Set in `/Library/Preferences/ManagedInstalls.plist` or via a configuration profile.

| Key | Type | Default | Description |
| :- | :- | :- | :- |
| `MaxSizeOverLowDataConnection` | Integer (KB) | `-1` | Size threshold for low-data deferral. `-1` disables this feature. `0` defers auto/default downloads on low-data connections. Positive values defer items whose `installer_item_size` exceeds the threshold. Same KB unit as `installer_item_size`. |
| `AllowLowDataOverride` | Boolean | `true` | Whether users may override a deferral in Managed Software Center. `false` makes it a hard block (no "Download anyway" button). |

## pkginfo key

`download_on_low_data` (string): `auto` (default when absent — follow the size threshold), `always` (download regardless), `never` (never on low data). Written by `makepkginfo --download_on_low_data <value>`; `makecatalogs` warns on invalid values.

## Decision logic (per item, before download)

```
On a low data connection?
  NO  -> download normally (no change)
  YES ->
    forced item past its force_install_after_date (or a required dependency of one)? -> download
    installer already fully cached (no data needed)? -> download
    user "Download anyway" override (and AllowLowDataOverride)? -> download (override consumed)
    MaxSizeOverLowDataConnection < 0? -> download (feature disabled)
    download_on_low_data == always? -> download
    download_on_low_data == never?  -> defer
    otherwise (auto): installer_item_size > MaxSizeOverLowDataConnection? -> defer, else download
```

With `MaxSizeOverLowDataConnection` set to `0`, auto/default items with a real download are deferred on low-data connections, but explicit exceptions still work: `download_on_low_data: always`, user overrides when allowed, fully cached installers, and items past `force_install_after_date` still download.

A deferred item is recorded with `low_data_deferred = true` and is surfaced like any other item that can't be downloaded right now (e.g. insufficient disk space): it shows in Managed Software Center but is not counted as a pending update. Precaching is skipped entirely on a low-data connection.

## Managed Software Center

Deferred items appear under "Problem updates" as **Download paused — low data connection**. When overrides are allowed, a **Download anyway** button (list + detail page) records the choice and triggers a check so the item downloads on that run; the override is consumed once the item downloads.

## Testing

Validated end-to-end against a staging repo on a Munki 7 client with Wi-Fi Low Data Mode enabled:
- `download_on_low_data: never` → item deferred; not downloaded; not counted as an update.
- MSC shows the paused row + **Download anyway**; confirming writes the override and the item downloads on the next check.
- `MaxSizeOverLowDataConnection`: `-1` disables deferral; `0` defers auto/default downloads; positive thresholds defer items over the threshold and allow under-threshold items to download.
- `download_on_low_data: always`, user overrides, already-cached installers, and a reached `force_install_after_date` still download.

Unit tests cover the `shouldDeferDownloadForLowData` decision matrix (`munkiCLItesting`). The branch also went through a `/simplify` code-quality pass and a `codex` functional review, whose findings are fixed here.

## Notes / limitations

- The size threshold is **per item**, not cumulative — it prevents any single large download but doesn't cap total data over a period. Set it to `0` to defer auto/default downloads on low-data connections, or use `download_on_low_data: never` on specific items that should never download on low data.
- Detection relies entirely on what macOS reports; Munki does not try to detect "slow" links or signal strength.

## Documentation (wiki edits, post-merge)

Munki's wiki isn't part of this repo, so no wiki changes are committed here. Below is the exact content to add to each wiki page once this merges.

### Add to [Supported Pkginfo Keys](https://github.com/munki/munki/wiki/Supported-Pkginfo-Keys)

**`download_on_low_data`** *(string)* — Controls whether this item is downloaded when the Mac is on a low-data connection (cellular, personal hotspot, or a Wi-Fi network in Low Data Mode). One of:
- `auto` — default when the key is absent; follow the `MaxSizeOverLowDataConnection` size threshold.
- `always` — always download, regardless of the connection or size threshold.
- `never` — never download on a low-data connection.

Set with `makepkginfo --download_on_low_data <value>`. `makecatalogs` warns on unrecognized values.

### Add to [Preferences](https://github.com/munki/munki/wiki/Preferences)

**`MaxSizeOverLowDataConnection`** *(integer, KB, default `-1`)* — Size threshold for the `auto` low-data behavior. `-1` disables low-data deferral. `0` defers auto/default downloads on low-data connections. Positive values defer items whose `installer_item_size` exceeds the threshold. Same KB unit as `installer_item_size`.

**`AllowLowDataOverride`** *(boolean, default `true`)* — Whether users may override a low-data deferral in Managed Software Center via the "Download anyway" button. `false` makes the deferral a hard block (no button shown).

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center/Controllers/MSCAlertController.swift
+++ b/code/apps/Managed Software Center/Managed Software Center/Controllers/MSCAlertController.swift
@@ -323,27 +323,27 @@ class MSCAlertController: NSObject {
     
     func confirmDownloadAnyway(_ itemName: String) {
         // Confirm the user wants to download a low-data-deferred item over
-        // their low data connection, then record the override so
-        // managedsoftwareupdate downloads it on its next run.
+        // their low data connection, record the override, then kick off a
+        // check so it downloads now rather than silently waiting for the next
+        // scheduled run.
         guard let mainWindow = window else {
             return
         }
         let alert = NSAlert()
         alert.messageText = NSLocalizedString(
-            "Download this item anyway?",
+            "Download anyway?",
             comment: "Download Anyway alert title")
-        alert.informativeText = NSLocalizedString(
-            "This item will be downloaded over a low data connection, which "
-                + "may use cellular data. It will download the next time "
-                + "Managed Software Center checks for updates.",
-            comment: "Download Anyway alert detail")
         alert.addButton(withTitle: NSLocalizedString(
             "Download anyway", comment: "Download anyway button title"))
         alert.addButton(withTitle: NSLocalizedString(
             "Cancel", comment: "Cancel button title/short action text"))
         alert.beginSheetModal(for: mainWindow, completionHandler: { (modalResponse) -> Void in
             if modalResponse == .alertFirstButtonReturn {
-                if !addLowDataOverride(itemName) {
+                if addLowDataOverride(itemName) {
+                    if let mainWindowController = (NSApp.delegate as? AppDelegate)?.mainWindowController {
+                        mainWindowController.checkForUpdates()
+                    }
+                } else {
                     msc_debug_log("Could not write low-data override for \(itemName)")
                 }
             }
--- a/code/apps/Managed Software Center/Managed Software Center/Controllers/MSCAlertController.swift
+++ b/code/apps/Managed Software Center/Managed Software Center/Controllers/MSCAlertController.swift
@@ -321,6 +321,35 @@ class MSCAlertController: NSObject {
         })
     }
     
+    func confirmDownloadAnyway(_ itemName: String) {
+        // Confirm the user wants to download a low-data-deferred item over
+        // their low data connection, then record the override so
+        // managedsoftwareupdate downloads it on its next run.
+        guard let mainWindow = window else {
+            return
+        }
+        let alert = NSAlert()
+        alert.messageText = NSLocalizedString(
+            "Download this item anyway?",
+            comment: "Download Anyway alert title")
+        alert.informativeText = NSLocalizedString(
+            "This item will be downloaded over a low data connection, which "
+                + "may use cellular data. It will download the next time "
+                + "Managed Software Center checks for updates.",
+            comment: "Download Anyway alert detail")
+        alert.addButton(withTitle: NSLocalizedString(
+            "Download anyway", comment: "Download anyway button title"))
+        alert.addButton(withTitle: NSLocalizedString(
+            "Cancel", comment: "Cancel button title/short action text"))
+        alert.beginSheetModal(for: mainWindow, completionHandler: { (modalResponse) -> Void in
+            if modalResponse == .alertFirstButtonReturn {
+                if !addLowDataOverride(itemName) {
+                    msc_debug_log("Could not write low-data override for \(itemName)")
+                }
+            }
+        })
+    }
+
     func confirmUpdatesAndInstall() {
         // Make sure it's OK to proceed with installing if logout or restart is
         // required
--- a/code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController+WKScriptMessageHandler.swift
+++ b/code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController+WKScriptMessageHandler.swift
@@ -51,6 +51,11 @@ extension MainWindowController: WKScriptMessageHandler {
                 updateOptionalInstallButtonFinishAction(item_name)
             }
         }
+        if message.name == "downloadAnywayButtonClicked" {
+            if let item_name = message.body as? String {
+                downloadAnywayButtonClicked(item_name)
+            }
+        }
         if message.name == "openExternalLink" {
             if let link = message.body as? String {
                 openExternalLink(link)
@@ -146,6 +151,13 @@ extension MainWindowController: WKScriptMessageHandler {
     
     func updateOptionalInstallButtonBeginAction(_ item_name: String) {
         webView.evaluateJavaScript("fadeOutAndRemove('\(item_name)')")
+    }
+
+    func downloadAnywayButtonClicked(_ item_name: String) {
+        // called from JavaScript when the user clicks "Download anyway" on a
+        // low-data-deferred update; confirm, then record the override
+        msc_debug_log("User clicked Download anyway for \(item_name)")
+        alert_controller.confirmDownloadAnyway(item_name)
     }
     
     func myItemsActionButtonClicked(_ item_name: String) {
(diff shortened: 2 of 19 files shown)


gregneagle:
Wondering if, in practice, admins might rather specify specific products/items/packages they would like downloaded (and therefore kept updated) while deferring others. IOW, I think it's possible that some admins might want to say "even on a low data connection, keep Chrome, Firefox, Slack, Zoom, and VS Code updated. Defer everything else." That's not possible with this implementation. Might this be worth discussing with the Munki admin community?

gregneagle:
As I think about it, an admin could get the behavior I decide by setting `MaxSizeOverLowDataConnection` to, say, 1, and then setting `download_on_low_data` to "always" for the items they want always kept updated. They'd have to remember to add that key to the pkginfo for each newly imported version (or set up AutoPkg to do so).
Perhaps that's good enough. Given that literally no-one has asked about or proposed functionality like this (that I remember) in the 15+ years of Munki's existence, I may be overthinking something that may not be used by very many orgs.

jc0b:
Just wanna chime in and say this would solve one of the more common feedback items we get from employees, so we’d definitely implement this. There are a select few applications that are large that we don’t need to force if folks are on low-data connections on their commute or wherever. 

> an admin could get the behavior I decide by setting MaxSizeOverLowDataConnection to, say, 1, and then setting download_on_low_data to "always" for the items they want always kept updated.

This is also what I would consider as an approach to “I need to make sure these items don’t get deferred”.

macmule:
FWIW.. In Jamf Auto Update there is a boolean to key `network_aware` it's true by default:

> When true, Jamf Auto Update does not run when Low Data Mode is enabled for the Wi-Fi network the Mac is connected to, or when connected to a hotspot.

This is triggered via our munki preflight script, and has been we received. Not as granular as mentioned here though.

gregneagle:
Feeling like we should make "When on low data, defer everything except items that have passed a `force_install_after_date`" an easy/obvious setting.

(and unless the admin sets very short timeframes for `force_install_after_date`, and users are on low data for days or weeks, the actual download of the item most likely has occurred earlier when not on low data -- it's just been pending)

grahamgilbert:
`allow_overdue_force_install_after_date`? As a Boolean value

gregneagle:
Alternately, `MaxSizeOverLowDataConnection`: change it so -1 is the default and means this feature is disabled. 0 means no downloads at all over a low data connection (except for those past a `force_install_after_date`. (Or leave it as it is currently defined and implemented, and tell admins to set it to 1, which defers all updates greater than 1KB in size, which is probably "good enough")

grahamgilbert:
I’m easy either way, you tell me what you prefer

grahamgilbert:
@gregneagle have made the change according to your suggestion

grahamgilbert:
Good catch. I’d been treating `0` as a hard block, but I agree that’s too restrictive.

I’ve pushed a follow-up so `-1` disables the feature, and `0` now means “defer all auto/default downloads” via the normal threshold comparison. Explicit exceptions still work, so `download_on_low_data = "always"` can download even when `MaxSizeOverLowDataConnection` is `0`.

I also updated the tests/help text and restored the MSC “Download anyway” path so it follows `AllowLowDataOverride` rather than requiring a positive threshold.

grahamgilbert:
@gregneagle do you have any other questions or suggestions or are we good to go on this one?

MagerValp:
I've had two tickets today, one from a researcher on a polar expedition and one somewhere in Africa, and both could be solved neatly with this. LGTM!

The only improvement I could suggest is also exposing the low-data state as a condition, but that could also be solved pretty neatly with an external fact.

gregneagle:
My primary question/hesitation with the PR as it stands today is that the low data logic applies only to individual items and not the entire updatecheck session.

It might prevent one 5 gig item from downloading, but in the same session, allow 5, 6, or more 1 gig downloads. Would people expect/accept that?

IOW, is it OK that multiple small updates that in aggregate are quite large allowed, but a single large update is deferred?

If the people who want this feature are OK with that, then I have no other hesitations.

grahamgilbert:
I think this provides the most flexibility for users to get high priority updates out whilst keeping the known chunky lower priority updates off low bandwidth connections. If we wanted to add an additional option for the total size I can do that in a separate PR. 

I like the option of exposing this in a condition, I’ll write a follow up with that.

MagerValp:
> IOW, is it OK that multiple small updates that in aggregate are quite large allowed, but a single large update is deferred?

Yes, for our use case I think it makes more sense to just block individual items. I'm still researching exactly how slow the networks are for the researchers and I haven't combed through our repo to find the perfect threshold, but it'll probably by something in the neighborhood of 1-10 MB. Small enough that our scripts, agents, and system tools slip through, but blocking anything resembling an application update.

Blocking based on the total update size would cause a single Word or Teams update (and there's always at least one rolling out) to clog everything up and we'd have to go in and add download_on_low_data: always to all our important scripts and system tools to get them through.

MagerValp:
In case anyone's curious, 1 MB seems like the perfect threshold for us. The largest system tool we deploy is 740 kB, and the smallest "app" we deploy is 1.5 MB.

MagerValp:
Can I suggest using a boolean named `low_data_connection` as the condition? I'm implementing it with an external fact for now as I need to have a quick fix ready, but it'd be nice if we could switch over to the built in condition if it becomes available.

gregneagle:
This auto-close was a mistake triggered by the deletion of the Munki7_3dev branch. I'll fix this shortly...

gregneagle:
I'm merging this into Munki7_4dev for a future Munki 7.4 release.
The new user-facing strings must be localized into the other 11 supported languages before a 7.4 release can happen.

MagerValp:
These three right?

/* Download anyway button title */
"Download anyway" = "Ladda ner ändå";

/* Low data connection deferred download note */
"Download paused — low data connection" = "Nerladdning pausad — låg dataförbrukning";

/* Download Anyway alert title */
"Download anyway?" = "Ladda ner ändå?";

gregneagle:
@grahamgilbert Could you start the documentation for this here, please? https://github.com/munki/munki/wiki/Low-Data-Connection-Support

gregneagle:
Ping

grahamgilbert:
In progress!

grahamgilbert:
All done!

PR #1374, opened 2026-07-27
Fix sidebar navigation on macOS 27

weswhet:
## What was the bug?

On macOS 27, clicking an item in the Managed Software Center sidebar did not always open the correct page. The click event could arrive before macOS finished updating the selected sidebar row. This meant the app could read an old or invalid row when deciding which page to open.

There was also a separate refresh issue. When Managed Software Center decided that it should show the Updates page, the refresh code could load that page and then immediately load it a second time.

## How did we fix it?

Sidebar navigation now waits until the next turn of the main event loop before reading the selected row. This gives macOS time to finish updating the selection. The app then opens the page linked to that selected item. The sidebar action is also connected when the window is first set up, so it does not depend on the window becoming active.

The window-mode check now reports whether it already opened the Updates page. The refresh code uses that result to avoid loading the same page again.

## Testing

Tested on macOS 27, have not testing on previous macOS versions yet.


gregneagle:
Going to merge this into a test branch and test on macOS 26, 27, and perhaps 15

gregneagle:
Brief testing on macOS 10.15, 26, and the current 27 beta all seems fine.

PR #1375, opened 2026-08-03
Add --rebuild-catalogs flag to munkiimport

discentem:
This PR adds a new --rebuild-catalogs flag to munkiimport for automatically rebuilding catalogs after importing items.

## Changes
- Add new --rebuild-catalogs flag to munkiimportOptions
- When flag is passed, automatically rebuild catalogs after import (works in both interactive and non-interactive modes)
- When flag is not passed and running interactive, prompt user as before (current behavior)
- Extract catalog rebuild logic into performCatalogRebuild() helper function to eliminate code duplication

## Usage Examples
```bash
# Force rebuild catalogs automatically
munkiimport --rebuild-catalogs package.pkg

# Interactive mode with flag (no prompt)
munkiimport --rebuild-catalogs -i package.pkg

# Non-interactive mode with flag
munkiimport --rebuild-catalogs --nointeractive package.pkg

# Interactive mode without flag (prompts as before)
munkiimport package.pkg
```

### DIFF
--- a/code/cli/munki/munkiimport/munkiimport.swift
+++ b/code/cli/munki/munkiimport/munkiimport.swift
@@ -426,20 +426,28 @@ struct MunkiImport: AsyncParsableCommand {
             print("Saved pkginfo to \(pkginfoPath).")
         }
         // Maybe rebuild the catalogs?
-        if !munkiImportOptions.nointeractive {
+        if munkiImportOptions.rebuildCatalogs {
+            print("Rebuilding catalogs...")
+            try await performCatalogRebuild(repo: repo)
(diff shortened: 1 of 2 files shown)


gregneagle:
What happens if `--rebuild-catalogs` is specified but an error occurs during the import process?

discentem:
> What happens if `--rebuild-catalogs` is specified but an error occurs during the import process?

In [the latest commit](https://github.com/munki/munki/pull/1375/changes/081692afc4cfae8a9164ce30a26e96c3ea5f7dad) I updated the behavior to only rebuild catalogs or prompt to rebuild if the pkginfo upload succeeds. Does that solve your concern?

PR #1377, opened 2026-08-05
always look for repo plugins in /usr/local/munki/repoplugins, instead of relative to munkiimport binary

discentem:
When I compiled `munkiimport` myself in order to test another pull request (https://github.com/munki/munki/pull/1375) and I was running it from a random directory, I was surprised to find out that `repoConnect()` and `findRepoInPlugins()` actually looks for plugins relative to the compiled binary. 

This behavior was confusing for me because the documentation implies the plugins are always discovered from `/usr/local/munki/repoplugins`. 

This PR changes repoPluginsDir to always be `/usr/local/munki/repoplugins` regardless of where the binary is running from.

### DIFF
--- a/code/cli/munki/shared/munkirepo/RepoFactory.swift
+++ b/code/cli/munki/shared/munkirepo/RepoFactory.swift
@@ -54,7 +54,7 @@ private func loadRepoPlugin(at path: String) throws -> RepoPluginBuilder {
 /// Try to load a Repo plugin from our RepoPlugins directory
 func findRepoInPlugins(_ name: String, url: String) throws -> Repo? {
     let pluginName = name + ".plugin"
-    let repoPluginsDir = (Bundle.main.bundlePath as NSString).appendingPathComponent("repoplugins")
+    let repoPluginsDir = "/usr/local/munki/RepoPlugins"
     let repoPluginNames = (try? FileManager.default.contentsOfDirectory(atPath: repoPluginsDir)) ?? []
     if repoPluginNames.contains(pluginName) {
         let pluginPath = (repoPluginsDir as NSString).appendingPathComponent(pluginName)


gregneagle:
I think leaving the current behavior is more useful, as it allows for alternate installation locations for the Munki tools, and definitely makes testing easier.
Are there concerns about the current behavior, other than clarifying the documentation?

discentem:
> I think leaving the current behavior is more useful, as it allows for alternate installation locations for the Munki tools, and definitely makes testing easier.
> 
> Are there concerns about the current behavior, other than clarifying the documentation?

I found it made testing harder personally but I understand your view. Clarifying the docs is certainly appreciated. 

Perhaps a compromise would be an alternante implementation that would allow --plugin to be a full path? Would that be acceptable?

Of course referring to a plugin by name would continue to work and use the existing logic.

gregneagle:
I can't see a common need for allowing --plugin to be a full path, and past experience has taught me that having a value be a name and possibly a path leads to unwanted/unexpected behavior (because we're not always perfect at disambiguation)

What actual problem (other than initial confusion) did you have? Once you realized the plugin actually had to be in path relative to the Munki admin tool binary, you could have copied the plugin or created a symlink.

Seems to me this is just a case of needing to be clearer/more accurate in the documentation, and requires no changes to code.

gregneagle:
I've updated the Wiki pages for Repo Plugins and Middleware to clarify the search location for the plugins.

discentem:
> I've updated the Wiki pages for Repo Plugins and Middleware to clarify the search location for the plugins.

The problem was just initial confusion. Could we at least update the -h info for plugins to also say they plugins are discovered by name relative to the binary? Also describing in -h how name maps to the actual plugin would be helpful. 

Something like "if you pass S3Repo to --plugin, it will look for a plugin at $munkiimport_dir/repoplugins/S3Repo.plugin"

gregneagle:
I don't think that change to the help output is necessary. That's the sort of thing that goes in a man page. But the Munki tools don't currently have man pages, so the wiki will have to suffice.

gregneagle:
In any case, closing this PR since I don't want to hard-code the `/usr/local/munki` path.

PR #1379, opened 2026-08-14
Support launch arguments when reopening blocking apps

nstrauss:
## Problem

I'm a huge fan of all the new assisted quit and relaunch of blocking applications features. However, the reopen path currently stores only each application's bundle path and calls `NSWorkspace.openApplication()` with an empty `NSWorkspace.OpenConfiguration`.

That works for applications whose normal launch behavior restores the expected state, but there is no way for an admin to supply application specific launch arguments. Google Chrome is one example. Chrome supports `--restore-last-session`, which asks it to restore the previous browsing session. In testing a Chrome update through Managed Software Center, I noticed Chrome reopened without restoring the expected tabs. In an effort to avoid pitchforks and angry Slack messages, I thought MSC should be able to assist with this as well.

There is no existing assisted quit setting that can do this. `blocking_applications_quit_script` runs before the update and replaces the normal quit operation. It does not provide a post-install relaunch hook and cannot participate in the user's "Reopen applications after update" choice.

## Solution

Add a new optional pkginfo key, `blocking_applications_launch_args`, whose value is an array of strings where each array element is passed as one application argument. . MSC associates the array with the pkginfo item's `blocking_applications` and passes it through `NSWorkspace.OpenConfiguration.arguments` when reopening an application after a successful assisted update.

For Google Chrome, a pkginfo can contain:

```xml
<array>
</array>
<array>
</array>
```

The key is scoped to one pkginfo item and applies to every application listed in that item's `blocking_applications`. If multiple pending items identify the same application with different argument arrays, the first value in resolved update-list order is retained and the conflict is written to the MSC debug log. Combining arrays could produce an invalid application command line, and allowing later items to replace the first value would make the selected behavior less obvious.

The key is copied into the processed managed-install, optional-install, and removal metadata so it is available wherever the existing Assisted Quit keys are available.

## No impact to existing deployments

This is an opt-in change:

- Key unset: Managed Software Center reopens the application with no additional arguments, matching current behavior.
- Empty array: Managed Software Center reopens the application with no additional arguments.
- Incorrect value type or an array containing non-string values: Managed Software Center ignores the value and reopens the application with no additional arguments.
- "Reopen applications after update" cleared: the application remains closed after the update, regardless of configured arguments.
- Application being removed: the existing removal safeguard remains unchanged and the application is not reopened.
- Update does not complete successfully: the existing behavior remains unchanged and queued applications are not reopened.

The arguments only affect applications that Managed Software Center closed through Assisted Quit and would already have reopened. This does not change application detection, quit behavior, force-quit behavior, installation behavior, or the user's reopen choice.

### End-to-end testing with Google Chrome

Tested against commit `6fe97b3c` using branch-built `managedsoftwareupdate` and Managed Software Center. I used a local Munki repo with no-payload items so each case exercised catalog generation, update analysis, the Assisted Quit sheet, successful installation, and a real Chrome relaunch without changing the installed Chrome version.

The following cases passed:

- `blocking_applications_launch_args` set to `["--restore-last-session"]`: Chrome quit and reopened under a new PID with `--restore-last-session` in the main process command line. Chrome tabs retained their state and re-opened as expected.
- Key omitted: Chrome quit and reopened under a new PID with no added arguments.
- Empty array: Chrome quit and reopened under a new PID with no added arguments.
- Malformed scalar string: MSC ignored the value and reopened Chrome without added arguments.
- Mixed-type array containing a string and integer: MSC ignored the value and reopened Chrome without added arguments.
- Two pending items with conflicting arrays: MSC used the first value, logged the conflict, and did not pass the second marker argument.
- Valid array with "Reopen applications after update" cleared: the update completed and Chrome remained closed until opened manually.

For the valid case, the generated `InstallInfo.plist` contained:

```text
"blocking_applications_launch_args" => [
  0 => "--restore-last-session"
]
```

Chrome was running as PID `90555` before the update and reopened as PID `91863` with:

```text
/Applications/Google Chrome.app/Contents/MacOS/Google Chrome --restore-last-session
```

The conflicting-item case recorded:

```text
MSC: debug  Found blocking_applications_launch_args for Google Chrome.app
MSC: debug  Ignoring conflicting blocking_applications_launch_args for Google Chrome.app
MSC: debug  Reopening app in background: /Applications/Google Chrome.app
```

Chrome restored the profile active at relaunch. Other Chrome profiles restored their saved sessions when those profiles were opened.

I also ran the full `munkiCLItesting` suite, including the new metadata-copy tests, and built Managed Software Center with `xcodebuild`. A separate AppKit test application received two arguments unchanged through `NSWorkspace.OpenConfiguration.arguments`, including an argument containing spaces.

## Documentation

If this is accepted, I'm happy to update the [Assisted Quit of Blocking Applications](https://github.com/munki/munki/wiki/Assisted-Quit-of-Blocking-Applications) wiki page with the new key, its array type, and the Chrome example above.


gregneagle:
I think this is interesting, and may be of use to some Munki admins (and can be safely ignored by others; IOW it does not impose new work on Munki admins; it's an opt-in feature).

I've also been working on some enhancements around the handling of blocking apps, and they are in the [blocking_apps_enhancement](https://github.com/munki/munki/tree/blocking_apps_enhancement) branch. I think it would be good to ensure that your PR works as expected with these changes. Specifically, processes that are not actually _apps_ can now be quit (and force quit) from MSC.app. I don't think there is an attempt to relaunch them, however, so I think your proposed changes have little risk of bad behavior with non-app processes. On the other hand, this might confuse admins...

gregneagle:
As an experiment I'm going to edit this PR to see if it would merge cleanly against the blocking_apps_enhancement branch...

gregneagle:
...and as I feared, there are some conflicts.

nstrauss:
@gregneagle I'm open to either...

1. Resolving the conflicts.
2. Convincing you to merge this into another feature branch (Munki7_3dev) and deal with integrating into the blocking apps enhancements later. 

As the use case is isolated to Chromium based browsers at this point it isn't a time sensitive thing, but it is a nice quality of life feature.

jordancalhoun (commented):
Overall, this looks good. I found one issue that should be fixed before merging:

Munki allows blocking apps to be written in several valid ways, such as Google Chrome.app, Google Chrome, or a full path. This saves launch arguments using the configured name but later looks them up using Google Chrome.app.

Launch arguments may be ignored unless the blocking app was written with the exact name. Different names for the same app could also bypass the “first value wins” conflict handling.

nstrauss:
Thanks @jordancalhoun I read through https://github.com/munki/munki/wiki/Blocking-Applications again to understand every supported form of `blocking_applications` and made changes to accommodate each in https://github.com/munki/munki/pull/1379/commits/2099fff80e8e7b571edf8d178cd6249096a9ec3b

I changed the association to use the resolved application bundle path returned by getRunningBlockingApps(). For Chrome, every supported form now resolves to the same key: `/Applications/Google Chrome.app`

The reopen queue uses that same resolved path to retrieve the arguments. This also means different forms that identify the same application participate in the same first-value-wins conflict handling.

| Pkginfo form | Behavior |
| --- | --- |
| Name without `.app` | Supported. `getRunningBlockingApps()` first checks the executable name, then falls back to matching `/<name>.app/Contents/MacOS/` in `munki.swift`. The resulting process path is normalized to the containing application bundle before the arguments are recorded. |
| Name with `.app` | Supported. The existing resolver matches executables inside the named application bundle and normalizes the process path to the bundle path. |
| Full executable path inside an app bundle | Supported. The resolver exact-matches the executable path, then removes the `/Contents/...` portion to produce the containing `.app` path. |
| `blocking_applications` omitted | Supported. `blockingApplicationsForItem()` derives application names from application entries in `installs`. Those names pass through the same resolver and produce the same bundle path. |
| Equivalent-form conflicts | Supported. Values such as `Google Chrome`, `Google Chrome.app`, and the full Chrome executable path resolve to one pathname. `recordLaunchArguments()` therefore preserves the first update-list configuration and logs a later conflicting value. |
| Non-application process blocker | Launch arguments are not associated with bare processes. The mapping only accepts resolved paths ending in `.app`, since `NSWorkspace.openApplication()` reopens application bundles. Existing blocking behavior for non-application processes is otherwise unchanged by this PR. |

flammable:
I can't speak to the code, but I like this idea and would use this. Thanks for identifying the issue and submitting a PR!

gregneagle:
I'm likely to merge _my_ changes to Munki7dev before yours, so I'd prefer you resolved the conflicts with the `blocking _apps_enhancement` branch.

gregneagle:
The auto-closing was a mistake -- I deleted the blocking_apps_enhancement branch. I'll reopen this and point it at the new Munki7_4dev branch.

nstrauss:
@gregneagle No problem. I resolved all merge conflicts with Munki7_4dev. My guess is this is still in progress and more conflicts might come up so I'll leave the PR alone for now, but can check back when we're closer to a 7.4 beta.

gregneagle:
OK, silly me, I merged this into Munki7_4dev, and then looked at the diff, and now have a serious question/reservation.
You have a new pkginfo key: "blocking_applications_launch_args", which is an array of strings.

`blocking_applications` is itself an array. It could list ["Google Chrome", "Firefox", "Safari"]. Would your new "blocking_applications_launch_args" now be applied against all three apps? IOW, each app would be relaunched with `--restore-last-session` ?

Feels like this needs to be a dictionary so that the correct launch arguments get mapped to the correct application:

```xml
<dict>
    <key>Google Chrome</key>
    <array>
    <array>
    <key>Firefox</key>
    <array>
        <string>--something-else</string>
    <array>
</dict>
```

nstrauss:
@gregneagle Yeah I agree. This was an oversight. I'll open a follow up PR to address.

gregneagle:
@nstrauss Could you start documenting this feature here? https://github.com/munki/munki/wiki/Launch-arguments-for-relaunched-applications ?

PR #1381, opened 2026-08-25
Scope blocking app launch arguments by application

nstrauss:
This is a follow up to #1379 based on [Greg's review comment](https://github.com/munki/munki/pull/1379#issuecomment-5413466181).

The original `blocking_applications_launch_args` schema uses one argument array for an entire pkginfo item. If an item lists several blocking applications, Managed Software Center passes the same application-specific arguments to all of them.

For example, a Chrome-specific `--restore-last-session` array would also be passed to Firefox and Safari if those applications appeared in the same `blocking_applications` list. Launch arguments need to be associated with individual applications instead.

### Changes

Changes `blocking_applications_launch_args` to a dictionary mapping blocking application identifiers to arrays of strings. Managed Software Center passes each array only to its corresponding application through `NSWorkspace.OpenConfiguration.arguments`.

For Google Chrome, a pkginfo can contain:

```xml
<array>
    <string>Firefox.app</string>
</array>
<dict>
    <key>Google Chrome.app</key>
    <array>
    </array>
</dict>
```

Chrome receives `--restore-last-session`; Firefox reopens without additional arguments. Each array element is passed as one application argument without shell parsing.

Dictionary keys accept the same forms as `blocking_applications`: names with or without `.app` and full executable paths. Mapping keys and blocker entries can use different equivalent forms because both resolve to the running application bundle path. Applications inferred from `installs` are supported as well.

Mappings are scoped to effective blockers for that pkginfo item. Unrelated mappings and non-bundle processes are ignored. Malformed values are ignored individually, so one bad mapping does not discard valid sibling mappings.

The key is copied into the processed managed-install, optional-install, and removal metadata so it is available wherever the existing Assisted Quit keys are available.

If equivalent keys or multiple pending items configure the same canonical application differently, the first deterministic value is retained and the conflict is written to the MSC debug log. Arrays are never combined.

### Compatibility

The array schema was introduced by #1379 on the unreleased Munki 7.4 development branch. This follow-up intentionally replaces it rather than supporting both ambiguous and per-application forms. Existing pkginfo files omit the new key and retain current behavior.

Existing assisted quit safeguards remain unchanged. Arguments affect only applications MSC closed and would already reopen, and do not override the user's reopen choice, removal handling, or failed update behavior.

### Testing

I tested this with branch-built `managedsoftwareupdate`, `makecatalogs`, and Managed Software Center using an isolated local Munki repo. This tested catalog generation, update analysis, the Assisted Quit sheet, successful installation, and real application relaunches without changing the installed application versions.

One test item blocked two running applications, Google Chrome and Safari. The processed `InstallInfo.plist` preserved the dictionary unchanged.

The following cases passed:

- Distinct valid mappings: Chrome reopened under a new PID with `--restore-last-session`, and Safari reopened under a new PID with `--args-for-safari-test`.
- Valid Chrome mapping with a malformed Safari scalar: Chrome still reopened with `--restore-last-session`, and Safari reopened with no additional arguments.
- Equivalent forms: the mapping key `Google Chrome` matched `Google Chrome.app` in `blocking_applications` and resolved to `/Applications/Google Chrome.app`.
- Unrelated mapping: a mapping for `Unrelated App.app` had no effect because it was not an effective blocker for the item.

Chrome's tab URL list was identical before and after the live tests.

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center/Controllers/MSCBlockingAppsController.swift
+++ b/code/apps/Managed Software Center/Managed Software Center/Controllers/MSCBlockingAppsController.swift
@@ -190,23 +190,40 @@ class MSCBlockingAppsController: NSObject {
                 }
             }
 
-            if let launchArguments = update_item["blocking_applications_launch_args"] as? [String] {
-                let appPaths = Array(Set(runningBlockingApps.map(\.pathname).filter {
+            if let launchArgumentMappings = update_item["blocking_applications_launch_args"] as? [String: Any] {
+                let itemAppPaths = Set(runningBlockingApps.map(\.pathname).filter {
                     $0.hasSuffix(".app")
-                }))
-                let configuredApps = appPaths.filter { appLaunchArguments[$0] == nil }
-                let conflicts = recordLaunchArguments(
-                    launchArguments,
-                    for: appPaths,
-                    in: &appLaunchArguments
-                )
-                for appPath in configuredApps {
-                    msc_debug_log("Found blocking_applications_launch_args for \(appPath)")
-                }
-                for appPath in conflicts {
-                    msc_debug_log(
-                        "Ignoring conflicting blocking_applications_launch_args for \(appPath)"
+                })
+                for appIdentifier in launchArgumentMappings.keys.sorted() {
+                    guard let launchArguments = launchArgumentMappings[appIdentifier] as? [String] else {
+                        msc_debug_log(
+                            "Ignoring invalid blocking_applications_launch_args for \(appIdentifier)"
+                        )
+                        continue
+                    }
+                    let appPaths = Array(Set(
+                        getRunningBlockingApps([appIdentifier]).map(\.pathname)
+                    ).intersection(itemAppPaths)).sorted()
+                    if appPaths.isEmpty {
+                        msc_debug_log(
+                            "Ignoring blocking_applications_launch_args for non-blocking application \(appIdentifier)"
+                        )
+                        continue
+                    }
+                    let configuredApps = appPaths.filter { appLaunchArguments[$0] == nil }
+                    let conflicts = recordLaunchArguments(
+                        launchArguments,
+                        for: appPaths,
+                        in: &appLaunchArguments
                     )
+                    for appPath in configuredApps {
+                        msc_debug_log("Found blocking_applications_launch_args for \(appPath)")
+                    }
+                    for appPath in conflicts {
+                        msc_debug_log(
+                            "Ignoring conflicting blocking_applications_launch_args for \(appPath)"
+                        )
+                    }
                 }
             }
         }
--- a/code/cli/munki/munkiCLItesting/analyzeTests.swift
+++ b/code/cli/munki/munkiCLItesting/analyzeTests.swift
@@ -23,15 +23,18 @@ struct assistedQuitMetadataTests {
     @Test func copiesLaunchArguments() {
         let pkginfo: PlistDict = [
             "blocking_applications_launch_args": [
-                "--restore-last-session",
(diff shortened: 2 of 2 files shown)


nstrauss:
Tagging @gregneagle for review.

gregneagle:
One more thought. There's some repetition here: 
```xml
<array>
    <string>Firefox.app</string>
</array>
<dict>
    <key>Google Chrome.app</key>
    <array>
    </array>
</dict>
```

What if the key was named `blocking_applications_with_launch_args`, and looked like this?
```xml
<key>blocking_applications_with_launch_args</key>
<dict>
    <key>Firefox</key>
    <array>
    </array>    
    <key>Google Chrome.app</key>
    <array>
    </array>
</dict>
```
That is to say, the _keys_ are the names of the blocking applications. So you would not need to specify an application both places. Maybe this could be legal as well:
```xml
<array>
    <string>Firefox.app</string>
</array>
<key>blocking_applications_with_launch_args</key>
<dict>
    <key>Google Chrome.app</key>
    <array>
    </array>
</dict>
```
and also:
```xml
<array>
    <string>Firefox.app</string>
</array>
<key>blocking_applications_with_launch_args</key>
<dict>
    <key>Google Chrome.app</key>
    <array>
    </array>
</dict>
```
In all these cases, the keys for `blocking_applications_with_launch_args` would be merged with the strings from the `blocking_applications` array -- perhaps as a set to avoid duplicates.

I wonder how difficult it would be to just say that `blocking_applications` can be an array of strings _or_ a dict. So either:

```xml
<array>
    <string>Firefox.app</string>
</array>
```
or
```xml
<dict>
    <key>Firefox</key>
    <array>
    </array>    
    <key>Google Chrome.app</key>
    <array>
    </array>
</dict>
```
This might wreck havoc with linting tools, and GUI editors like MunkiAdmin and MunkiWebAdmin, which expect each key to have a single supported type. Still it might simplfy things and cut down on repetition, especially in the common case of a single blocking application with launch arguments.

nstrauss:
Thanks @gregneagle  I went back through the current Munki7_4dev implementation and both PRs. Since neither the original array schema from #1379 nor the dictionary schema from #1381 has shipped, I agree we should choose the better final schema rather than preserve the current key for compatibility. I like `blocking_applications_with_launch_args`, with its keys merged with the strings in `blocking_applications`. It removes the repetition in the common case without changing the established type of `blocking_applications`. I would define the behavior as follows:
- If either `blocking_applications` or `blocking_applications_with_launch_args` is present, use the union of the array strings and dictionary keys.
- If neither key is present, retain the existing inference from application entries in installs.
- Resolve and deduplicate equivalent identifiers by the running application path, as the current #1381 implementation does.
- If an application appears in both keys, treat it as one blocker and use the arguments from the dictionary.
- If a dictionary value is malformed, keep its key as a blocker but ignore the arguments and reopen the application normally.

I would avoid changing the data structure of `blocking_applications`. The current Swift code casts it to `[String]`, the older Python code iterates it directly, and tools such as MunkiAdmin expect an array. Supporting both types under the same key would require every consumer and editor to understand both forms from then on. I think keeping it as is is better for backwards compatibility. 

My preference is to replace the current unreleased `blocking_applications_launch_args` key with `blocking_applications_with_launch_args`, and keep `blocking_applications` as an array. If that all works for you I'll open a new PR with those changes. Let me know if any further edits or ideas for the implementation.

gregneagle:
This all sounds good. I agree about redefining `blocking_applications` to be either an array of strings or a dictionary. That would be more elegant in one way, but probably would break other things and therefore perhaps not worth the disruption.

nstrauss:
@gregneagle My initial try at this is now available at https://github.com/munki/munki/pull/1393

PR #1383, opened 2026-08-26
MSC OnDemand/optional_installs/uninstalls button customization(/localization) support

arubdesu:
OnDemand has always displayed the same button text/labels as optional_installs do, which are often not applicable to the 'action' being performed. In looking at the code paths required to enable this, both optional_installs and the new auto-optional_uninstalls capability COULD leverage this as well, and the lines of code/files changed were small, so I went for it.
I also chose (with the guidance/assistance of a friendly LLM clanker) to split functions that handle the status items and convert the handling from `Text` to `.setInnerHTML` so that it's more consistent with other escaping of pkginfo/admin-provided customized or substituted text. Please let me know if we'd like to rename keys, undo the 'consistency' change, or otherwise roll back affecting optional_installs as part of this at all.
Tested by only updating the app and basing off the munki7_4dev's currently posted package., screenshots of working QA procedures posted to Slack (I can post them here as well).
Ready with wiki edits when the time comes. Thanks for your consideration!

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
+++ b/code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
@@ -1565,13 +1565,13 @@ class MainWindowController: NSWindowController {
                 }
                 if btn_classes.contains("install-updates") {
                     //(btn as! DOMHTMLElement).innerText = item["myitem_action_text"] as? String ?? ""
-                    self.setInnerText(item["myitem_action_text"] as? String ?? "", elementID: btn_id)
+                    self.setInnerHTML(item["myitem_action_text"] as? String ?? "", elementID: btn_id)
                 } else if btn_classes.contains("long_action_text") {
                     //(btn as! DOMHTMLElement).innerText = item["long_action_text"] as? String ?? ""
-                    self.setInnerText(item["long_action_text"] as? String ?? "", elementID: btn_id)
+                    self.setInnerHTML(item["long_action_text"] as? String ?? "", elementID: btn_id)
                 } else {
                     //(btn as! DOMHTMLElement).innerText = item["short_action_text"] as? String ?? ""
-                    self.setInnerText(item["short_action_text"] as? String ?? "", elementID: btn_id)
+                    self.setInnerHTML(item["short_action_text"] as? String ?? "", elementID: btn_id)
                 }
                 // use innerHTML instead of innerText because sometimes the status
                 // text contains html, like '<span class="warning">Some warning</span>'
--- a/code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
+++ b/code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
@@ -1565,13 +1565,13 @@ class MainWindowController: NSWindowController {
                 }
                 if btn_classes.contains("install-updates") {
                     //(btn as! DOMHTMLElement).innerText = item["myitem_action_text"] as? String ?? ""
-                    self.setInnerHTML(item["myitem_action_text"] as? String ?? "", elementID: btn_id)
+                    self.setInnerText(item["myitem_action_text"] as? String ?? "", elementID: btn_id)
                 } else if btn_classes.contains("long_action_text") {
                     //(btn as! DOMHTMLElement).innerText = item["long_action_text"] as? String ?? ""
-                    self.setInnerHTML(item["long_action_text"] as? String ?? "", elementID: btn_id)
+                    self.setInnerText(item["long_action_text"] as? String ?? "", elementID: btn_id)
                 } else {
                     //(btn as! DOMHTMLElement).innerText = item["short_action_text"] as? String ?? ""
-                    self.setInnerHTML(item["short_action_text"] as? String ?? "", elementID: btn_id)
+                    self.setInnerText(item["short_action_text"] as? String ?? "", elementID: btn_id)
                 }
                 // use innerHTML instead of innerText because sometimes the status
                 // text contains html, like '<span class="warning">Some warning</span>'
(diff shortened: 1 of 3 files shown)


code/apps/Managed Software Center/Managed Software Center/Controllers/MainWindowController.swift
```diff
@@ -1565,13 +1565,13 @@ class MainWindowController: NSWindowController {
                 }
                 if btn_classes.contains("install-updates") {
                     //(btn as! DOMHTMLElement).innerText = item["myitem_action_text"] as? String ?? ""
-                    self.setInnerText(item["myitem_action_text"] as? String ?? "", elementID: btn_id)
+                    self.setInnerHTML(item["myitem_action_text"] as? String ?? "", elementID: btn_id)
```
gregneagle, line 1568: Was there a specific reason for switching the method from`setInnerText` to `setInnerHTML`?

gregneagle: "it's more consistent" doesn't really capture the trade-offs here. It's not 100% clear what the differences are between `setInnerText` and `setInnerHTML` are, so it would be good to understand the risks/benefits of this change.

gregneagle: Is the intention to support more than simple text strings for the button labels, and if so, do you have an example of what one might do?

arubdesu: > Is the intention to support more than simple text strings for the button labels...

Hadn't thought of that, I approved/went with this approach just for more-than-zero 'input validation', so the text gets escaped, but also because re:
> the trade-offs ... risks/benefits

for this window from the first 'draw', specifically the [buttons, is in an HTML template](https://github.com/munki/munki/blob/10e08290ad2bf9e0638aa498215d25349c7f6f2c/code/apps/Managed%20Software%20Center/Managed%20Software%20Center/Resources/templates/list_item_template.html#L24-L33). we're both plugging the default/hardcoded/translated strings and then later live-updating 'rendered' HTML. I'm not considering it so much of a tradeoff as considering the previous behavior an omission, (especially since instances of `status_text` already uses it while going through the same code path).
The primary risk kindof was leaving it. Of course, future localizations that are added which go through this code path POTENTIALLY could include chars like < that stop further rendering - there IS the one localized string (French, ` "Cancel install" = "Annuler l'installation"`, of the 195 key/locale pairs present) with an apostrophe that (harmlessly) turns into `&#39;`, but both (initialize and update) paths handle/decode it: the template is parsed as HTML on first draw, and `setInnerHTML` on live update - and it renders with its apostrophe either way, but we can't prevent inadvertent or intentional failures in future customizations or localizations.

As a swift/compiled app novice, though, I can't really gauge for myself how dangerous or wrong or worth changing this is, to be perfectly honest, so if you'd like me to revert and have the clanker help me deliver the feature (or some permutation of it, since I'm still not absolutely certain the optional_(un or)installs support is endorsed) while still leaving it as it was with setInnerText I can do that, I just can't come up with a decent explanation why NOT moving forward with setInnerHTML it is 'better'.

Thanks for probing on this point though, it made me realize if we're being preemptively 'safer', localized strings riding on this code path for buttons should also get escaped just the same in `MunkiItems.swift`, commit added to this PR. I also re-QA'd to confirm intended behavior remains, markup isn't allowed to pass through, and the French localized 'cancel' string remains decoded back to its apostrophe (in the top right of this screenshot)
Capture d’écran 2026-08-27 à 13 38 30
markup-safety confirmation screenshot (the lower-right "`Fixing <b>now<\b>`" button label):
Capture d’écran 2026-08-27 à 13 27 11

gregneagle: That last example ("Fixing &lt;b>now&lt;\b>") shows that we are getting no benefit at all from using `setInnerHTML`. Button labels should be _text_ and not HTML. _Descriptions_, on the other hand, can contain links and "styled" text and `<br>` elements, and so setting the description of the item uses `setInnerHTML`.

I guess what I'm asking here: can you show be an example where using `setInnerText` causes a problem that using `setInnetHTML` fixes? If not, my inclination is to _not_ change that.

gregneagle: Additionally, I think it's good to attempt to protect admins from mistakes, but OTOH I'd expect admins to _test_ their custom button labels and action messages as well.

arubdesu: Understood. Pardon the lag, I'll have this converted back and QA'd ASAP. Marking as draft in the interim
Do you think the optional_(un)installs support is ok to leave in?

gregneagle:
I don't see any harm in letting admins customize the button labels for any/all items -- no need to limit it to OnDemand.

arubdesu:
Ok, I've reverted the escapeHTML/setInnerHTML and confirmed the optional_(un)install behavior is QA'd working. I also checked that this does allow the one relatively-contrived failure scenario of **_IF_** there's, as an example, an interpreted invalid tag in an opening pointy bracket, all following characters are swallowed. I can update the relevant parts of the wiki accordingly with the corner-case potential 'gotcha'.

gregneagle:
Had a chance to pay attention to this again today. Wondering if the key name "action_labels" is too generic/not describing itself as well as it might. Maybe "custom_action_labels" or "custom_labels" would be better?

arubdesu:
`custom_action_labels` is a touch verbose but with 'action' in the middle it does feel like it's making the distinction of what class of button labels can be customized, that sound good? (I know I said I'd add the wiki docs in anticipation but I didn't get the chance just yet, pardon.)

gregneagle:
So let's go with `custom_action_labels`.

arubdesu:
Ok, updated, created https://github.com/munki/munki/wiki/Customizing-Text-Button-Labels-for-On-Demand-or-Optional-Installs and referenced its contents in https://github.com/munki/munki/wiki/On-Demand-Items

PR #1384, opened 2026-08-27
Handle cached manifests blocking nested paths

nstrauss:
## Summary

Munki 7.3 can leave a client unable to complete update checks when a repository changes a manifest path between a file and a directory prefix. For example, a client that previously cached the manifest `policy/conditions` cannot retrieve a new manifest named `policy/conditions/test_manifest`. The inverse transition also fails when a cached `policy/conditions` directory blocks a new manifest at that path.

The Swift client treats the cached `policy/conditions` file as if it were the directory needed for the new manifest. The subsequent download receives an HTTP 200 response, but Munki cannot create the local `.download` file below that path and reports an unrelated connection error:

```text
ERROR: Could not retrieve manifest policy/conditions/test_manifest from the server: Download error 200: no error
```

This change prepares the complete local manifest destination for both transition directions. It removes a regular cached manifest when that file must become a directory, and removes a stale cached directory when that path must become a manifest file.

Recovery is confined to the manifest cache. Symlinks below the configured cache root are replaced without following or modifying their targets, while unsupported path components are rejected. Removal failures report descriptive errors from Foundation `FileManager`.

## Regression from the Swift client

The legacy Python `get_manifest()` implementation uses `os.path.isdir()` before creating a nested manifest directory. An existing regular file fails that check and produces a local folder-creation error before an HTTP request is attempted.

The Swift `getManifest()` implementation calls `createMissingDirs()`. That helper currently returns `true` for any existing path:

```swift
if filemanager.fileExists(atPath: path) {
    return true
}
```

It does not verify that the existing path is a directory. This behavior was introduced with the Swift implementation and is present in Munki 7.3.0.

After that incorrect success, `Gurl` attempts to create `<destination>.download`. `FileManager.createFile()` returns `false` because an intermediate path component is a regular file, and `FileHandle(forWritingAtPath:)` returns `nil`. Neither result is checked. The URL session continues, the response body has no writable destination, and the eventual error is presented as `Download error 200: no error`.

The normal manifest cleanup cannot recover the client because cleanup runs after a successful update check. The manifest retrieval error aborts the check first, leaving the blocking cached file in place for the next run.

The inverse transition reaches the final download move, but that move fails because the generic fetch layer replaces regular files rather than directories. Manifest-specific preparation removes the stale directory before retrieval without changing replacement behavior for packages, catalogs, icons, or other resources.

## Reproduction

I reproduced this with Munki `7.3.0.5790`:

1. Serve a manifest named `policy/conditions` and run an update check so it is cached locally.
2. Replace that repository manifest with `policy/conditions/test_manifest` and include the new name from another manifest.
3. Run another update check.

The client retains this regular file:

```text
/Library/Managed Installs/manifests/policy/conditions
```

The next check needs the same path to be a directory:

```text
/Library/Managed Installs/manifests/policy/conditions/test_manifest
```

The server and load balancer recorded a complete HTTP 200 response for the child manifest while the client produced:

```text
Getting manifest policy/conditions/test_manifest...
Status: 200
ERROR: Could not retrieve manifest policy/conditions/test_manifest from the server: Download error 200: no error
```

Starting from an empty `/Library/Managed Installs` directory did not reproduce the error. Munki created the nested directory and retrieved the same manifest successfully. Restoring the former `policy/conditions` manifest as a regular file reproduced the error consistently.

## Changes

- Make `createMissingDirs()` return `false` when the requested path exists but is not a directory.
- Add manifest-specific destination preparation for both file-to-directory and directory-to-file cache transitions.
- Preserve an existing regular manifest for normal conditional retrieval and replacement.
- Confine recovery to the manifest cache and remove stale cache items with Foundation `FileManager` APIs.
- Replace symlinks below the configured manifest cache root without following them, while preserving administrator-configured path indirection at or above the cache root.
- Leave unsupported path types in place and fail closed rather than removing them.

The manifest-specific removal is intentionally narrow. The generic directory helper does not delete existing files, since it is also used by installer and client-resource paths.

## Tests

Added regression coverage for:

- An existing directory accepted by `createMissingDirs()`.
- An existing regular file rejected by `createMissingDirs()`.
- Replacement of a cached manifest file with the required directory, including when the file is above the immediate destination directory.
- Removal of a cached directory and its descendant manifests when the directory path becomes a manifest file.
- Preservation of an existing regular manifest at the final destination.
- Replacement of symlinks occupying a required parent path or final manifest destination without modifying their targets.
- Replacement of a dangling symlink at the final manifest destination.

The full `munkiCLItesting` Xcode suite passes with its normal parallel test configuration.

### DIFF
--- a/code/cli/munki/shared/updatecheck/manifests.swift
+++ b/code/cli/munki/shared/updatecheck/manifests.swift
@@ -17,6 +17,7 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 
+import Darwin.C
 import Foundation
 
 private let PRIMARY_MANIFEST_TAG = "_primary_manifest_"
@@ -75,6 +76,47 @@ class Manifests {
     }
 }
 
+/// Makes a local directory available for a manifest download. A cached
+/// manifest may occupy this path when a repository changes a manifest name
+/// into a directory prefix.
+func prepareManifestDirectory(_ path: String, cacheRoot: String) -> Bool {
+    let standardizedPath = (path as NSString).standardizingPath
+    let standardizedRoot = (cacheRoot as NSString).standardizingPath
+    if standardizedPath != standardizedRoot,
+       !standardizedPath.hasPrefix(standardizedRoot + "/")
+    {
+        display.error("Manifest directory is outside the manifest cache: \(path)")
+        return false
+    }
+
+    let relativePath = String(standardizedPath.dropFirst(standardizedRoot.count + 1))
+    var candidatePath = standardizedRoot
+    for component in relativePath.split(separator: "/") {
+        candidatePath = (candidatePath as NSString).appendingPathComponent(String(component))
+        if pathIsSymlink(candidatePath) {
+            display.error("Symlink blocks manifest directory creation at \(candidatePath)")
+            return false
+        }
+        if !pathExists(candidatePath) {
+            break
+        }
+        if pathIsDirectory(candidatePath) {
+            continue
+        }
+        if !pathIsRegularFile(candidatePath) {
+            display.error("Unsupported file type blocks manifest directory creation at \(candidatePath)")
+            return false
+        }
+        if unlink(candidatePath) != 0 {
+            let description = String(cString: strerror(errno))
+            display.error("Could not remove manifest blocking directory creation at \(candidatePath): \(description)")
+            return false
+        }
+        break
+    }
+    return createMissingDirs(path)
+}
+
 /// Gets a manifest from the server.
 ///
 /// Returns:
@@ -87,10 +129,11 @@ func getManifest(_ name: String, suppressErrors: Bool = false) throws -> String 
         return manifestLocalPath
     }
 
+    let manifestCacheRoot = managedInstallsDir(subpath: "manifests")
     let manifestLocalPath = managedInstallsDir(subpath: "manifests/\(name)")
     // make sure the directory exists to store it
     let manifestLocalPathDir = (manifestLocalPath as NSString).deletingLastPathComponent
-    if !createMissingDirs(manifestLocalPathDir) {
+    if !prepareManifestDirectory(manifestLocalPathDir, cacheRoot: manifestCacheRoot) {
         throw ManifestError.notRetrieved(
             "Could not create a local directory to store manifest")
     }
--- a/code/cli/munki/shared/updatecheck/manifests.swift
+++ b/code/cli/munki/shared/updatecheck/manifests.swift
@@ -17,7 +17,6 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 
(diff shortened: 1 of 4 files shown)


code/cli/munki/shared/updatecheck/manifests.swift
```diff
@@ -17,6 +17,7 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 
+import Darwin.C
```
gregneagle: What's the purpose of this include?

code/cli/munki/shared/updatecheck/manifests.swift
```diff
@@ -75,6 +76,47 @@ class Manifests {
...
+            break
+        }
+        if pathIsDirectory(candidatePath) {
+            continue
+        }
+        if !pathIsRegularFile(candidatePath) {
+            display.error("Unsupported file type blocks manifest directory creation at \(candidatePath)")
+            return false
+        }
+        if unlink(candidatePath) != 0 {
```
gregneagle: Oh I see now what Darwin.C is used for now. Any reason not to use Foundation FileManager methods instead?

gregneagle:
Curious: a manifest can block the creation of a directory, and you've addressed that. How about the opposite? Previously `policy/conditions/test_manifest` was cached; now there's an included manifest named `policy/conditions`...

code/cli/munki/shared/updatecheck/manifests.swift
gregneagle: `unlink` could be replaced with `try FileManager.default.removeItem(atPath: candidatePath)`, and you could catch the error to get a descriptive error message.

nstrauss:
@gregneagle I switched the removal to FileManager.removeItem(atPath:), included the localized error when removal fails, and removed the Darwin.C import.

I also tested the inverse transition you mentioned. A cached policy/conditions directory prevents policy/conditions.download from being moved into place when policy/conditions becomes a manifest file. The preparation now handles both directions:

- A cached manifest file is removed when its path needs to become a directory.
- A cached directory and its descendants are removed when its path needs to become a manifest file.

I reviewed the pre-Swift behavior for symlinks as well. The Python client followed parent-directory symlinks through normal path resolution, but generally replaced a symlink at the final destination through os.rename(). The updated code does not follow symlinks within the manifest cache. It removes the link itself and restores the required cache path, leaving the link target unchanged. A configured symlinked ManagedInstallDir continues to work, but symlinked manifests below it are properly handled.

gregneagle:
I'm going to merge this. As for handling of symlinks or anything other than directories and regular files in the manifests cache directory -- I'm of the opinion that that directory is "owned" by `managedsoftwareupdate`, and so should never contain anything but files and directories created by `managedsoftwareupdate`. Therefore, one could argue for removing any filesystem item that blocks the creation of a cached manifest.

PR #1385, opened 2026-08-27
Fix low data connection deferral UI

grahamgilbert:
Addresses feedback from the Munki 7.4 dev preview thread on the low data download deferral UI.

## Changes

**1. Duplicate "Download paused — low data connection" note (updates list)**

The paused note rendered twice for a deferred item: once in the update-list row (`${note}` in `update_item_template.html`) and again via the item `description()`, which prefixes the note as a warning. Removed the list-row `${note}`; the note still shows once via the description, and the detail view (which only uses `description()`) is unchanged.

**2. Redundant "Download anyway?" confirmation alert**

The confirm alert title echoed the button it was triggered from. The guard behaviour is unchanged (it protects against accidental clicks), but the title no longer repeats the button:

- Title: `Low data connection detected and item size exceeds limit` — no exact byte sizes (avoids users asking for limit tweaks, and avoids value-ordering issues across languages).
- New informative line: `Are you sure you want to download it now?`
- Buttons unchanged: **Download anyway** / **Cancel**.

Swedish strings updated to match — **please verify the translation**.

## Dependencies

Confirmed (no code change): munki will not download or install an item while any of its `requires` items is deferred — `processInstall` marks the parent's `dependenciesMet` false and records a "can't verify requirements" note instead. Forced items past their deadline still propagate a low-data exemption to their requirements. The one rough edge is that a user "Download anyway" override does not cascade to requirements, so a chain of large items must be overridden item-by-item across check cycles.

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center/MunkiItems.swift
+++ b/code/apps/Managed Software Center/Managed Software Center/MunkiItems.swift
@@ -1017,17 +1017,15 @@ class UpdateItem: GenericItem {
             // Show a localized "paused" message instead of the plain-English
             // note recorded by managedsoftwareupdate on the CLI side.
             my["note"] = lowDataPausedNote()
-            // Offer a neutral "More Info" button (unless the admin disabled
-            // user overrides via AllowLowDataOverride). It opens a dialog that
-            // explains the deferral and offers the actual "Download anyway"
-            // action, so the compact row button stays short and doesn't imply
-            // an action on its own.
-            if munkiPref("AllowLowDataOverride") as? Bool ?? true {
-                my["hide_low_data_button"] = ""
-                my["low_data_action_text"] = NSLocalizedString(
-                    "More Info",
-                    comment: "Low data More Info button title")
-            }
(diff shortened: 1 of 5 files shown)


code/apps/Managed Software Center/Managed Software Center/MunkiItems.swift
```diff
@@ -1017,13 +1017,16 @@ class UpdateItem: GenericItem {
             // Show a localized "paused" message instead of the plain-English
             // note recorded by managedsoftwareupdate on the CLI side.
             my["note"] = lowDataPausedNote()
-            // Offer a "Download anyway" button unless the admin disabled
-            // user overrides via AllowLowDataOverride.
+            // Offer a neutral "More Info" button (unless the admin disabled
```
gregneagle: Even if the admin has disabled user overrides, I'd think we'd still want the More Info button and the dialog/alert.

If the admin has disabled user overrides, the alert will not have a "Download anyway" button.

grahamgilbert: Updated!

PR #1388, opened 2026-08-29
Log the error domain and code when a download fails

rodchristiansen:
A download failure that isn't a `URLError` logs only its `localizedDescription`, and `FetchError.connection` carries a placeholder code of `0`, so both the domain and the real errno are lost. When a machine runs out of ephemeral ports and `connect()` returns POSIX `EEXIST`, the run reports:

```
ERROR: Could not retrieve catalog Production from server: Connection error 0: The operation couldn't be completed. File exists
```

That reads like a filesystem problem, and it sends you looking at the local cache and the download path rather than at the socket layer. With the domain and code in the line it identifies itself as `NSPOSIXErrorDomain 17` and the search starts in the right place.

This reports the domain and code for non-`URLError` failures, carries the real code into `FetchError.connection` in place of `0`, and logs any `NSUnderlyingError`. `URLError` handling and the SSL branch are untouched.

Builds and runs clean against a live repo; the improved line itself was written from the error object rather than captured on a live failure, since the machines that were failing have since drifted back under the port limit.

### DIFF
--- a/code/cli/munki/shared/network/fetch.swift
+++ b/code/cli/munki/shared/network/fetch.swift
@@ -292,8 +292,18 @@ func getURL(
             errorDescription = urlError.localizedDescription
             display.detail("Download error \(errorCode): \(errorDescription)")
         } else {
+            // not a URLError -- most usefully a POSIX errno from the socket
+            // layer, so report the domain and code and not just the
+            // description, which on its own can be very misleading
+            errorCode = error.code
             errorDescription = error.localizedDescription
-            display.detail("Download error: \(errorDescription)")
+            display.detail("Download error \(error.domain) \(errorCode): \(errorDescription)")
+        }
+        if let underlyingError = error.userInfo[NSUnderlyingErrorKey] as? NSError {
+            display.detail(
+                "Underlying error \(underlyingError.domain) \(underlyingError.code): " +
+                    "\(underlyingError.localizedDescription)"
+            )
         }
         if session.SSLerror != 0 {
             errorCode = session.SSLerror


gregneagle:
Is the claim here that URLSessions are persisting even after `managedsoftwareupdate` exits?

rodchristiansen:
No the claim was only ever within a single process: each `Gurl` makes a session, nothing invalidates it, so the sessions and their connection pools accumulate for the life of that run. But digging further, with some help from my coding agent, that turns out not to be what breaks the run at all.

Found a couple more machines that reproduce it, and the cause is that they have run out of ephemeral ports.

macOS allocates outbound source ports from 49152–65535 — 16,384 of them. On the three failing machines, `netstat -an -p tcp` counts 13,970 / 16,448 / 16,442 sockets sitting in TIME_WAIT, so the range is effectively full. A healthy machine on the same subnet has one. Almost all of those sockets are to a single third-party host on :443, unrelated to Munki, from one source address, and they never drain — the counts are identical minutes apart, and a packet capture shows no traffic to that host at all.

With no free port left for that destination, `connect()` fails with `EEXIST`, and CFNetwork passes the errno straight through as "The operation couldn't be completed. File exists" — which reads like a filesystem problem and is why this looked like a download or cache bug.

Three checks confirmed it, and they don't involve Munki:

- A minimal `NSURLSession` harness fails on the second fetch with `NSPOSIXErrorDomain` 17, whether or not each session is invalidated, and equally with an ephemeral configuration or the URL cache disabled — so session lifetime is not the variable.
- `curl` to the same URL fails too, so it is not URLSession at all.
- `sysctl net.inet.ip.portrange.first=32768` widens the range and every failure disappears on the spot; restoring 49152 brings them straight back.

So Munki is the victim here, not the cause.
Still useful to improve the logs?
Today the failure logs as:

```
ERROR: Could not retrieve catalog Production from server: Connection error 0: The operation couldn't be completed. File exists
```

For anything that isn't a `URLError` Munki logs only `localizedDescription`, and the code falls back to a placeholder `0`, so both the domain and the real errno are lost.`NSPOSIXErrorDomain 17` in that line points at the socket layer immediately.

I'll strip this PR back to exactly that — log the domain, the code and any underlying error on a connection failure, and carry the real code into `FetchError.connection` instead of `0` — and drop the session-invalidation and retry changes.

gregneagle:
I like the simpler, more focused change here. It still might make sense to (in another PR) explicitly invalidate NSURLSessions if that's best/better practice than what Munki is doing currently.

PR #1390, opened 2026-09-02
Treat a missing application usage table as no data instead of an error

rodchristiansen:
**The claim:** on any Mac where nobody has ever clicked Install or Remove in Managed Software Center, every run of `managedsoftwareupdate` logs one ERROR per installed optional item that carries `unused_software_removal_info`. The condition is an expected state, not a fault, and the message does not say what actually happened.

**What is wrong**

`shouldBeRemovedIfUnused()` runs for every currently installed optional install that has `unused_software_removal_info`. Its second step calls `daysSinceLastInstallEvent()`, which prepares

```
SELECT last_time FROM install_requests WHERE event=? AND item_name=?
```

`install_requests` is created lazily by `ApplicationUsageRecorder.log_install_request()`, the first time somebody makes a self-service request in MSC. Until then the table does not exist, `sqlite3_prepare` fails with `no such table: install_requests`, and `ApplicationUsageQuery` reports it at error level.

The text is also unhelpful. `SQL3Error` does not adopt `LocalizedError`, so `localizedDescription` throws away the sqlite3 message and substitutes Foundation's placeholder. The `error 1` is Swift's default code for a non-`LocalizedError`, not `SQLITE_ERROR`. Interpolating the error itself gives the real message.

**How to reproduce**

1. A Mac with Munki 7.3.0 (or current `Munki7_4dev`) where MSC has never been used to install or remove anything, so `/Library/Managed Installs/application_usage.sqlite` has an `application_usage` table but no `install_requests` table. Confirm with `sqlite3 application_usage.sqlite .tables`.
2. Have at least one optional install that is currently installed and has `unused_software_removal_info` with `removal_days` smaller than the days of usage data on the machine.
3. Run `sudo managedsoftwareupdate --checkonly`.

**Before**

```
ERROR: Error querying /Library/Managed Installs/application_usage.sqlite: The operation couldn't be completed. (managedsoftwareupdate.SQL3Error error 1.)
```

once per such item, every run.

**After**

Nothing at the default log level. With `-vv`:

```
No data yet in /Library/Managed Installs/application_usage.sqlite: Could not prepare statement: no such table: install_requests
```

A genuine query failure still logs at error level, now as `Error querying <db>: Could not prepare statement: <sqlite3 message>`.

**What the change does**

Detects the missing-table case with the same prefix match `_detect_table` in `ApplicationUsageRecorder` already uses, logs it at debug level, and returns the documented no-data value: `-1` from `daysSinceLastInstallEvent`, since the table's absence proves no install request has ever been recorded, and `-2` from `daysSinceLastUsageEvent`, which stays conservative because no usage data exists at all. Removal decisions do not change: the caller tests `daysSinceInstallRequest >= 0`, so `-1` and `-2` fall through identically. One file, 21 lines of code, the rest is comments.

**Versions**

Present in 7.3.0 and `Munki7_4dev`. The Python implementation created `install_requests` lazily too and caught the same exception, but `str(err)` carried the real sqlite3 text, so the port lost the message rather than introduced the condition.

Discussion in MacAdmins Slack: https://macadmins.slack.com/archives/C04QVPFGU/p1784744636360719

Running in our fleet since late July. The error lines are gone from every run and the unused-software check behaves as before.

### DIFF
--- a/code/cli/munki/shared/appusage.swift
+++ b/code/cli/munki/shared/appusage.swift
@@ -233,7 +233,7 @@ class ApplicationUsageQuery {
         do {
             try conn = SQL3Connection(db)
         } catch {
-            logger.error("Error connecting to \(db): \(error.localizedDescription)")
+            logger.error("Error connecting to \(db): \(error)")
             conn = nil
         }
     }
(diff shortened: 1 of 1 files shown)


gregneagle:
Congratulations for being one of the few people to actually use unused_software_removal (as far as I know). :-) I wrote the feature on the request of a past director (as in my manager's manager) here, but in reality we've never really used it.

rodchristiansen:
We use it quite often! :) 

Never caught this cuz MunkiReport never brought it to surface, but [ReportMate](https://github.com/reportmate) does

PR #1393, opened 2026-09-09
Merge configured blocking applications

nstrauss:
This is a follow-up to [#1379](https://github.com/munki/munki/pull/1379) and [#1381](https://github.com/munki/munki/pull/1381), based on [Greg's follow-up comment](https://github.com/munki/munki/pull/1381#issuecomment-5593078051).

#1381 changed `blocking_applications_launch_args` from an array to a dictionary keyed by application identifier. This correctly associates arguments with individual applications, but requires an application with launch arguments to also appear in `blocking_applications`.

For example, Chrome is repeated here:

```xml
<array>
    <string>Firefox.app</string>
</array>
<dict>
    <key>Google Chrome.app</key>
    <array>
    </array>
</dict>
```

Since neither schema introduced by #1379 or #1381 has shipped, Greg and I agreed to replace the current key rather than retain that repetition. We also agreed to keep `blocking_applications` as an array instead of allowing multiple types under the established key.

## Changes
This replaces `blocking_applications_launch_args` with `blocking_applications_with_launch_args`. The dictionary keys are blocking applications. Munki merges them with the strings in `blocking_applications`, so an application with launch arguments does not need to be specified twice:

```xml
<array>
    <string>Firefox.app</string>
</array>
<key>blocking_applications_with_launch_args</key>
<dict>
    <key>Google Chrome.app</key>
    <array>
    </array>
</dict>
```

In this example, both Firefox and Chrome block the update. Chrome receives `--restore-last-session` if Managed Software Center closes and reopens it. Firefox reopens without additional arguments. A pkginfo with only a configured blocker is also valid:

```xml
<key>blocking_applications_with_launch_args</key>
<dict>
    <key>Google Chrome.app</key>
    <array>
    </array>
</dict>
```
`blocking_applications` remains an array of strings. This avoids changing the type expected by existing clients, linting tools, and editors.

## Behavior
The agreed upon behavior is as follow:
- If either `blocking_applications` or `blocking_applications_with_launch_args` is present, use the union of the array strings and dictionary keys.
- If neither key is present, retain the existing inference from application entries in installs.
- Resolve and deduplicate equivalent identifiers by the running application path, as the current #1381 implementation does.
- If an application appears in both keys, treat it as one blocker and use the arguments from the dictionary.
- If a dictionary value is malformed, keep its key as a blocker but ignore the arguments and reopen the application normally.

The same blocker selection is used by managedsoftwareupdate, the Assisted Quit sheet, and Managed Software Center's fallback blocking-app alert. The key is copied into processed managed-install, optional-install, optional-uninstall, and removal metadata.

## Testing
I added tests for:
- Copying blocking_applications_with_launch_args into processed update metadata.
- Combining blocking_applications strings and dictionary keys.
- Deduplicating an application present in both sources.
- Treating a malformed dictionary value as a blocker without arguments.
- Suppressing installs inference when an explicit blocker source is present.
- Suppressing inference with an explicitly empty dictionary.
- Retaining installs inference when both explicit sources are absent.
- Retaining installs inference when the outer dictionary has the wrong type.

The Assisted Quit relaunch path and direct argument handling remain the implementations tested in #1379 and #1381.

### DIFF
--- a/code/apps/Managed Software Center/Managed Software Center/Controllers/MSCAlertController.swift
+++ b/code/apps/Managed Software Center/Managed Software Center/Controllers/MSCAlertController.swift
@@ -591,15 +591,7 @@ class MSCAlertController: NSObject {
         }
         var apps_to_check = [String]()
         for update_item in getUpdateList() {
-            if let blocking_apps = update_item["blocking_applications"] as? [String] {
-                apps_to_check += blocking_apps
-            } else if let installs_items = update_item["installs"] as? [PlistDict] {
-                let installs_apps = installs_items.filter(
-                    { ($0["type"] as? String ?? "" == "application" &&
-                        !($0["path"] as? String ?? "").isEmpty) }).map(
-                            { ($0["path"] as? NSString ?? "").lastPathComponent })
-                apps_to_check += installs_apps
-            }
+            apps_to_check += blockingApplicationsForItem(update_item.my)
         }
         let running_apps = getRunningBlockingApps(apps_to_check)
         if running_apps.isEmpty {
--- a/code/apps/Managed Software Center/Managed Software Center/Controllers/MSCBlockingAppsController.swift
+++ b/code/apps/Managed Software Center/Managed Software Center/Controllers/MSCBlockingAppsController.swift
@@ -75,17 +75,15 @@ class MSCBlockingAppsController: NSObject {
     private let forceQuitDelay: TimeInterval = 5.0
 
     // Manual quit tracking - apps that cannot/should not be quit by us
-    private var manualQuitAppNames: Set<String> = [] // app names that require manual quit
     private var manualQuitAppPaths: Set<String> = [] // app paths that require manual quit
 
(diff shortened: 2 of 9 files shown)


gregneagle:
There are a lot more changes here than I expected. Nothing jumps out at me as problematic, but as the number of changes increases, the chances for unexpected/unwanted behavior also increases. I'm going to merge this, but when this goes into beta we need to encouage people to pay close attention to the assisted quit of blocking apps behavior to ensure there aren't any unexpected regressions.

gregneagle:
Can you start doing some preliminary/early documentation here? https://github.com/munki/munki/wiki/Launch-arguments-for-relaunched-applications

nstrauss:
> Can you start doing some preliminary/early documentation here? https://github.com/munki/munki/wiki/Launch-arguments-for-relaunched-applications

https://github.com/munki/munki/wiki/Launch-arguments-for-relaunched-applications First draft up.

